PR #1, opened 2019-08-07
feat: select keyboard navigation (#583)

smellyshovel:
Implement support for VaSelect keyboard navigation.

Use arrow keys to switch between options, tab/shift+tab to switch between different selects and/or search-field/options, backspace to clear selection.

### DIFF
--- a/src/components/vuestic-components/va-checkbox/KeyboardOnlyFocusMixin.js
+++ b/src/components/vuestic-components/va-checkbox/KeyboardOnlyFocusMixin.js
@@ -7,8 +7,9 @@ export const KeyboardOnlyFocusMixin = {
   },
   methods: {
     onFocus (e, index) {
-      if (this.hasMouseDown) return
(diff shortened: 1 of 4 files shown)


asvae:
Couple of things I noticed:

1) We need more pronounced selection on focus. For now the circle on chevron is the only indication:
image

2) Select normally doesn't cycle through options. We can add a prop to cycle, or we can remove. Up to you.

3) Keyboard navigation behaves weirdly if mouse cursor is placed on one of list items.

4) Chevron background doesn't look ok on colored select.
image

5) Disabled select still can be selected via keyboard navigation. That should never be the case (see checkbox, input etc).

6) Keyboard cursor should move from last mouse cursor position.

6) For search select keyboard cursor should go to search input.

Kreezag:
1 - fixed
2 - can not reproduce
3 - can not reproduce
4 - fixed 
5-  fixed
6 - need conversation
7 - can not reproduce

PR #3, opened 2019-08-08
feat: va-topbar

smellyshovel:
:warning: **Important**: make sure to review/merge the [admin-part of this PR](https://github.com/epicmaxco/vuestic-admin/pull/646) as well.

Also make sure to checkout respective branches on *both* repos during testing/reviewing.

### DIFF
--- a/src/components/vuestic-components/va-topbar/VaTopbarLink.vue
+++ b/src/components/vuestic-components/va-topbar/VaTopbarLink.vue
@@ -1,0 +1,188 @@
+<template>
+  <router-link
+    tag="li"
+    :class="computedLinkClass"
+    @mouseenter.native="updateHoverState(true)"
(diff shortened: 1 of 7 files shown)


src/components/vuestic-components/va-topbar/VaTopbarLink.vue
```diff
@@ -0,0 +1,188 @@
+<template>
+  <router-link
```
asvae: `li` is not ideal. Should be `a`. Reasons:
* target does nothing for `li`.
* Can't `Ctrl + Click` to open in new page.
* Can't copy url.

Also notice this kinda makes some `ul`'s lose value, so we might want to swap those with `div`'s

Kreezag: fixed

PR #4, opened 2019-08-14
fix: sidebar mobile height

lkho:
forgot to override sidebar height when the media query changed navbar to top: $sidebar-mobile-top


asvae:
Thanks!

Would be very nice to have a screenshot like this to explain what PR fixes/improves next time. Takes a bit of time to figure :).

image

PR #5, opened 2019-08-14
In datatable keep ?page= on refresh

jclappiway:
1/ on @page-selected => update route query with "page"
2/ on "created" check if route query contains page

### DIFF
--- a/src/components/vuestic-components/va-data-table/VaDataTable.vue
+++ b/src/components/vuestic-components/va-data-table/VaDataTable.vue
@@ -73,6 +73,10 @@ export default {
       type: Number,
       default: 4,
     },
+    queryPage: {
+      type: String,
+      default: 1,
+    },
     apiMode: Boolean,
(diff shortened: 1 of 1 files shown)


lkho:
I think the global side effects (changing routes) should not be built (or hard coded) into the UI library, instead we could easily add custom logic to the parent page component.

Any ideas?

asvae:
The logic you mentioned in many cases is different depending on specific app.
So I suggest it being done on per-app basis.

@lkho I completely agree.

jean-moldovan:
I haven't tested myself, but I think that the best way to handle this is to trigger table's `inputPage` method on query param change:

https://github.com/epicmaxco/vuestic-ui/blob/master/src/components/vuestic-components/va-data-table/VaDataTable.vue#L177

```
watch: {
  page (value) {
    this.$refs.mytable.inputPage(value)
  }
}
```

PR #19, opened 2019-09-09
data-table allow to pass currentPage as prop

sidneyarida:


asvae:
Thanks. Sorry it took so long.

PR #43, opened 2019-09-27
feat: data manager in va-data-table exposed

ssemenkoff:
Added property to set custom dataManager. Created test example.

Issue [#655](https://github.com/epicmaxco/vuestic-admin/issues/655) in vuestic-admin.


asvae:
Nice work. Thanks!

PR #44, opened 2019-09-27
fix: remove unnecessary z-index from va-card

ssemenkoff:
Fix for the issue (#656)[https://github.com/epicmaxco/vuestic-admin/issues/656].

Removed z-index property for the card stripe.


asvae:
Thanks.

PR #47, opened 2019-09-30
Add sort order to data table

homeforce:
Add a sort order without creating a data-manager


asvae:
Thanks!

homeforce:
> Thanks!

You're welcome, glad i could contribute.

PR #48, opened 2019-09-30
fix: added bottom padding in tabs wrapper when scrolling is available

ssemenkoff:
Added bottom padding in tabs wrapper when scrolling is available to prevent scroll handle being over tabs slider.

Changed overflow property of wrapper to fix scroll problems in Safari on iOS (tabs were horizontally scrollable).

Connected with this [issue](https://github.com/epicmaxco/vuestic-admin/issues/641).


asvae:
Merged manually.

PR #65, opened 2019-10-04
fix: restore mechanism for themes

ssemenkoff:
Added theme restore mechanism for ColorThemePlugin.
Updated vue-epic-bus version to prevent issues with transpileDependencies. 

Connected with this [PR](https://github.com/epicmaxco/vuestic-admin/pull/665)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "vue-chartjs": "^3.4.0",
     "vue-clipboard2": "^0.3.0",
     "vue-color": "^2.7.0",
-    "vue-epic-bus": "^0.1.2",
+    "vue-epic-bus": "^0.1.3",
     "vue-flatpickr-component": "^8.0.0",
     "vue-router": "^3.0.3",
     "vue-toasted": "^1.1.25",


src/services/ColorThemePlugin.js
```diff
@@ -19,8 +19,16 @@ export const ColorThemePlugin = {
       Object.assign(defaultOptions.themes, options.themes)
     }
 
+    let themeCache = Object.assign({}, defaultOptions.themes)
+
     Vue.prototype.$themes = defaultOptions.themes
 
+    Vue.prototype.$themes.restoreThemes = () => {
```
asvae: That shouldn't be a public api. Please just implement in component.

ssemenkoff: Fixed in PR in vuestic-admin repository referenced above.

asvae: @ssemenkoff I'm not following? Do I close this PR then?

"vue-epic-bus": "^0.1.3",
This change seems useful.

ssemenkoff: @asvae You can merge it now.

Issue #64, opened 2019-10-04
Radio-button and input type radio does not look good


Kreezag:
Checkbox dot is crooked.
Need fix paddings
<img width="184" alt="Vue App 2019-10-04 12-14-34" src="https://user-images.githubusercontent.com/13301570/66196190-c7a2c580-e6a0-11e9-8766-d3eb8a1c169d.png">

asvae:
Reimplementation is planned for radio button. We'll likely fix as a part.

PR #78, opened 2019-10-08
Fix/radio button styles

Kreezag:
Fix issue  [#64](https://github.com/epicmaxco/vuestic-ui/issues/64) 

Change display of radio-button
Fixed disabled state styles 
Removed unusually custom logic
Improve UX of component (add tag label).

Proposal: need additional support for use checkbox with the same name property as [a browser standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio)

### DIFF
--- a/src/components/vuestic-components/va-radio-button/VaRadioButton.demo.vue
+++ b/src/components/vuestic-components/va-radio-button/VaRadioButton.demo.vue
@@ -1,14 +1,15 @@
 <template>
-  <div class="demo-container">
-    <div class="demo-container__item">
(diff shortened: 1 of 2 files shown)


asvae:
Please create another PR with only things that are stated in the issue. From my point of view, SVG should handle that visual discrepancy.

Currently your implementation breaks more things than it fixes (keyboard navigation and statelesness went away). Furthermore radio-button is an old component and has to be researched + reimplemented pretty much.

Please discuss with me if something is not understood, no heroic acts :).

PR #85, opened 2019-10-12
add lock files to gitignore

Kreezag:
add package-lock.json and yarn.lock to ignored files


asvae:
Sorry, but this does not make any sense for library.

PR #89, opened 2019-10-15
Feat/breadcrumbs component

### LINKED ISSUE #39: Breadcrumbs component
# Docs

- [x] Do after [docs boilerplate](#17) is done.

### Props
- [x] `separator: String` - default: `/` - specifies the dividing character between items
- [x] `align: String` - default: `left` - should use flex mixins (check quasar implementation + please don't use global classes!), also please move into mixin (both css and js, pretty sure we'll need this elsewhere),
- [x] `color: String` - colors everything (use color theme mixin) 
- [x] `active-color: String`  - colors active items (that means not last one) (use color theme mixin) 
- [x] `separator-color (*): String` - separator color (use color theme mixin) 

### Slots
- [x] `default`- expected: breadcrumbs item
- [x] `separator` - HTML or component you can slot in to separate the breadcrumbs - slot takes priority if prop and slot are both defined

### Props
- [x] `to / replace / exact  / append ` - router props (use mixin for vue-router props)
- [x] `disabled: Boolean`

# Design
- [x] Present to @xx13 after finishing.

# Refs
* [Vuetify](https://vuetifyjs.com/en/components/breadcrumbs)
* [Quasar](https://quasar.dev/vue-components/breadcrumbs)
* [Ant Design](https://vue.ant.design/components/breadcrumb/)
* [iView](http://iview.talkingdata.com/#/components/breadcrumb-en)
* [Element UI](https://element.eleme.io/#/en-US/component/breadcrumb)

Kreezag:
close #39
Create Breadcrumbs components
Add align mixin
Add RouterLinkMixin
Update RouterLinkMixin to Button component

### DIFF
--- a/src/components/vuestic-components/va-breadcrumbs/VaBreadcrumbItem.vue
+++ b/src/components/vuestic-components/va-breadcrumbs/VaBreadcrumbItem.vue
@@ -56,8 +56,7 @@ export default {
       cursor: pointer;
 
       &:hover {
-        text-decoration: underline;
-        filter: sepia(20%) grayscale(10%);
+        opacity: 0.7;
       }
 
       &:focus {
--- a/src/components/vuestic-components/va-breadcrumbs/VaBreadcrumbItem.vue
+++ b/src/components/vuestic-components/va-breadcrumbs/VaBreadcrumbItem.vue
@@ -1,0 +1,81 @@
+<template>
+  <span  class="va-breadcrumb-item">
+    <router-link
+      v-if="!isDisabled"
+      class="va-breadcrumb-item__label va-breadcrumb-item__label--link"
+      :to="to"
+      :replace="replace"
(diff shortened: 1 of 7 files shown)


asvae:
I did some tweaks to demos. Component was pretty much excellent!

Couple of things I noticed:

1) Active color + disabled doesn't look right
image

2) Why default color is gray?
image

3) Can you sent this to @xx13 for some feedback?

xx13:
@asvae 
1. We ended up having 2 types of breadcrumb items: 1) link, 2) text (disabled link). So there are only 2 styles involved.
2. Grey was the default color from the very start, but we used to have green as a hover color. We changed that and made a lighter version as a hover color to have a flexible solution for any color that users might pick for their breadcrumbs / dividers.
3. We were in touch when creating that component ;)

PR #111, opened 2019-10-30
fix(va-toggle): make va-toggle inline, display: inline-flex

### LINKED ISSUE #71: va-toggle has incorrect touched area
va-toggle touched area is more than element
Example: 
ezgif com-video-to-gif

jean-moldovan:
Fixes #71


Kreezag:
It looks good to me! Thanks

PR #120, opened 2019-11-10
context 2.0

asvae:
* Make context fully dynamic
* Apply new syntax to button component

### DIFF
--- a/src/components/vuestic-components/va-button/VaButton.vue
+++ b/src/components/vuestic-components/va-button/VaButton.vue
@@ -4,10 +4,10 @@
     class="va-button"
     :class="computedClass"
     :style="computedStyle"
-    :disabled="disabled"
-    :type="type"
-    :href="href"
-    :target="target"
+    :disabled="c_disabled"
+    :type="c_type"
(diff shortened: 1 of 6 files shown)


src/components/vuestic-components/va-button/VaButton.vue
```diff
@@ -53,86 +53,40 @@ import {
...
-      default () {
-        return getContextPropValue(this, 'flat', false)
-      },
-    },
-    size: {
+  mixins: [
+    ColorThemeMixin,
+    RouterLinkMixin,
+
+    getContextablePropMixin('color', { type: String, default: 'success' }),
```
Kreezag: I suggest using the mixin factory before initializing the component.
This may be to make component rendering easier.

For example:
```
const contextMixinButton = getContextablePropMixinFuctory([
{'color', { type: String, default: 'success' }},
{'tag', { type: String, default: 'button' }},
...
])()

const component = {
   mixins: [ contextMixinButton ]
}
```

The second benefit can be splitting component config with component construction.

Issue #104, opened 2019-10-28
Valideteable trait


asvae:
## Gist
We want simple template-based validation. Which is what quasar/vuetify do.
We also want to support some external libs to support more complex validation cases. (see #112)

- [x] Implement this check for one component (f.i. input). Then do demos and send for review.
- [ ] After review apply implementation to other form components.

## Props
- [x] `rules` - see vuetify solution from below.

## Public methods
- [x] `validate (): boolean` - validates input element according to rules and stores validation results internally
- [x] `resetValidation (): void` - should wipe any internal error messages.

## Add prop to va-form
- [ ] `lazy-validation: Boolean` - trigger validation only on blur (not on input).

## Additional information

Here's some of the approaches:
* [**Vuetify**](https://vuetifyjs.com/en/components/forms)
rule is either defined in template (reactive prop) or/and function. 
If it's not a function - we check if result is `true`. If it is - input is considered valid. In other cases - invalid.
If it's function we run the function (passing value as argument) then perform the same validity check.
Validation keeps its internal state. I.e. if input is changed, but validate method was not run - validation messages should stay the same.

* **Quasar**
Pretty much vuetify solution with rule aliases.

* **Ant**
Ant looks very weird. Seems like a mix between jQuery and vue. Also not too well documented.

* **iView**
Uses async validator internally. validation rules + data is passed to the form. Additionally props are defined for each form item, which allows the form to figure for each data item validation is triggered. Looks fairly complicated and gimmicky. Nested data doesn't seem to be possible.

PR #121, opened 2019-11-11
Feat/va form

### LINKED ISSUE #103: Form component
## Docs
- [x] Do after docs boilerplate (#17) is done.

## Props
- [x] `autofocus: Boolean` - focus on first element on render

## Slots
- [x] `default` - form components. For now we support only one level deep.

## Events
- [ ] `focus: () => void` - triggered when some form element is triggered. Might be useful for multi-tab forms. 
*- can use `focusin` native event*
- [x] `validation: (valid: Boolean) => void` - triggered on validation update

## Notes
* See validetable trait #104. It's a prerequisite for form.
- [ ] It should be possible to use one form inside of another form. The way it should work is that external form should trigger relevant method on internal form. (medium priority: if it's too hard - leave)

## Methods
- [x] `validate(): boolean` - performs validation and updates error state for each form component.
- [x] `focus(): void` - focus on first **focusable** element in the form.
- [x] `focusInvalid(): void` - focus on first invalid **focusable** component in the form. If there are none - should default to `focus()`.
- [x] `resetValidation(): void` - reset validation for all **validateable** form elements
- [x] `reset(): void` - reset value + validation stater for all form elements.

# Refs
* [Vuetify](https://vuetifyjs.com)
* [Quasar](https://quasar.dev)
* [Ant Design](https://ant.design)
* [iView](http://iview.talkingdata.com)
* [Element UI](https://element.eleme.io)

Kreezag:
## Description
Close #103 

Add va-form component

## Markup:

<details>

```vue
<va-form autofocus tag="form">
  <va-input label="name" v-model="Input"/>
  <va-select label="city" v-model="Select" options="options"/>
  <va-button @click="submitForm">Submit</va-button>
</va-form>
```
</details>

## Types of changes


src/components/vuestic-components/va-checkbox/VaCheckbox.vue
```diff
@@ -128,6 +129,12 @@ export default {
       },
     },
 
+    autofocus: {
```
asvae: I'm not seeing too much value for autofocus being contextable.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
+<script>
+import { ContextPluginMixin, getContextPropValue } from '../../context-test/context-provide/ContextPlugin'
+
+const availableFormTags = ['VaInput', 'VaColorInput', 'VaDatePicker', 'VaCheckbox', 'VaRadio', 'VaSelect', 'VaFileUpload']
```
asvae: Component should tell form if it's form component, not the other way round.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+const getAllChildren = (vm, children = []) => {
+  vm.$children.forEach(function (child) {
```
asvae: arrow function

asvae: This approach won't work. We have to stop as soon as we find form component.
We should not handle its internals.

I.e. you can have double input marked as form component. It handles validation and such internally.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+    tag: {
+      type: String,
+      default () {
+        return getContextPropValue(this, 'tag', 'div')
+      },
+    },
+  },
+  data () {
+    return {
+      valid: true,
```
asvae: I'm not seeing any purpose, keeping valid state for form itself.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+      hasResetButton: false,
```
asvae: Same here. Buttons can be added dynamically and state won't play nicely.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+      hasSubmitButton: false,
+    }
+  },
+  mounted () {
+    this.$el.addEventListener('focusin', this.focus)
```
asvae: Why not handle this in template as we do with normal components?

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+    const formChild = getAllChildren(this)
```
asvae: `focusable` should focus on first item.
Yet `this.$children` ordering has no relation whatsoever to visual ordering.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+        submitButton.addEventListener('click', this.submit)
+
+        this.hasSubmitButton = true
+      }
+    }
+  },
+  beforeDestroy () {
+    this.$el.removeEventListener('focusin', this.focus)
+
+    if (this.hasSubmitButton) {
```
asvae: Pretty sure that's pointless. Are you worried vue will reuse the same dom element?

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,162 @@
...
+          if (this.lazyValidation) {
+            break
+          }
+        }
+      }
+    },
+  },
+
+  render (createElement) {
+    return createElement(this.tag, {
```
asvae: Any reason to use render function instead of template?

src/components/vuestic-components/va-input/VaInput.demo.vue
```diff
@@ -6,6 +6,13 @@
         placeholder="Name"
       />
     </VbCard>
+    <VbCard title="autofocus">
```
asvae: That's native input attribute and there is no reason to test it.
We don't bring any new behavior in.

src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -37,9 +37,11 @@
           :disabled="disabled"
           :readonly="readonly"
           :value="value"
+          :autofocus="autofocus"
```
asvae: These should be covered via `v-bind="$attrs"`, no?

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
     },
   },
   methods: {
-    onClick () {
+    click (e) {
       if (!this.disabled) {
-        this.$emit('input', this.option)
+        return this.$emit('input', this.option, e)
```
asvae: o.O why return here?

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
     },
   },
   methods: {
-    onClick () {
+    click (e) {
```
asvae: Naming was correct. `onClick` or `clicked` is used for internal event handling.
`click`/`focus` is used as public method.

For your solution, if you trigger focus via ref - nothing good will happen.

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
...
       if (!this.disabled) {
-        this.$emit('focus', e)
+        return this.$emit('focus', e)
       }
     },
+    validate () {
+      return true
+    },
+    clear (e) {
+      return this.$emit('input', null, e)
```
asvae: I don't completely understand why would you need to pass native event.

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
...
-  }
-
-  &--disabled {
-    cursor: default;
-  }
+  @import "../../vuestic-sass/resources/resources";
 
-  &--leftLabel {
-    flex-direction: row-reverse;
+  .va-radio {
```
asvae: There should be no default tab for styles block. Only template should have tab.

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
...
-    transition: $transition-primary;
-    display: inline-flex;
-    align-items: center;
-    width: 1.4rem;
-    height: 1.4rem;
-    border-radius: 100%;
-    position: relative;
-    border: $gray solid 0.125rem;
-    box-sizing: border-box;
+    &--leftLabel {
```
asvae: Should be no capital letters in scss classes.

src/components/vuestic-components/va-radio/VaRadio.vue
```diff
@@ -108,135 +138,141 @@ export default {
...
-      z-index: 0;
-      opacity: 0;
+      position: relative;
+      border: $gray solid 0.125rem;
+      box-sizing: border-box;
 
-      .va-radio__input:focus + .va-radio__icon & {
-        opacity: 0.2;
+      .va-radio__input:disabled + & {
+        opacity: 0.4;
```
asvae: We should have variable for this opacity value.

src/components/vuestic-components/va-checkbox/VaCheckbox.vue
Kreezag: Yeap. Now it's an unnecessary prop

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: Yes! We need to know form components because need call focus and validate on them.

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: We need to know all child because we need to call reset method on them when a form will be clear

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: After the implementation of all methods to form-compoents (like a reset, clear, focus, etc.) we need to remove this availableFormElements

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: This need to implement focusInvalid method at the form and resetValidation

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: These fields need to watch native form event like a submit and reset in cases when a form has tag div. Probably we won't use dynamically buttons with type submit and reset

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: Focusable components can be wrapped in another component. We need handling focus event on any nesting level

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: Focusable components can be wrapped in another component.

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: In some browsers memory leak is a big problem.
https://stackoverflow.com/questions/6033821/do-i-need-to-remove-event-listeners-before-removing-elements

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,163 @@
...
+
+    if (this.autofocus) {
+      const firstFormChild = getAllChildren(this).find((child) => availableFormComponents.includes(child.$options.name))
+
+      if (firstFormChild) {
+        firstFormChild.$el.focus()
+      }
+    }
+
+    if (this.tag !== 'form') { // component dont have tag form, we need set listeners manually
```
jean-moldovan: @asvae , not clear to me why would anyone use anything other than `form` tag.

src/components/vuestic-components/va-form/VaForm.vue
```diff
@@ -0,0 +1,163 @@
...
+      if (resetButton) {
+        resetButton.addEventListener('click', this.reset)
+
+        this.hasResetButton = true
+      }
+
+      const submitButton = this.$el.querySelector('button[type=submit]')
+
+      if (submitButton) {
+        submitButton.addEventListener('click', this.submit)
```
jean-moldovan: I'm not sure we need to listen to the click event on the submit button. Clicking on the submit button will trigger form submit event by default:
```
    <form @submit="doSmth">
      <button type="submit">
        go
      </button>
    </form>
```

src/components/vuestic-components/va-form/VaForm.vue
asvae: @Kreezag so are you implying this line is temporary and you agree with what I said? :)

src/components/vuestic-components/va-form/VaForm.vue
asvae: @Kreezag No we don't.
If component handles its clear state - form should never interfere.
I.e. for double input I mentioned - default (reset) state could be: one input 0, second 100. And it should be provided from parent, given parent handles this logic.

src/components/vuestic-components/va-form/VaForm.vue
asvae: @Kreezag 

Still not seeing any use. Why not make these dynamic?

src/components/vuestic-components/va-form/VaForm.vue
asvae: Couple of approaches for reference.

* `this.$form.reset()`, `this.$form.submit()`. `$form` here could be inject/provided as per usual. Probably the top level form would make more sense to have priority in case we have nested forms.
* Let the user handle it on its own. So to get reset - update model, to submit the form - do `this.$refs.form.submit()`.

src/components/vuestic-components/va-form/VaForm.vue
asvae: @Kreezag 

How is this solution different from `@focusin` on `form` element?

src/components/vuestic-components/va-form/VaForm.vue
asvae: @Kreezag sorry, but this has no relation whatsoever to what I said.

src/components/vuestic-components/va-form/VaForm.vue
asvae: In issue you sent it says we're fine for IE11+ which is the case for vuestic.
So please remove if you don't have any other reasons to keep this code.

src/components/vuestic-components/va-form/VaForm.vue
asvae: @jean-moldovan could be useful for nested forms (i.e. you can reuse form snippets).

For `form` tags nesting is invalid.

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: We can support form inside form cases. Low-level form can be as a div tag

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: We can support form inside form cases.

src/components/vuestic-components/va-form/VaForm.vue
asvae: @jean-moldovan it's likely we're going to choose the strategy you're suggesting.

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: I think to an interface as `focus()` method on the parent form is better than `@focusin`

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: Yes, I can remove this field

src/components/vuestic-components/va-form/VaForm.vue
asvae: As discussed in chat, focusin can go away. Native event is good enough.

src/components/vuestic-components/va-form/VaForm.vue
asvae: Fixed

src/components/vuestic-components/va-form/VaForm.vue
asvae: Fixed

src/components/vuestic-components/va-radio/VaRadio.vue
asvae: Can we do `focus` + `onFocus` here? Pretty sure focus is useful for form etc.

src/components/vuestic-components/va-radio/VaRadio.vue
asvae: Aaand I still don't. Please explain :).

src/components/vuestic-components/va-radio/VaRadio.vue
asvae: I checked, we have mixin: `@include va-disabled;`. Please use that.

src/components/vuestic-components/va-form/VaForm.vue
Kreezag: removed

src/components/vuestic-components/va-radio/VaRadio.vue
Kreezag: fixed

src/components/vuestic-components/va-radio/VaRadio.vue
Kreezag: fixed

Kreezag:
The `validation` event does not currently work with form elements. There may be errors in console for va-input.
It fixed in #104

PR #135, opened 2019-11-29
fix: va-data-table num/alphanum fields sorting

rottingcorpse:
## Description

- Currently va-data-table doesn't take in account that fields could have non-string values and, thereby, fails to sort them with error trying to call localeCompare() on non-string variable. This fix forcefully converts value in comparator to localeString and uses "numeric" flag for localeCompare() to provide accurate sorting for numeric and alphanumeric values.
- Unit tests for changed functionality were added.

## Types of changes


asvae:
Thanks.

PR #136, opened 2019-12-02
Feat/va components fix

AlexanderBoriskin:
Related to #679 (vuestic-admin issue):
1) Bug 1 - added property 'is-collapse' to provide collapse behaviour for va-button
2) Bug 2 - added property 'delay' and updated popover demo to provide possibility to set delay from user side
3) Bug 3 - added property 'default-header' to remove va-table styles from header; 'row-class' property provides necessary styles for rows; further (after merging PR data table presentation should be updated
4) Bug 5 - area of prepend slot for input now placed inside input container
5) Bug 9 - va-input styles applied to custom input inside chrome picker (see va-advanced-color-picker component)

### DIFF
--- a/src/components/vuestic-components/va-button/VaButton.vue
+++ b/src/components/vuestic-components/va-button/VaButton.vue
@@ -87,6 +87,10 @@ export default {
     },
     type: {
       type: String,
+      default: 'button',
+      validator: value => {
+        return ['button', 'collapse'].includes(value)
+      },
     },
     disabled: {
       type: Boolean,
@@ -131,8 +135,8 @@ export default {
         'va-button--flat': this.flat,
         'va-button--outline': this.outline,
         'va-button--disabled': this.disabled,
-        'va-button--hover': this.hoverState,
-        'va-button--focus': this.focusState,
+        'va-button--hover': this.hoverState && this.type !== 'collapse',
+        'va-button--focus': this.focusState && this.type !== 'collapse',
         'va-button--without-title': !this.hasTitleData,
         'va-button--with-left-icon': this.icon,
         'va-button--with-right-icon': this.iconRight,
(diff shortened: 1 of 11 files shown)


src/components/vuestic-components/va-button/VaButton.vue
```diff
@@ -265,11 +269,12 @@ export default {
   &--default {
     color: $white;
 
-    &:hover {
+    &.va-button--hover {
```
Kreezag: On styleguide should be `&--hover`

src/components/vuestic-components/va-data-table/VaDataTable.demo.vue
```diff
@@ -7,20 +7,28 @@
...
+    VaDataTableCustomStyling,
     VaDataTableStaticDataExample,
     VaDataTableApiModeExample,
   },
 }
 </script>
 
 <style lang="scss">
+  @import '../../vuestic-sass/resources/resources';
+  @import './../../vuestic-sass/global/typography';
```
Kreezag: Please, leave a comment about these imports, if these imports really need it.

src/components/vuestic-components/va-collapse/VaCollapse.demo.vue
```diff
@@ -10,7 +10,11 @@
     </VbCard>
     <VbCard title="Custom header">
       <va-collapse style="width: 400px;" customHeader>
-        <va-button slot="header" style="width: 100%">
+        <va-button
+          slot="header"
+          style="width: 100%"
+          type="collapse"
```
Kreezag: Wontfix:
Please, change prop type to `is-collapse` boolean prop.  In this case, type look worse

asvae:
Merged manually

PR #169, opened 2020-02-04
Feat/color helpers

Mikkou:
finished

### DIFF
--- a/src/components/vuestic-utilities/index.js
+++ b/src/components/vuestic-utilities/index.js
@@ -1,0 +1,24 @@
+import { colorConfig } from '../vuestic-components/va-color-picker/VuesticTheme'
+
+const createBackgroundHelpers = () => {
+  let result = ''
+  Object.keys(colorConfig).map((name) => {
+    result += `.background--${name.toLowerCase()} { background-color: ${colorConfig[name]}; }`
+  })
+  return result
+}
+
+const createTextColorHelpers = () => {
+  let result = ''
+  Object.keys(colorConfig).map((name) => {
+    result += `.text--${name.toLowerCase()} { color: ${colorConfig[name]}; }`
+  })
+  return result
+}
+
+export default (() => {
+  const $style = document.createElement('style')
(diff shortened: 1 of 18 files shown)


src/components/vuestic-utilities/utilities.demo.vue
```diff
@@ -0,0 +1,20 @@
+<template>
```
asvae: Please check other demos. Use provided wrappers.

asvae: One demo for one case. So file name should be `color-helpers.scss`

src/components/vuestic-utilities/index.js
```diff
@@ -0,0 +1,24 @@
+import { colorConfig } from '../vuestic-components/va-color-picker/VuesticTheme'
```
asvae: We don't use `index.js` names. Instead you can use `color-helpers-plugin.`

src/components/vuestic-utilities/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+const createTextColorHelpers = () => {
+  let result = ''
+  Object.keys(colorConfig).map((name) => {
+    result += `.text--${name.toLowerCase()} { color: ${colorConfig[name]}; }`
+  })
+  return result
+}
+
+export default (() => {
```
asvae: This is not how Vue plugins are normally implemented. Check docs.
https://vuejs.org/v2/guide/plugins.html

src/components/vuestic-utilities/index.js
```diff
@@ -0,0 +1,24 @@
+import { colorConfig } from '../vuestic-components/va-color-picker/VuesticTheme'
+
+const createBackgroundHelpers = () => {
+  let result = ''
+  Object.keys(colorConfig).map((name) => {
+    result += `.background--${name.toLowerCase()} { background-color: ${colorConfig[name]}; }`
```
asvae: Color should have !important. Please check other frameworks implementations.

Issue #148, opened 2020-01-08
color helpers


asvae:
Initial idea is to add some color helpers, like `text-danger`, `text-primary`. But we can probably add some other things as we go. I.e. background color.

- [x] Check other framework solutions in regards to naming. We want standard approach. Replace naming sсheme if you think something else works better.
- [x] `color-[color]` for all theme colors - example: `color-danger`
- [x] `bacground-[color]` for all theme colors - example: `background-danger`.
- [x] Implementation should be similar to what is done for [epic-spinners](https://github.com/epicmaxco/epic-spinners) colors. There style block is injected in body initially, then dynamically overriden each time we have a change.
- [x] color helpers should be independent of components (separate system/plugin).

Mikkou:
https://github.com/epicmaxco/vuestic-ui/pull/169

PR #170, opened 2020-02-04
Feat/color helpers

Mikkou:
finished https://github.com/epicmaxco/vuestic-ui/issues/148 only without "The things we're most intrested are dynamic interactions".


ilearnjs:
Hi @Mikkou. Thanks a lot! Added dynamic interaction

PR #177, opened 2020-02-09
Improve context interface

### LINKED ISSUE #173: Investigate alternative interfaces for context
Currently we only support object component-based interface. Which is too verbose as well as not too useful in many cases. Ideally we want slicker interfaces.

asvae:
close #173

Played a bit with context interface + documented some pieces.

## Per prop config

That's our standard config. No changes here.
```vue
<va-context
  :config="{ ContextTestInterfaces: {value: 'full object config'} }"
>
  <ContextTestInterfaces/>
</va-context>
```

That's per prop config. It has higher priority for prop in case it's declared in both config types.
```vue
<va-context
  :config="{ ContextTestInterfaces: {value: 'full object config'} }"
  ContextTestInterfaces::value="per-value config"
>
  <ContextTestInterfaces/>
</va-context>
```

## All components config
placing `all` instead of component name allows to apply prop value to all components. If value is specified for component on same context level - it takes priority instead of `all`.
```vue
<va-context
  :config="{ all: {value: 'all-config'} }"
  all::value="all-config"
>
  <ContextTestInterfaces/>
</va-context>
```


src/components/context-test/context-provide/ContextPlugin.js
```diff
@@ -133,3 +146,19 @@ function getLocalConfigWithComponentProp (configs, componentName, propName) {
     return componentConfig && componentConfig.hasOwnProperty(propName)
   }) || undefined
 }
+
+// Just 2 levels deep merge. B has priority.
+export function mergeConfigs (configA, configB) {
```
ilearnjs: Can we use lodash merge or mergeWith?

asvae: @ilearnjs nah. This one is more like merge strategy. We might change it to something else.

PR #182, opened 2020-03-04
Feat/component api

### LINKED ISSUE #75: API presenter
image

We want this to function in following way:
* There should be a data type, describing api with all data for api display inside. (ApiOptions)
* There should be a mapper build/function, taking in string, and returning ApiOptions.
* There could be a number of intermediate types:
  * ComponentOptionsComponentConfig (api options derived from dynamic analysis of component options).
  * OverridesComponentConfig (defiled in separate file, items we can't get from automated approaches, mixins can also be defined here).
  * [low priority] ComponentOptionsStaticComponentConfig (api options derived from dynamic analysis of component options).
  * [low priority] DeclarationComponentConfig (derived from .d.ts)
* Ideally we want intermediate types to be compatible with ApiOptions (maybe ApiOptions could be defined as union type?)

## Features
- [ ] Link to github source
- [ ] Search

#### All sections
- [ ] name
- [ ] added in (should not be shown for initial props)
- [ ] description - optional

#### Props
- [ ] type (should use typescript declaration)
- [ ] default value
- [ ] examples - optional

#### Events
On top of section
- [ ] Propagation warning

For each event:
- [ ] typescript signature

#### Slots
* only "all sections" stuff here

#### Methods
- [ ] typescript signature

* we want to get reuse value in cases where mixin is used

## Examples from different frameworks
Quasar - a lot of features
image

Vuetify - fairly feature rich, component selector
image

Bootstrap-vue - poor man's API (just table)
image

vue-ant - poor man's API (just table)
image

asvae:
closes #75

What is done here:
* I took vue src from [vue repo](https://github.com/vuejs/vue). Idea was to use vue native functions to work with component options. Might be not a greatest approach.
* Added some tests for component options to api options conversion. We should continue using unit tests for these.
* Added a `api-docs` component, that works as follows.
```vue
<api-docs :componentOptions="VaCheckbox" />
```
Currently it generates a table with props.

I also investigated vuetify documentation principles, so if someone wants to continue this - discussing with me first might be a good idea.


asvae:
A bit of premature merge, but that's needed in develop.

Issue #151, opened 2020-01-12
Button without text should looks the same as normal button


asvae:
![image](https://user-images.githubusercontent.com/5394573/72225269-6ee2a300-3594-11ea-9122-8a93b39e3a96.png)

- [ ] check on all button sizes
- [ ] check refs for implementation principles

PR #184, opened 2020-03-13
fix: empty button appearance

bdrtsky:
#  Description

Fixing https://github.com/epicmaxco/vuestic-ui/issues/151 by adding min-width to button and removing y-axis padding as sizing factor, adding fixed height instead. This method is used by multiple frameworks, like Vuetify.

## Types of changes


asvae:
Thanks. All good.

PR #191, opened 2020-03-17
fix: remove epic-spinners

### LINKED ISSUE #188: Remove epic-spinners from vuestic-ui
Rationale:

* Is used only for loaders.
* Breaks tests (we want tests to be part of CI like right now)

Currently it's used to show spinner on loading in select and data table. We should just use icon instead. You may add `spin` prop to va-icon for this to work.

bdrtsky:
closes #188

## Description
Remove epic-spinners

## Types of changes

### DIFF
--- a/src/components/vuestic-components/va-icon/VaIcon.vue
+++ b/src/components/vuestic-components/va-icon/VaIcon.vue
@@ -63,7 +63,7 @@ export default {
       return (this.icon && this.icon.component) || this.c_component || this.c_tag
     },
     computedClass () {
-      return this.icon ? this.icon.iconClass : this.spin ? 'fa fa-circle-o-notch fa-spin' : ''
+      return this.icon ? this.icon.iconClass : this.spin ? 'material-icons spin' : ''
     },
     hasClickListener () {
       return this.$listeners && this.$listeners.click
@@ -89,7 +89,7 @@ export default {
       }
     },
     computedContent () {
-      return this.icon && this.icon.content
+      return this.icon ? this.icon.content : this.spin ? 'refresh' : ''
     },
   },
 }
@@ -99,5 +99,19 @@ export default {
 .va-icon {
   vertical-align: middle;
   user-select: none;
+
+  &.spin {
+    animation: rotate 1000ms infinite;
+  }
+}
+
+@keyframes rotate {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
 }
 </style>
--- a/src/components/vuestic-components/va-icon/VaIcon.vue
+++ b/src/components/vuestic-components/va-icon/VaIcon.vue
@@ -55,6 +55,17 @@ const IconContextMixin = makeContextablePropsMixin({
 export default {
   name: 'VaIcon',
   mixins: [ColorThemeMixin, SizeMixin, IconContextMixin, vaIconMixin],
+  data () {
+    return {
+      spinKeyframes: `
+        from {
+          transform: rotate(0deg);
+        }
+        to {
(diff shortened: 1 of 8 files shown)


src/components/vuestic-components/va-icon/VaIcon.vue
```diff
@@ -85,7 +89,7 @@ export default {
       }
     },
     computedContent () {
-      return this.icon && this.icon.content
+      return this.icon ? this.icon.content : this.spin ? 'refresh' : ''
```
asvae: Spin prop should function similarly to `fa-spin`. It shouldn't modify existing icon.

src/components/vuestic-components/va-icon/VaIcon.vue
```diff
@@ -95,5 +99,19 @@ export default {
 .va-icon {
   vertical-align: middle;
   user-select: none;
+
+  &.spin {
```
asvae: Bem.

src/components/vuestic-components/va-icon/VaIcon.vue
```diff
@@ -95,5 +99,19 @@ export default {
...
+    animation: rotate 1000ms infinite;
+  }
+}
+
+@keyframes rotate {
```
asvae: Please use something similar to [this solution](https://github.com/epicmaxco/epic-spinners/blob/master/src/services/utils.js).
We ideally do not want to introduce any global css.

src/components/vuestic-components/va-icon/VaIcon.vue
```diff
@@ -46,6 +46,10 @@ const IconContextMixin = makeContextablePropsMixin({
     type: [String, Number],
     default: '',
   },
+  spin: {
```
asvae: All props should be represented in demo.

Please also add all icontypes as well as svg/component icons to card to ensure spinning works properly in all cases.

src/components/vuestic-components/va-icon/VaIcon.vue
```diff
@@ -59,7 +63,7 @@ export default {
       return (this.icon && this.icon.component) || this.c_component || this.c_tag
     },
     computedClass () {
-      return (this.icon && this.icon.iconClass) || ''
+      return this.icon ? this.icon.iconClass : this.spin ? 'material-icons spin' : ''
```
asvae: Why would you add `material-icons` here? This should work with any iconfont.

src/components/vuestic-components/va-icon/VaIcon.vue
bdrtsky: What is `icontypes`?

asvae: Meant iconfont, sorry.

src/components/vuestic-components/va-icon/VaIcon.vue
bdrtsky: Material icons working by adding text in `i` tag, not CSS classes

src/components/vuestic-components/va-icon/VaIcon.vue
bdrtsky: Different icon fonts working differently. Awesome working by adding CSS class, Material - by adding text in `i`. How I suppose to detect which font is used?

src/components/vuestic-components/va-icon/VaIcon.vue
bdrtsky: Still not clear what `all cases` are meant? What `all cases`?

asvae:
One notable miss was that animation didn't have linear flag on it (when fa-spin did).

I fixed that, and also a couple of minor things.

PR #199, opened 2020-03-23
Selectable List Mixin implemented:

### LINKED ISSUE #192: SelectableListMixin
Current use cases are [OptionGroup](https://github.com/epicmaxco/vuestic-ui/issues/88) and Select components.

Please discuss with @asvae if during your research you uncovered something that doesn't fit into current issue.

- [x] Should include `FormComponentMixin`

## Props

- [x] `options: Array`: select options
- [x] `textBy: String | option => (string | number)` - see `item-text` for [vuetify implementation](https://vuetifyjs.com/en/components/selects/). 
- [x] `valueBy: String | option => any`. See below for exact details
- [x] `disabledBy: String | option => boolean` - if provided, should disable options, that return true.
- [x] `outputObject: Boolean` - when `true` - component returns full option(s). Otherwise valueBy is applied on option(s).

### `valueBy`
I'm assuming `valueBy` should function as follows: 
* If it's not defined, equality comparison should be performed. I.e. if optionA === optionB - then we consider options the same.
* If it's string and doesn't have dot symbol in its name - then we're working with options assuming they're objects. So optionA[trackBy] === optionB[trackBy] means options are the same.
* If it's a string and dot symbol is being used, then we perform nested search in object.
* If it's function than that form of comparison takes place: trackBy(optionA) === trackBy(optionB).

## Additional things

- [ ] We want to test this heavily, as the functionality is framework core. Good solution would be to provide a method, similar to `testIsFormComponent`, but `testIsSelectableList`. This will allow us to test mixin works properly, as well as ensure mixin is properly implemented in context of component.

misterpekert:
- Add SelectableListMixin
- Add testIsSelectableList helper to ensure component uses the mixin
- Add unit tests

Closes #192

### DIFF
--- a/src/components/vuestic-mixins/SelectableList/SelectableListMixin.js
+++ b/src/components/vuestic-mixins/SelectableList/SelectableListMixin.js
@@ -1,0 +1,66 @@
+import { FormComponentMixin } from '../FormComponent/FormComponentMixin'
+
+//Maybe we need to add some kind of fallback value to these helper methods
+//And to extract some in some kind of utility file 'cause I suppose they 
+//might become useful somewhere else at some point
+
+const getNestedValue = (option, propsArray) => {
+  const lastPropIndex = propsArray.length - 1
+  if (lastPropIndex < 0) return option
+
+  let nestedObj = option[propsArray[0]]
+
+  for (let i = 1; i < lastPropIndex; i++) {
+    if (nestedObj === null) {
+      return option
+    }
+    nestedObj = nestedObj[propsArray[i]]
+  }
+
+  if (nestedObj === null) return option
+
+  return nestedObj[propsArray[lastPropIndex]]
+}
+
+const getValueByPath = (option, prop) => {
+  if (option === null || typeof prop !=='string' || !prop) return option
+  if (option[prop] !== undefined) return option[prop]
+  prop = prop.replace(/^\./, '') //remove first point symbol
+  return getNestedValue(option, prop.split('.'))
+}
+
+const getProp = (option, prop) => {
+  if (!prop) return option
+  if (typeof prop === 'string') return getValueByPath(option, prop)
+  if (typeof prop === 'function') return prop(option)
+  return option
+}
+
+export const SelectableListMixin = {
+  mixins: [FormComponentMixin],
+  props: {
+    options: { type: Array, default: () => [] },
+    textBy: { type: [String, Function], default: 'text' },
+    valueBy: { type: [String, Function], default: 'value' },
+    trackBy: { type: [String, Function], default: 'value' },
+    disabledBy: { type: [String, Function], default: 'disabled' },
+    outputObject: { type: Boolean, default: false }
+  },
+  methods: {
+    getValue (option) {
+      return getProp(option, this.valueBy)
+    },
+    getText (option) {
+      return getProp(option, this.textBy)
+    },
+    getDisabled (option) {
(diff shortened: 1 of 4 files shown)


src/components/vuestic-mixins/testIsSelectableList.js
```diff
@@ -0,0 +1,11 @@
+import { shallowMount } from '@vue/test-utils'
+
+// Ensure that component is proper form component
```
asvae: Maybe just remove this comment.

src/components/vuestic-mixins/testIsSelectableList.js
```diff
@@ -0,0 +1,11 @@
...
+export function testIsSelectableList (componentOptions) {
```
asvae: Hmm. Not sure exactly how to improve, but that tests nothing right now.
Maybe you can test for testIsFormComponent here at least?

src/components/vuestic-mixins/SelectableList/SelectableListMixin.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    expect(wrapper.props().textBy).toEqual('text')
+    expect(wrapper.props().valueBy).toEqual('value')
+    expect(wrapper.props().trackBy).toEqual('value')
+    expect(wrapper.props().disabledBy).toEqual('disabled')
+    expect(wrapper.props().outputObject).toBeFalsy()
+  })
+
+  describe('getValue', () => {
+    it('should work with a string', () => {
+      const wrapper = prepareWrapper('valueBy', 'testValue')
```
asvae: Maybe just use destructuring instead? You won't need `beforeEach` in that case.
```
prepareWrapper({valueBy: 'testValue'})
```
Also, base component options it might make sense to define once before `describe`.

src/components/vuestic-mixins/SelectableList/SelectableListMixin.js
```diff
@@ -0,0 +1,66 @@
+import { FormComponentMixin } from '../FormComponent/FormComponentMixin'
+
+//Maybe we need to add some kind of fallback value to these helper methods
+//And to extract some in some kind of utility file 'cause I suppose they
```
asvae: Yeah. We have `utils` file. Please move it there + provide some test coverage.
Is this taken from vuetify? If yes - we can adapt their tests?

src/components/vuestic-mixins/SelectableList/SelectableListMixin.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        },
+      })
+
+      expect(actuallValue).toEqual(expectedValue)
+    })
+
+    it('should work with a function', () => {
+      const wrapper = prepareWrapper('valueBy', opt => opt.test)
+      const expectedValue = 'VALUE'
+      const actuallValue = wrapper.vm.getValue({ test: expectedValue })
```
asvae: actual :]

src/components/vuestic-mixins/testIsSelectableList.js
misterpekert: I thought it can be just a placeholder for now. Later in components which uses this mixin we might want to test them correctly using it, so this helper might be useful? 

I'll add check `isFormComponent` to `SelectableListMixin` spec file though

asvae: @misterpekert yeah, makes sense.

PR #200, opened 2020-03-23
Feat 190 added insufficient styles to css

### LINKED ISSUE #190: Slider has incorrect record for inactive line color
Should have same coloring as on left side:
image

SergeiPleshkov:
Fixes #190 incorrect slider track display

## Description

Added proper css, fixed default props.

## Types of changes

### DIFF
--- a/src/components/vuestic-components/va-slider/VaSlider.vue
+++ b/src/components/vuestic-components/va-slider/VaSlider.vue
@@ -210,7 +210,7 @@ export default {
       iconPrepend: { type: String, default: '' },
       iconAppend: { type: String, default: '' },
       vertical: { type: Boolean, default: false },
-      showTrack: { type: Boolean, default: false },
+      showTrack: { type: Boolean, default: true },
     }),
   ],
   data () {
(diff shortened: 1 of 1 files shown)


src/components/vuestic-components/va-slider/VaSlider.vue
```diff
@@ -210,7 +210,7 @@ export default {
       iconPrepend: { type: String, default: '' },
       iconAppend: { type: String, default: '' },
       vertical: { type: Boolean, default: false },
-      showTrack: { type: Boolean, default: false },
+      showTrack: { type: Boolean, default: true },
```
asvae: Nice catch!

src/components/vuestic-components/va-slider/VaSlider.vue
```diff
@@ -744,6 +744,12 @@ export default {
       position: absolute;
       border-radius: 0.25rem;
       transition: none;
+      opacity: 0.2;
+
+      &--active,
```
asvae: I think it would make sense to change naming a bit here.

We have
&__track - which is transparent one
&__track--inactive
&__track--active

From naming it's not apparent we have 3 states.

Instead we can use

&__track
&__track--active
&__track--background

That way
&__track &__track--active - active
&__track - inactive
&__track &__track--background - background

PR #202, opened 2020-03-25
Feat/slider bugs

### LINKED ISSUE #105: VaSlider bugs
- [x] on click and drag thumb should follow cursor without transition when initially clicked outside of thumb.
- [x] area for drag should be larger.

- [x] Keyboard navigation should allow left thumb to go past right thumb and vice-versa.
- [x] If keyboard navigation is used after thumbs are swapped by mouse - incorrect thumb is moving.

- [x] Can't get to the end of slider while using keyboard navigation when current value is between end and last pin.

- [x] On thumb swap value changes unpredictably.

- [x] Fix demo sizing (make them of decent width)
image

- [x] Disabled slider still works on click + uses drag cursor. Which shouldn't be the case.
image

- [x] Also please check this small problem: https://github.com/epicmaxco/vuestic-ui/pull/200/files#r396530884

## Refs
* [quasar](https://quasar.dev/vue-components/slider)

SergeiPleshkov:
Fixes #105 

## Description

Multiple slider bugs fixes. 
Keyboard navigation logic refactor.

## Types of changes

### DIFF
--- a/src/components/vuestic-components/va-slider/VaSlider.vue
+++ b/src/components/vuestic-components/va-slider/VaSlider.vue
@@ -224,6 +224,12 @@ export default {
     }
   },
   computed: {
+    moreToLess () {
+      return this.val[1] - this.step < this.val[0]
+    },
+    lessToMore () {
+      return this.val[0] + this.step > this.val[1]
+    },
     sliderClass () {
       return {
         'va-slider--active': this.hasMouseDown,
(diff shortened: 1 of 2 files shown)


src/components/vuestic-components/va-slider/VaSlider.vue
```diff
@@ -52,8 +52,8 @@
       <template v-if="isRange">
         <div
           ref="process"
-          class="container__track"
-          :class="{'container__track--active': hasMouseDown, 'container__track--inactive': !hasMouseDown}"
+          class="container__track--background"
```
asvae: Bem is very much off.

PR #203, opened 2020-03-25
Add OptionList Component

### LINKED ISSUE #88: OptionList component
# Docs

This component should act similarly to select, except it will use checkboxes and radio buttons instead of dropdown + list.

There could be some non streamlined spots in this issue, but you have artistic freedom to improve my oversights :).

## Radio Component
* Currently Radio component exists and is stand-alone. We can keep stand-alone behavior, but it takes low priority (no much compromises should be taken).
* Radio component is not FormComponent, but just a building piece

## OptionList Component
- [x] Should use `SelectableListMixin` (#192) and implement all of its functions (please also test: see `VaCheckbox.spec.js` -> `is Form Component` as example).

### Props
- [ ] `type: 'radio' | 'checkbox'` - determines options type.
- [x] `disabled: Boolean` - disables all radio buttons.
- [x] `readonly: Boolean` - sets all radio buttons to readonly
- [x] `default-value: String | Number` - default selected value 
- [x] `name: String | Number` - the name property of all radio buttons (you can probably use context to implement this seamlessly)
- [x] `color: String` - colors a selected radio component (use color theme mixin)
- [x] `left-label: Boolean` - add left-label prop to all radio buttons
- [x] `value: any` 

### Slots
- [x] `default`- we expect only va-radio components (radio button) will be put in this slot. They might have several wrappers though.

# Refs
* [Quasar](https://quasar.dev/vue-components/option-group)
* ~[Ant Design](https://vue.ant.design/components/radio/)~
* ~[iView](http://iview.talkingdata.com/#/components/radio-en)~
* ~[Element UI](https://element.eleme.io/#/en-US/component/radio)~

misterpekert:
- Add OptionList component
- Add tests
- Add some changes to SelectableListMixin
- Add left-label prop to VaCheckbox component
Closes #88

### DIFF
--- a/src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
+++ b/src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
@@ -5,25 +5,18 @@
     >
       <va-option-list
         :options="baseData.options"
+        v-model="baseData.selected"
       />
+      Selected {{ baseData.selected }}
     </VbCard>
     <VbCard
       title="Color options"
     >
       <va-option-list
         :options="baseData.options"
+        v-model="baseData.selected"
         color="danger"
       />
-    </VbCard>
-
-    <VbCard
-      title="With v-model"
-    >
-      <va-option-list
-        :options="baseData.options"
-        v-model="baseData.selected"
-      />
-      Selected {{ baseData.selected }}
     </VbCard>
     <VbCard
       title="With default value and v-model"
@@ -34,6 +27,16 @@
         :default-value="[baseData.options[0], baseData.options[2]]"
       />
       Selected {{ baseData.selectedWithVmodel }}
+    </VbCard>
+    <VbCard
+      title="With default value (when 1 default option array is not needed) and v-model"
+    >
+      <va-option-list
+        :options="baseData.options"
+        v-model="baseData.selectedWithVmodel1"
+        :default-value="baseData.options[0]"
+      />
+      Selected {{ baseData.selectedWithVmodel1 }}
     </VbCard>
     <VbCard
       title="Disable all"
@@ -50,6 +53,7 @@
     >
       <va-option-list
         :options="baseData.options"
+        v-model="baseData.selected"
         left-label
       />
     </VbCard>
@@ -82,14 +86,21 @@
       Selected <pre> {{ complexData.selectedObject }} </pre>
     </VbCard>
     <VbCard
+      title="Complex data with default value and v-model"
+    >
+      <va-option-list
+        :options="complexData.options"
+        :default-value="[complexData.options[0], complexData.options[2]]"
+        v-model="complexData.withDefault"
+      />
+      Selected <pre> {{ complexData.withDefault }} </pre>
+    </VbCard>
+
+    <VbCard
       title="Using slot"
     >
       <va-option-list
         :options="complexData.options"
-        v-model="complexData.selectedValueFromSlot"
-        :default-value="complexData.options[0]"
-        color="danger"
-        name="test-7"
       >
         <template v-slot="{ props }">
           <pre>{{ props }}</pre>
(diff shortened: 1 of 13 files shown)


src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <VbDemo>
+    <VbCard
+      title="Options as array of string"
+    >
+      <va-option-list
```
asvae: I suggest you provide a minimal demo.
In our case it would be v-model and options I believe.

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
...
+        v-model="baseData.selected"
+      />
+      Selected {{ baseData.selected }}
+    </VbCard>
+    <VbCard
+      title="Disable all"
+    >
+      <va-option-list
+        :options="baseData.options"
+        type="radio"
```
asvae: Can we use type `checkbox` as default?

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
...
+        color="primary"
+        type="radio"
+        v-model="baseData.selected"
+      />
+      Selected {{ baseData.selected }}
+    </VbCard>
+    <VbCard
+      title="Disable all"
+    >
+      <va-option-list
```
asvae: Please add v-model for here. That will allow us to test disabled works as intended and value is not emitted.

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
...
+        disabled
+      />
+    </VbCard>
+    <VbCard
+      title="Left label"
+    >
+      <va-option-list
+        :options="baseData.options"
+        type="radio"
+        color="primary"
```
asvae: You probably want to test colors separately.

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
...
+        left-label
+      />
+    </VbCard>
+    <VbCard
+      title="Readonly"
+    >
+      <va-option-list
+        :options="baseData.options"
+        :default-value="baseData.options[0]"
```
asvae: again, probably no much a reason to mash all props in one case.

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
```diff
@@ -0,0 +1,167 @@
...
+    >
+      <va-option-list
+        :options="complexData.options"
+        v-model="complexData.selectedValueFromSlot"
+        :default-value="complexData.options[0]"
+        color="danger"
+      >
+        <template v-slot="{ props }">
+          <va-radio
+            :option="props.option"
```
asvae: Maybe just show all props as object? `<pre>{{ props }}</pre>`. Up to you though.

src/components/vuestic-components/va-option-list/VaOptionList.vue
```diff
@@ -0,0 +1,111 @@
...
+  name: 'VaOptionList',
+  components: { VaRadio, VaCheckbox },
+  data () {
+    return {
+      selectedOptions: [],
+      selectedButton: '',
+    }
+  },
+  mixins: [
+    SelectableListMixin,
```
asvae: Please have contextable props in this mixin. Same as in FormComponentMixn.

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
misterpekert: Yep, `type: checkbox` is default in SelectableListMixin

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
misterpekert: It's not clear without color prop which radio button is selected. However, I can set `color: primary` as a default prop for OptionList component and show color changing in another card. Would that be ok?

src/components/vuestic-components/va-option-list/VaOptionList.demo.vue
misterpekert: I guess I have done exactly what you are saying.
We have the array of options and v-model changing upon selection. Or am I missing something?

src/services/utils.spec.js
```diff
@@ -97,4 +97,11 @@ describe('utils', () => {
       .toEqual(false)
     expect(deepEqual({ r: [circular] }, { r: [circular] })).toEqual(true)
   })
+  it('getProp', () => {
```
asvae: For `getProp` and `getValueByPath` would be nice to have a brief comment on their purpose.

PR #225, opened 2020-04-08
Rework pagination component:

### LINKED ISSUE #32: Pagination component
# Docs

- [ ] Do after [docs boilerplate](#17) is done.

## Mixins

- [x] Context
- [x] Stateful
- [x] Color

# Props
- [x] `total: Number` - total number of data items
- [x] `page-size: Number` - number of data items per page
- [x] `text-color: String` - use theme plugin for text
- ~`circle (*): Boolean` - shape pagination elements as circles~
- [x] `input: Boolean` - use an input instead of buttons
- [x] `boundary-numbers: Boolean` - always show first and last page buttons
- [x] `hide-on-single-page: Boolean` - whether to hide when there's only one page
- [x] `flat: Boolean` - shouldn't display border if set to true

# Events
- [x] `input` - emitted when the component needs to change the model;
- ~`size-change` - triggers when `page-size` changes~
- [ ] `next (*)` - emitted when going to next item
- [ ] `previous (*)` - emitted when going to previous item

# Design
- [x] Present to @xx13 after finishing.

# Notes
- [ ] Should work with `va-data-table`
- [x] Check button props

# Refs
* [Vuetify](https://vuetifyjs.com/en/components/paginations)
* [Quasar](https://quasar.dev/vue-components/pagination)
* [Ant Design](https://vue.ant.design/components/pagination/)
* [Element UI](https://element.eleme.io/#/en-US/component/pagination)

misterpekert:
- Add input support
- Add boundary numbers support
- Add possibility to set total and page-size alongside with pages

Closes #32

- [x] Confirm final result with @xx13 
- [ ] Add unit tests for core logic
- [x] Add TS (?)

### UI call updates
- [x] ~Add input mask~ Set value to current page + select it 
- [x] ~Add prop to control button width~ Add scss variable to control button width
- [x] Add color for ellipsis (inactive)
- [x] Remove `text-color` prop
- [x] Remove top border when `input` + `flat`
- ~[ ] Add `show-total` function to render total label~ User can do this itself


packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -130,6 +134,10 @@ const InputContextMixin = makeContextablePropsMixin({
     type: Number,
     default: 0,
   },
+  insidePagination: {
```
asvae, line 137: This prop doesn't make a lot of sense for user. Can't we just style input from pagination component?

asvae: Actually, do we need input component in pagination? Maybe just input is good enough? None of its advanced features are being used.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+    />
     <va-button
-      v-if="this.pages !== visiblePages && directionLinks"
-      :style="directionButtonStyle"
+      v-if="showDirectionLinks"
       outline
       :color="color"
       :size="size"
-      :disabled="disabled || value === this.pages"
+      :disabled="disabled || value === lastPage"
```
asvae, line 65: Some context handling would be great. Currently there is none.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+  iconSet: { type: Object, default: null },
+  iconSetRight: { type: Object, default: null },
+  color: { type: String, default: 'primary' },
+})
+
+const mixinsArr = [
+  ContextPluginMixin,
+  StatefulMixin,
+  ColorThemeMixin,
+  paginationPropsMixin,
```
asvae, line 124: Capital "P".

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+import VaInput from '../va-input/VaInput.vue'
+import { StatefulMixin } from '../../vuestic-mixins/StatefullMixin/StatefulMixin'
 import { setPaginationRange } from './setPaginationRange'
-import { ContextPluginMixin, getContextPropValue } from '../../context-test/context-provide/ContextPlugin'
+import { ContextPluginMixin, makeContextablePropsMixin } from '../../context-test/context-provide/ContextPlugin'
+import Component, { mixins } from 'vue-class-component'
+import { Watch } from 'vue-property-decorator'
+import { ColorThemeMixin } from '../../../services/ColorThemePlugin'
+
+interface IconSet {
```
asvae, line 94: What's the point of this? Why not just use props with left and right icons?

Also IconSet is reserved.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+
+  @Watch('pageSize')
+  onPageSizeChange (newPageSize: number, oldPageSize: number) {
+    let { valueComputed } = (this as any)
+    const newPage = Math.ceil(((valueComputed - 1) * oldPageSize + 1) / newPageSize)
+    valueComputed = newPage
+  }
+
+  private mounted () {
+    this.useTotal = !!((this as any).total && (this as any).pageSize)
```
asvae, line 196: Probably should be computed.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
 <style lang='scss'>
 @import "../../vuestic-sass/resources/resources";
 
 .va-pagination {
-  .va-button--disabled {
-    i {
-      color: $brand-secondary;
-    }
+  .va-input.va-button,
+  .va-button--ellipsis {
```
asvae, line 254: Button shouldn't know about pagination.

packages/ui/src/components/vuestic-components/va-pagination/setPaginationRange.ts
```diff
@@ -14,10 +14,16 @@ export const setPaginationRange = (value, visiblePages, pages) => {
       : Math.ceil(value - paginationMiddlePage)
   }
 
-  const range = []
+  const range: Array<string | number> = []
```
asvae, line 17: `(number | '...')[]` might suit a bit better.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
-  computed: {
-    directionButtonStyle () {
+    (this as any).valueComputed = pageNum
+  }
+
+  private resetInput () {
+    (this as any).$refs.input.reset()
+  }
+
+  private changeValue () {
```
asvae, line 214: onUserInput? changeValue doesn't sound too different from `changePage` to me.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+  private get showPagination () {
+    return this.lastPage > 1 || (!(this as any).hideOnSinglePage && this.lastPage <= 1)
+  }
+
+  private get fontColor () {
+    return (this as any).computeColor((this as any).color)
+  }
+
+  @Watch('pageSize')
+  onPageSizeChange (newPageSize: number, oldPageSize: number) {
```
asvae, line 189: This does nothing.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+  private get lastPage () {
+    const { total, pageSize, pages } = (this as any)
+    return this.useTotal
+      ? Math.ceil(total / pageSize)
+      : pages
+  }
+
+  private get paginationRange () {
+    const { valueComputed, visiblePages, total, pageSize, boundaryNumbers, pages } = (this as any)
+    return this.useTotal
```
asvae, line 166: For total/pages we may want to:
* add error for case when user specifies both (as that shouldn't be possible).
* for total mode - value should be items, not pages.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
```diff
@@ -1,215 +1,258 @@
...
+      <va-button
+        :style="activeButtonStyle(n)"
+        outline
+        v-for="(n, key) in paginationRange"
+        :key="key"
+        :color="color"
+        :size="size"
+        :disabled="disabled || n === '...'"
+        :class="{
+          'va-button--active': n === value,
```
asvae, line 36: Here's how I would name it:

`va-pagination__va-button--active`
Normally we use component name as bem block only in component context.

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.demo.vue
```diff
@@ -63,7 +63,7 @@
                 v-model="activePage"
```
asvae, line 63: Context and stateful demo would be great.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -269,6 +288,12 @@ export default {
     reset () {
       this.$emit('input', '')
     },
+    selectAll (value) {
+      this.$emit('input', value)
```
asvae, line 292: Pretty sure select should not emit anything. User actions do.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -188,14 +196,25 @@ export default {
...
-        borderColor:
-          this.computedError ? this.computeColor('danger')
-            : this.c_success ? this.computeColor('success')
-              : this.isFocused ? this.computeColor('dark') : this.computeColor('gray'),
+      if (!this.insidePagination) {
+        return {
+          minHeight: '2.375rem',
+          backgroundColor:
+            this.computedError ? (this.computeColor('danger') ? getHoverColor(this.computeColor('danger')) : '')
+              : this.c_success ? (this.computeColor('success') ? getHoverColor(this.computeColor('success')) : '') : '#f5f8f9',
```
asvae, line 204: Can we replace hex here with something less cryptic?

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -188,14 +196,25 @@ export default {
...
+          borderColor:
+            this.computedError ? this.computeColor('danger')
+              : this.c_success ? this.computeColor('success')
+                : this.isFocused ? this.computeColor('dark') : this.computeColor('gray'),
+        }
+      } else {
+        const isPaginationFlat = this.$parent.$parent.flat // not really good :(
+        return {
+          borderColor: this.computeColor(this.color),
+          borderWidth: isPaginationFlat ? '0 0 2px' : '2px 0',
```
asvae, line 214: We might want to use scss to handle some of these.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -188,14 +196,25 @@ export default {
...
+          backgroundColor:
+            this.computedError ? (this.computeColor('danger') ? getHoverColor(this.computeColor('danger')) : '')
+              : this.c_success ? (this.computeColor('success') ? getHoverColor(this.computeColor('success')) : '') : '#f5f8f9',
+          borderColor:
+            this.computedError ? this.computeColor('danger')
+              : this.c_success ? this.computeColor('success')
+                : this.isFocused ? this.computeColor('dark') : this.computeColor('gray'),
+        }
+      } else {
+        const isPaginationFlat = this.$parent.$parent.flat // not really good :(
```
asvae, line 211: If placed in app root - this will break horribly.

There should be ways to play around that.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -49,7 +49,11 @@
           :id="id"
           :name="name"
           class="va-input__container__input"
-          :style="{ paddingBottom: c_label ? '0.125rem' : '0.875rem' }"
+          :style="{
+            paddingBottom: c_label ? '0.125rem' : '0.875rem',
+            textAlign: insidePagination ? 'center': 'inherit',
```
asvae, line 54: This definitely can be handled via scss.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -49,7 +49,11 @@
           :id="id"
```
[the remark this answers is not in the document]
misterpekert: Yes, it is. @xx13 what do you think?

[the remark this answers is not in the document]
misterpekert: You think we should have smth like this: `< 1 ... 3 4 5 6 7 ... 20 >` in this case?

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
misterpekert: It's a watcher for pageSize prop. And it looks like it's working :)

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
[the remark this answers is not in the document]
misterpekert: done.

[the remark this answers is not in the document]
misterpekert: done.

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
misterpekert: done

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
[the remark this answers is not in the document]
misterpekert: [ ] Set visible-pages default to 0 (display all pages by default)

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
misterpekert: Make 4 props for icons

packages/ui/src/components/vuestic-components/va-pagination/VaPagination.vue
misterpekert: Make as computed prop

asvae:
Thanks!

PR #295, opened 2020-05-15
Feat/switch stateful fix

### LINKED ISSUE #281: Stateful switch doesn't work
image
http://localhost:8080/#/demo/vuestic-components/va-switch/VaSwitch.demo.vue

Kushich:
## Description
closes #281

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
@@ -166,7 +166,7 @@
     <VbCard title="Stateless switch without v-model">
       <va-switch />
     </VbCard>
-    <VbCard title="Stateful" style="border: 4px solid red;">
+    <VbCard title="Stateful">
       <va-switch stateful />
     </VbCard>
   </VbDemo>
--- a/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
@@ -167,7 +167,7 @@
       <va-switch />
     </VbCard>
     <VbCard title="Stateful">
-      <va-switch stateful />
+      <va-switch stateful/>
     </VbCard>
   </VbDemo>
 </template>
--- a/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.spec.js
+++ b/packages/ui/src/components/vuestic-components/va-switch/VaSwitch.spec.js
@@ -1,11 +1,7 @@
-import Vue from 'vue'
 import { shallowMount } from '@vue/test-utils'
 import VaSwitch from './VaSwitch'
 
-import { ColorThemePlugin } from '../../../services/ColorThemePlugin'
 import { testIsSelectableComponent } from '../../vuestic-mixins/testIsSelectableComponent'
-
-Vue.use(ColorThemePlugin)
 
 describe('VaSwitch', () => {
   it('default', () => {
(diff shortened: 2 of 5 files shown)


packages/ui/src/components/vuestic-mixins/SelectableComponent/SelectableMixin.js
```diff
@@ -73,21 +82,21 @@ export const SelectableMixin = {
         return
       }
       if (this.modelIsArray) {
-        if (!this.value) {
-          this.$emit('input', [this.c_arrayValue])
-        } else if (this.value.includes(this.c_arrayValue)) {
-          this.$emit('input', this.value.filter(option => option !== this.c_arrayValue))
+        if (!this.c_value) {
+          this.valueComputed = [this.c_arrayValue]
```
asvae, line 86: Couple of things I want to point here and in some other places.

1) The way `valueComputed` and statefullness is intended to be used is that you don't use `c_value` at all! You always interact with `valueComputed` as well as perform checks on it.
But:
2) Array access + stateful makes no sense. So for array access we pretend `computedValue` does not exist and use `c_value` + emit input directly.

To surmise:
* for array access use `c_value`.
* otherwise use `valueComputed`

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.spec.js
```diff
@@ -0,0 +1,26 @@
+import Vue from 'vue'
+import { shallowMount } from '@vue/test-utils'
+import VaSwitch from './VaSwitch'
+
+import { ColorThemePlugin } from '../../../services/ColorThemePlugin'
```
asvae, line 5: Component should work without color theme plugin. Ideally we want to test without it.

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
```diff
@@ -166,7 +166,7 @@
     <VbCard title="Stateless switch without v-model">
```
asvae, line 166: padding
image

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.spec.js
```diff
@@ -0,0 +1,22 @@
+import { shallowMount } from '@vue/test-utils'
```
asvae, line 1: Please add context test (see `VaRating.spec.js > is contextable component`)

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.vue
```diff
@@ -265,19 +255,46 @@ export default {
...
+      &__input-wrapper {
+        transform: translateX(100%);
+      }
+    }
+  }
+
+  &--error {
+    .va-switch {
+      &__label {
+        color: $theme-danger;
```
asvae, line 283: js color

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.demo.vue
asvae, line 166: On keyboard navigation I was able to land focus on disabled switch. That should not be the case.
image

asvae, line 166: Should have cursor pointer on label

packages/ui/src/components/vuestic-components/va-switch/VaSwitch.vue
```diff
@@ -287,20 +304,57 @@ export default {
...
+      .va-switch--large#{&} {
+        border-radius: 1.25rem;
+      }
     }
   }
 
-  &__track-label {
+  #{&}__track-label {
     color: $white;
+    margin: auto 7px auto 33px;
```
asvae, line 323: can't we do rems?
Also in other places?
Ideally we want rems except for cases, where we can't do rems.

asvae:
Looks good. Thanks.

PR #314, opened 2020-05-21
feat(310): Add avatar docs

### LINKED ISSUE #310: Docs for va-avatar
Add docs for va-avatar component

misterpekert:
closes #310


packages/docs/components/page-configs/va-avatar/api-options.ts
```diff
@@ -0,0 +1,26 @@
+import { ManualApiOptions } from 'vuestic-ui/src/services/api-docs/ManualApiOptions'
```
asvae, line 1: We don't need empty avatar in examples at all as it serves no purpose in app.
image
So I recommend to either remove or replace examples with content-full avatars.

asvae:
Looks good. Small fix is needed though.
Then just merge.

PR #321, opened 2020-05-22
Docs for Button component

### LINKED ISSUE #187: Button: incorrect relative positions for buttons with icons
[WARNING: Harder than it looks]

### Expected behavior

Shift **should not** occur in all cases.

### Current behavior

We're speaking of 2 buttons, that are positionned inline one after other.
* If one button has 2 icons, and other button has 1 icon - shift occurs.
* If one button has 1 icon, and other has none - shift occurs.
* If one button has 2 icons, and other button has none - shift occurs.
* If both button have same icons in same positions - shift does not occur.

### Steps to reproduce
http://vuestic-ui-demo.sub.asva.by/#/demo/vuestic-components/va-button/VaButton.demo.vue
image

### Other information
* We already spent some time trying to fix it. Tried different paddings, display modes etc without much success. Still not exactly clear as to what's causing this.
* I think the simplest way to fix this would be to take heavy inspiration from [vuetify button](https://vuetifyjs.com/en/components/buttons/) as their implementation is very close to our own.

### LINKED ISSUE #316: Docs for Button component
Add docs for button component
- [x] Description
- [ ] Examples (Color, Size, Styles (flat, outlined, square), With icons)
- [ ] API

misterpekert:
Closes #316 and closes #187

### DIFF
--- a/packages/docs/components/page-configs/va-button/api-options.ts
+++ b/packages/docs/components/page-configs/va-button/api-options.ts
@@ -17,6 +17,6 @@ export default {
     default: { local: true },
   },
   events: {
-    click: { local: true, types: '`() => void`' },
+    click: { local: true, types: '`() => Event`' },
   },
 } as ManualApiOptions
--- a/packages/docs/components/page-configs/va-button/api-options.ts
+++ b/packages/docs/components/page-configs/va-button/api-options.ts
@@ -1,0 +1,22 @@
+import { ManualApiOptions } from 'vuestic-ui/src/services/api-docs/ManualApiOptions'
+
+export default {
+  props: {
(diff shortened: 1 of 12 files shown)


packages/docs/components/page-configs/va-button/api-options.ts
```diff
@@ -0,0 +1,22 @@
...
+    block: { local: true },
+    round: { local: true },
+    href: { local: true },
+    targer: { local: true },
+  },
+  slots: {
+    default: { local: true },
+  },
+  events: {
+    click: { local: true, types: '`() => void`' },
```
asvae, line 20: Can we make it return event?

PR #330, opened 2020-05-26
Fix issue with avatar font size

misterpekert:
- rename prop inside SizeMixin
- Add text explaining font-size prop behavior in avatar component


asvae:
All good. Thanks.

PR #332, opened 2020-05-27
Feat/image docs

### LINKED ISSUE #329: VaImage docs

SergeiPleshkov:
Closes #329 

## Types of changes


packages/docs/examples/va-image/Default.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="row">
+    <div class="flex md6 lg4">
+      <va-image src="https://picsum.photos/1500"/>
+    </div>
+  </div>
+</template>
+
+<script>
```
asvae, line 9: This part is not needed?

packages/docs/locales/en/en.json
```diff
@@ -293,6 +294,21 @@
         "outlined": "Toggles shadow of `va-card`"
       }
     },
+    "VaImage": {
+      "props": {
+        "ratio": "Width to height ratio.",
+        "contain": "Background size. If false, results to `cover`"
```
asvae, line 300: Doesn't tell me much.
In fact, obfuscates, that `background-size`  is a css property.
Imagine user doesn't know what background-size means by heart and just wants to solve some particular problem.

packages/docs/examples/va-image/Slots.vue
```diff
@@ -0,0 +1,60 @@
+<template>
```
asvae, line 1: We probably want 3 examples instead of this one.

packages/ui/src/components/vuestic-components/va-image/VaImage.vue
```diff
@@ -26,90 +26,86 @@
...
-      this.destroyLoader()
-      this.loadingError = true
-      this.loading = false
-      this.$emit('error', err)
-    },
-  },
+  }
+
+  @Watch('src')
+  onSrcChange () {
```
asvae, line 62: where is immediate?

SergeiPleshkov: Seems like it works correct without immediate, but have memory leak with it set to true.

asvae:
Thanks. Looks good.

PR #333, opened 2020-05-27
Feat/va progress bar docs

raichev-dima:
Should close https://github.com/epicmaxco/vuestic-ui/issues/326

## Types of changes


packages/docs/examples/va-progress-bar/Coloring.vue
```diff
@@ -0,0 +1,21 @@
...
+    <va-progress-bar
+      value="35"
+      color="purple"
+    />
+    <div class="mb-4" />
+  </div>
+</template>
+
+<style scoped>
+.muted {
```
asvae, line 18: We really do not want to introduce any unneded styling to examples.
Maybe move to global class?

packages/docs/examples/va-progress-bar/Default.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <div class="flex lg6 xs12">
+    <div class="mt-2 mb-5">
+      <div class="my-2 text--bold muted">No value:</div>
```
asvae, line 4: We probably don't need no value example.

packages/docs/examples/va-progress-bar/Slots.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <div class="flex lg6 xs12">
+    <div class="mt-2 text--bold muted">Dynamic value:</div>
```
asvae, line 3: Ideally we do not want to introduce any complicated texts in examples as they're no translated. So might be tricky for some users to understand.
Recommended approach - split into smaller examples or separate by prop values.

I.e.

```
## Sizes
`small` [ small example ]
`large` [ large example ]
```

is  all right, but

```
small component preset [ small example ]
large component preset [ large example ]
```

is not.

packages/docs/examples/va-progress-bar/Coloring.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div class="flex lg6 xs12">
+    <div class="my-2 text--bold muted">Danger color:</div>
```
asvae, line 3: I recommend we skip description here as well.

packages/docs/examples/va-progress-bar/Sizing.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div class="flex lg6 xs12">
+    <div class="my-2 text--bold muted">Size "small":</div>
```
asvae, line 3: Here we probably want to just write 'small', '25px' etc. Let's leave to user figuring out what the example is about.

packages/docs/examples/va-progress-circle/Thickness.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <div class="flex lg6 xs12">
+    <div class="my-2 text--bold muted">Thickness 0.01:</div>
```
asvae, line 3: Barely visible. Maybe remove this one and add '1.0'

asvae:
Overall good. Some small points in regards to examples. Please merge after fixes.

PR #341, opened 2020-05-30
(fix) language selector

### LINKED ISSUE #338: Language selector works incorrectly
After each change of language url is updated correctly. It should have language code which corresponds to selected language.
URL is changing only when Spanish language is selected (/es/)

Kushich:
closes #338

## Description
We passed 'gb' value instead of 'en' to the vue router and router didn't recognise this value in locales folder. Fix replaces 'gb' with 'en'

## Types of changes


asvae:
How does that solve a problem?
(problem being `"gb"` was being used instead of `"en"`)

asvae:
All good. Thanks.

PR #347, opened 2020-06-03
Feat/loading mixin

### LINKED ISSUE #302: LoadingMixin
- [ ] Create `LoadingMixin.ts`.
- [ ] All component that support loading prop should be using this mixin.
- [ ] In same folder as loading mixin - create file called LoadingMixin.md. In there describe that:
* Loading state for component should always be taken from this prop alone, we do not make guesses on whether component is loaded based on value. 
* There could be some cases where couple of loaders for single component would make sense, such as select with infinite scroll. We want to give user a slot + example in docs in that case.
- [ ] Ensure loading prop description takes translation from `all`.
- [ ] create `testIsLoadingMixin` and test all relevant components.

Kushich:
closes #302

## Types of changes

### DIFF
--- a/packages/docs/components/page-configs/va-avatar/api-options.ts
+++ b/packages/docs/components/page-configs/va-avatar/api-options.ts
@@ -11,9 +11,6 @@ export default {
     fontSize: {
       local: true,
     },
-    loading: {
-      local: true,
-    },
     email: {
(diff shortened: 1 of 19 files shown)


asvae:
Excellent. Thanks.

PR #349, opened 2020-06-04
(fix) slider transition

### LINKED ISSUE #348: Slider transition should follow thumb
Currently thumb and track look separated, which shouldn't be the case:

Kushich:
#closes #348 

d3ST0tTao7

## Types of changes


asvae:
Small problem here.
When user clicks on track - thumb should move there instantly (you can compare with ).

So we need 2 behaviors:
1) When user is interacting with thumb all movement should be instant.
2) When component reacts to external change - movement should go with transition.

asvae:
Works nicely.

PR #350, opened 2020-06-05
feat/ docs manual type

### LINKED ISSUE #345: API docs should allow manual type for props
image

Here's the case where vue native types do not support intent we want to cover.
In this case type should be `() => Promise<any>`, as `Function` doesn't tell much.

- [ ] Add manual options support for props
- [ ] If manual options for prop is not defined, value should be taken from component as before.

Kushich:
closes #345 

## Description
chrome_vBH5InJ8Bw

## Types of changes


asvae:
Looks good. Thanks.

Issue #152, opened 2020-01-12
Button dropdown issues


asvae:
* [original issue](https://github.com/epicmaxco/vuestic-admin/issues/438)

1) Small padding problem
![image](https://user-images.githubusercontent.com/5394573/72225433-86bb2680-3596-11ea-8de6-23539aa234c1.png)

2) Icon size is inconsistent. Should be same for same button size.
![image](https://user-images.githubusercontent.com/5394573/72225441-9b97ba00-3596-11ea-9445-7a752f2c6530.png)

PR #371, opened 2020-06-28
Fix: VaButtonDropdown class-style refactor & alignment fix

### LINKED ISSUE #368: VaButtonGroup/VaButtonDropdown can't handle button size
Components have no size props

Components refers on large/small (boolean) props

Pass props into the components

* **Other information** 
Component should refers to size ('small', 'medium', 'large') prop like VaButton.

untael:
Closes #368, #152

## Types of changes


packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.vue
```diff
@@ -1,166 +1,145 @@
...
+    return this.showDropdown ? resultedIcon : this.c_icon
+  }
+
+  get computedClass () {
     return {
-      visible: false,
+      'va-button-dropdown': true,
+      'va-button-dropdown--split': this.c_split,
+      [`va-button-dropdown--${this.c_size}`]: true,
     }
```
asvae, line 122: Split into 2 classes with conditions. Generally we do not want to use string composition for css classes as that complicates search.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -22,9 +22,9 @@
         class="va-dropdown__anchor-width-container"
         :style="anchorWidthContainerStyles"
       >
-        <slot />
+        <slot/>
       </div>
-      <slot v-else />
+      <slot v-else/>
```
asvae, line 27: can you configure ide in a way so that it styles self closing tags the way they are now? Alsom can you check if linter can autofix these?

untael: eslint config was extended

PR #376, opened 2020-07-07
Feat/va-button public methods (blur/focus)

### LINKED ISSUE #361: Add `focus` and `blur` to VaButton
- [ ] `focus` public method
- [ ] `blur` public method
- [ ] demos, docs, tests
- [ ] check if it works for other components fix them in context of this issue as well.

untael:
Closes #361

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-button/VaButton.vue
+++ b/packages/ui/src/components/vuestic-components/va-button/VaButton.vue
@@ -20,7 +20,6 @@
     @focus="updateFocusState(true)"
     @blur="updateFocusState(false)"
     :tabindex="c_loading ? -1 : 0"
-    ref="button"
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/vuestic-components/va-button/VaButton.vue
```diff
@@ -20,6 +20,7 @@
     @focus="updateFocusState(true)"
     @blur="updateFocusState(false)"
     :tabindex="c_loading ? -1 : 0"
+    ref="button"
```
m1mzz, line 23: you could use $el

untael: thanks

PR #426, opened 2020-09-28
docs/toast

### LINKED ISSUE #354: Toast Component
Current State is that we're using `vue-toasted` library. That's not OK as it's not modular enough. We also want to have ownership for any, but too complicated components.

Features analysis: [google spreadsheet](https://docs.google.com/spreadsheets/d/1rOB9fOPQN1-06KKti542a1QtCeKnH81mR6Y1OyTGTAI/edit?usp=sharing)

Analysis points:
* Primarily we're interested in service as that's the most common usecase.
* Ant implementations have Message and Notification, which look similar, but have slightly different purpose. I think we want to support only one way, but make it configurable enough to cover both cases.

## Other
- [ ] Documentation
- [ ] Tests

# Refs
* [vue-toasted](https://shakee93.github.io/vue-toasted/) - not well polished or sophisticated enough.
* [Vuetify](https://vuetifyjs.com/en/components/snackbars/) - out of all solutions, only vuetify has component :/, others are service only.
* [Quasar](https://quasar.dev/quasar-plugins/notify) - Has a lot of props an these are intended to allow customization when there's no options to pass component/render. I guess they chose that route to support mobile toasts.
* [ant Message](https://www.antdv.com/components/message/) [Ant Notification](https://www.antdv.com/components/notification/) - both are similar, but have slightly different purpose. I think we want to support only one way, but make it configurable enough to cover both cases. You can think of Message as of lightweight Notification.
* [element Message](https://element.eleme.io/#/en-US/component/message) [element Notification](https://element.eleme.io/#/en-US/component/notification)
* [iview Message](https://www.iviewui.com/components/message-en) [iview Notice](https://www.iviewui.com/components/notice-en)

Kushich:
close #354 

## Types of changes


packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
```diff
@@ -52,12 +52,12 @@ const ToastPropsMixin = makeContextablePropsMixin({
...
   offsetX: { type: Number, default: 16 },
   message: { type: [String, Function], default: '' },
-  iconClass: { type: String, default: '' },
+  icon: { type: String, default: 'close' },
   customClass: { type: String, default: '' },
   duration: { type: Number, default: 20000 },
   color: { type: String, default: '' },
   closeable: { type: Boolean, default: true },
-  dangerouslyUseHTMLString: { type: Boolean, default: false },
+  // dangerouslyUseHTMLString: { type: Boolean, default: false },
```
asvae, line 60: Remove maybe? Looks weird.

packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
```diff
@@ -17,8 +17,8 @@
         <h2 v-if="title" class="va-toast__title" v-text="title"></h2>
         <div class="va-toast__content" v-show="message">
           <slot>
-            <p v-if="!dangerouslyUseHTMLString" v-text="message"></p>
-            <p v-else v-html="message"></p>
+            <!-- <p v-if="!dangerouslyUseHTMLString" v-text="message"></p> -->
+            <p v-html="message"></p>
```
asvae, line 21: v-html shouldn't be default. Too dangerous.

packages/docs/locales/en/en.json
```diff
@@ -639,12 +639,30 @@
...
       "slots": {
         "default": "Slot for backtop content"
       }
+    },
+    "VaToast": {
+      "props": {
+        "title": "Sets the title for the notification",
+        "offsetY": "Sets the Y offset",
+        "offsetX": "Sets the X offset",
+        "message": "Notification message",
```
asvae, line 654: Either html or component handling is mandatory. That's a common case.

packages/docs/locales/en/en.json
```diff
@@ -1872,5 +1891,35 @@
         "text": "By default a `va-backtop` watches for body element offset height. You can change target element, positioning and scrolling speed."
       }
     }
+  },
+  "toast": {
+    "title": "Toast",
```
asvae, line 1896: We want at least several lines on how service is different from component. Maybe in QA below the page.

packages/docs/locales/en/en.json
```diff
@@ -1872,5 +1891,35 @@
...
+        "title": "Offset",
+        "text": "Use `offset` property to set the offset of the toast."
+      },
+      "position": {
+        "title": "Position",
+        "text": "Use `position` property to set the custom position of the toast. Available are `top-right`, `top-left`, `bottom-right`, `bottom-left`."
+      },
+      "close": {
+        "title": "Close",
+        "text": "You can use close methods to close the notification and you can set custom onClose event."
```
asvae, line 1917: close method?

packages/docs/components/page-configs/va-toast/api-options.ts
```diff
@@ -0,0 +1,24 @@
...
+    title: { local: true },
+    offsetY: { local: true },
+    offsetX: { local: true },
+    message: { local: true },
+    icon: { local: true },
+    customClass: { local: true },
+    duration: { local: true },
+    closeable: { local: true },
+    // dangerouslyUseHtmlString: { local: true },
+    onClose: { local: true },
```
asvae, line 14: Custom onClose handler didn't really work for me.

packages/docs/examples/va-toast/Close.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <div>
+    <va-button @click="$vaToast.init({
+      message: 'Notification with onClose handler',
+      onClose: () => closeHandler
```
asvae, line 5: That doesn't seem right, also doesn't work. `() => closeHandler()`?

packages/docs/components/page-configs/va-toast/api-options.ts
```diff
@@ -0,0 +1,24 @@
+import { ManualApiOptions } from 'vuestic-ui/src/services/api-docs/ManualApiOptions'
```
asvae, line 1: Duration 20000 seems like an overkill to me.
Compare with refs.

asvae (changes requested):
Docs are mostly fine.
But there are issues with original implementation.

packages/docs/examples/va-toast/Close.vue
Kushich: Sorry, I fixed it but forgot to push :)

packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
```diff
@@ -17,8 +17,7 @@
         <h2 v-if="title" class="va-toast__title" v-text="title"></h2>
```
asvae, line 17: Component is done poorly:
- [ ] `window.setTimeout(` -> `setTimeout`
- [ ] `Reset timer if user press Delete or Backspace. I'm not sure we need this` - remove or figure it out.
- [ ] `onClose` handler is super buggy. Should be triggered only on manual click. Is being triggered on any close.
- [ ] Why do we have both `visible` and `closed` in component?
- [ ] Should support component or render (see [doc](https://docs.google.com/spreadsheets/d/1rOB9fOPQN1-06KKti542a1QtCeKnH81mR6Y1OyTGTAI/edit#gid=0)). Actually it seems to be supported from
image, so maybe just docs + demo issue.
- [ ] `options.message = 'REPLACED_BY_VNODE'` looks weird. Can we remove that?

packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
```diff
@@ -40,29 +39,30 @@
...
   closeable: { type: Boolean, default: true },
-  dangerouslyUseHTMLString: { type: Boolean, default: false },
-  onClose: { type: [Function as PropType<() => void>, undefined] },
-  onClick: { type: [Function as PropType<() => void>, undefined] },
+  onClose: { type: Function, default: undefined },
+  onClick: { type: Function, default: undefined },
   position: {
-    type: String as PropType<NotificationPosition>,
+    type: String,
     default: 'top-right',
```
asvae, line 62: ideally we also want all options in api docs

packages/docs/components/page-configs/va-toast/api-options.ts
```diff
@@ -0,0 +1,23 @@
...
```
asvae, line 1: In demo this doesn't work properly
image

I'm not seeing how that could be useful for end user.
Instead can we return said id on notification create.

asvae, line 1: @untael would be greate if we can implement it in vue 3, as I'm pretty sure there will be issues.
Also I might have skipped a bug or two, so ~~keep your eyes open~~ remember to wipe your space helmet.

asvae (changes requested):
A couple of things are not up to standards.

packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
damirJa: I think it's ok if we could stay `onClose` handler on any close, not only after the manual close. Of course, we could provide a special variable to show if notification has been closed manually or not, but think it's not very helpful for end-users

damirJa: I would like to remove all `slots` things in here? Render method it looks more powerful and easy to use, can I?

packages/docs/components/page-configs/va-toast/api-options.ts
damirJa: I want to clarify this point.
I understanding it something like this: 
`
const toast = $vaToast.init('Hello')
toast // is unique id
`
Is it correct?

asvae: @damirJa yeah. That would work.

packages/ui/src/components/vuestic-components/va-toast/VaToast.vue
asvae: @damirJa yeah. Render is fine.

asvae (approved):

Issue #325, opened 2020-05-25
Themes 2.0


asvae:
A collection of ideas to improve the current state of varibles/themes/customizability.
Excellent theme support is meant to be key point of 1.0 release so we want to focus it heavily.

Sources of inspiration:
## Refs 
* [vuetify](https://github.com/vuetifyjs/vuetify/tree/master/packages/vuetify/src/styles) - they have widespread use of css variables, which is something we strive to do. The issue is with exact point of application for these.
* [bulma](https://github.com/jgthms/bulma) - our system works similarly to bulma in a sense that styles are not applied to tags by default, but only when you wrap them in `.content`.

## Docs
* [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

## Ideally we want theme system to be:
* Dynamic. i.e. for specific page or even block in page there should be a way to define theme.
* Smart. User should get result with minimal number of changes.
* Powerful. We should allow to modify styles to significant degree from a number of angles. We should also use low priority selectors when possible.

## As for implementation:
- ![red](https://via.placeholder.com/15/BB0000/000000?text=+) [**Global Config**](https://github.com/epicmaxco/vuestic-ui/issues/293). We have a way to set global config, but it doesn't have good interface. Colors and context are split. Docs are missing.
- ![yellow](https://via.placeholder.com/15/FFFF00/000000?text=+) **Components support**. Components support color well enough and we have a way to change implementation. There are still some issues we want to address, such as adapting font color to background color etc.
- ![yellow](https://via.placeholder.com/15/FFFF00/000000?text=+) **Context system**. Docs are missing. We also have a problem with nested configs. I.e. for flat props - context system works excellent. But if you interact with complicated configs, such as color configs or icons, then it doesn't present itself well enough. I.e. we might want to merge configs from parent, which is not possible at the moment.

## With some details
Sketch: https://www.figma.com/file/nrR3CNZZSvUsQwpUYVbQLG/Untitled?node-id=0%3A1
- [x] At highest level we want to have one global config (reactive)
- [x] User will pass default config via vue-use.
- [x] In all config there should be color options.
- [x] If you don't need merge - you may pass function.
- [x] Should have debounce on color change. (low priority)
- [ ] Should have recommendation paragraph in docs: it is recommended to change config via debounce if user can change it multiple times per second. Some computation happens on change. (low priority)
- [ ] Color config should have type declaration. ()
[...]

PR #437, opened 2020-10-07
Feat/theme 2.0

raichev-dima:
Should cover https://github.com/epicmaxco/vuestic-ui/issues/325 partially:

- [x] At highest level we want to have one global config (reactive)
- [x] User will pass default config via vue-use.
- [x] In all config there should be color options.
- [x] If you don't need merge - you may pass function.
- [ ] Should have debounce on color change. (low priority)
- [ ] Should have recommendation paragraph in docs: it is recommended to change config via debounce if user can change it multiple times per second. Some computation happens on change. (low priority)
- [ ] Color config should have type declaration. ()
- [x] Vuestic plugin should include context plugin.
- [x] ContextPlugin should be independent.
- [x] Components should not require ContextPlugin to function.
- [ ] Check vuetify defaults for different components colors. ('gray', '#333333', 'primary')
- [x] All components should use theme colors for defaults (i.e. primary). When color is not found - they should fallback to default color.
- [ ] $themes should be grouped with color helpers and handled in the same way. I.e. they should not be used in components or demo, but ok in user app or docs.
- [ ] Check if we need other theme-bound features, like border color helper.
- [ ] Move dark theme to mixin.
- [ ] Add test to dark theme mixin. Test should compare html of dark and non dark component and fail if they're the same. Also add mixin flag.
- [x] global context config should be derived from default config and service refs.
- [x] Merge is a strategy by default for objects only.

## Types of changes


asvae:
What's up with this PR: https://github.com/epicmaxco/vuestic-ui/pull/336s?
If you don't need it - close?

raichev-dima:
> What's up with this PR: #336?
> If you don't need it - close?

Closed

PR #451, opened 2020-10-20
Feat/build

### LINKED ISSUE #119: Compilation step
## Introduction
Currently application has no compilation step. That means that you have to compile it from outside app. While this approach is super easy for us to support, it introduces some downsides for user. For instance, user won't be able to use components if typescript is not added to the main project build process.

## Steps
- [x] Compile from TS to last generation ES. Double-check it's so.
- [x] Allow user to import + define everything globally. (high priority)
- [ ] Allow user to import only specific components. These should include relevant scss imports. Investigate if it's possible to import same scss twice and prevent that possibility if it exists.
- [ ] Provide typescript declarations (low priority - split into separate issue if that's hard to achieve).

## Notes

SergeiPleshkov:
Closes #119 

## Types of changes


asvae:
Rollup didn't work :/.

PR #456, opened 2020-11-05
Landing

asvae:


packages/docs/assets/colors.scss
```diff
@@ -0,0 +1,5 @@
+$active-blue: #2550C0;
```
asvae, line 1: Routes are not right:
* `/en/landing` replace with `/en`.

asvae, line 1: Language switching doesn't seem to work for me
image

packages/docs/components/landing/Admin.vue
```diff
@@ -0,0 +1,255 @@
...
+          <va-button class="admin__buttons--button"
+            href="https://vuestic.epicmax.co/admin/dashboard"
+            :round="false"
+            color="#2550C0"
+            flat
+          >
+            {{$t('landing.admin.buttons.demo')}}
+          </va-button>
+          <va-button-group class="admin__buttons--button">
+            <va-button class="star-button" color="#f4f4f4" href="https://github.com/epicmaxco/vuestic-admin" :round="false" size="small">
```
asvae, line 25: image
Move closer

image
Same here.
Both low priority.

packages/docs/assets/colors.scss
asvae, line 1: Can we do a couple of px of padding between buttons?
image

asvae, line 1: Styling as different frameworks - either split as separate task or finish in context of this MR.
image

packages/docs/components/landing/Customize.vue
```diff
@@ -0,0 +1,321 @@
...
+      color: #ffffff;
+      padding-top: 1rem;
+      line-height: 1.5rem;
+    }
+  }
+}
+
+.block {
+  &__components {
+    @include row-flex();
```
asvae, line 239: Can you check if we can use vuestic grids, and if yes, can we use them for simple cases at least?

asvae (commented):
Looks good

packages/docs/assets/colors.scss
Kushich: It's because we don't have Russian translation yet

PR #465, opened 2020-11-16
fix/sidebar-active-expand

### LINKED ISSUE #463: active expand in the sidebar component
The filter of the active expand works incorrectly. When we use the same links in different expands (ex: Browsers List in Introduction and List in Ui-components) and click on one of them both this expands are opened

Kushich:
closes #463 

## Types of changes


asvae:
Thanks.
Some explanation would be nice to have, but looks ok.

PR #506, opened 2021-02-08
fix #501 changed dot0 and dot1 ref types from array to HTMLElement.

### LINKED ISSUE #501: Fix VaSlider (range variant)
Range variant of VaSlider are broken both left and right side.

m0ksem:
## Description

closes #501

Concrete vue ref was expected as an array. This caused an error because the `focus` method cannot be called for `undefiend`.

## Types of changes


eqxs (approved):
looks good

PR #507, opened 2021-02-09
fix #502 now the first tab does not become active by default

### LINKED ISSUE #502: Fix VaTabs selection
Tabs selection doesn't work properly.

Step to reproduce:
1. Open VaTabs.demo.vue
2. Check examples with more than 3 tabs.

repro

PS: possible problems with reactivity

m0ksem:
## Description
closes #502 

## Markup:

Also removed unnecessary function. Becouse we have same functionality in setup function.
```js
  // TODO: check if this even works
  parseItems () {
    const content = (this as any).$slots.default || 0
```

## Types of changes


eqxs (approved):
looks good

PR #511, opened 2021-02-09
Fix#508/strange stateful mixin logic

### LINKED ISSUE #508: Fix strange stateful mixin logic
Some components have strange stateful logic usage, similar like computedValue in stateful mixin itself.

For example:
VaModal
VaChip

look on checks like

`    if (this.$props.stateful) {
      this.valueComputed = true
    }
`

m0ksem:
## Description
closes #508 

Replaced undefiend to prop modelValue value:
```ts
export class StatefulMixin extends mixins(PropsMixin) {
  valueState: ValueState = {
    modelValue: this.$props.modelValue,
  }
// ...
}
```

From Readme
> `:value` allows to set default value for stateful component, and `@input` just informs parent of the change.

Also Readme was updated for vue3
>  `:modelValue` allows to set default value for stateful component, and `@update:modelValue` just informs parent of the change.

This changes allow us to set default value for `:modelValue` for stateful components using following code:

```ts
modelValue = prop<boolean>({ type: Boolean, default: true }) 
// Then
this.valueComputed === true // true

// Or

modelValue = prop<boolean>({ type: Boolean, default: false}) 
// Then
this.valueComputed === false // true
```

Tested for components: `VaChip`, `VaModal`

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
+++ b/packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="tagComputed"
-    v-if="valueComputed"
+    v-if="doShowChip"
     :href="hrefComputed"
     :target="target"
     :to="to"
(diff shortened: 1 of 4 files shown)


asvae:
Looks good, thanks!

asvae:
@m0ksem can you fix conflicts, then merge?

PR #516, opened 2021-02-12
fix #515 move v-on listeners from router-link to child div

### LINKED ISSUE #515: va-button doesn't use vue-router history mode as expected
after clicking on the button, the location should be changed without reloading

after clicking on the button, the location changes with reloading

[codesandbox](https://codesandbox.io/s/qye05?file=/src/App.vue)
[website](https://vuestic-test.vercel.app)
play with buttons:
image

* **Other information**

m0ksem:
## Description
closes #515 

`v-on="inputListeners"` was bind to route-link (`<component  tag="router-link" />`).

Route-link does not support `v-on`, therefore we need to use `.native` modifier (for vue2)  or move all listeners to a div wrapper. 

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
+++ b/packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
@@ -1,25 +1,25 @@
 <template>
-  <div
-    class="va-chip__wrapper"
-    @mouseenter="updateHoverState(true)"
-    @mouseleave="updateHoverState(false)"
-    @focus="updateFocusState(true)"
-    @blur="updateFocusState(false)"
+  <component
+    v-if="valueComputed"
+    class="va-chip"
+    :is="tagComputed"
+    :href="hrefComputed"
+    :target="target"
+    :to="to"
+    :replace="replace"
+    :exact="exact"
+    :active-class="activeClass"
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
```diff
@@ -1,40 +1,44 @@
...
-    :to="to"
-    :replace="replace"
-    :exact="exact"
-    :active-class="activeClass"
-    :exact-active-class="exactActiveClass"
-    class="va-chip"
-    :class="computedClass"
-    :style="computedStyle"
-    :tabindex="indexComputed"
+  <div
```
eqxs, line 2: i think this element should be 'inline-flex', like a child element

eqxs (changes requested):
please check the comment

packages/ui/src/components/vuestic-components/va-chip/VaChip.vue
m0ksem: fixed in https://github.com/epicmaxco/vuestic-ui/pull/516/commits/d1245b9d935941edb873b3ba355ba74d65006006

eqxs (approved):
good

PR #536, opened 2021-02-25
Fix#533/input double class

### LINKED ISSUE #533: VaInput uses the class twice
изображение
изображение

изображение
изображение

m0ksem:
## Description
closes #533 

I propose to use two attributes for `va-input` and `input` classes, 

Example:
```
<va-input
        class="mb-4"
        inputClass="p-2 text-red"
        ...
/>
```

This way `mb-4` class will not affect `input` element inside `va-input` and will prevent this class from being used twice.

## Types of changes


untael:
Looks good. Thanks.

PR #547, opened 2021-03-05
Fix broken select

m0ksem:
## Description
Data in demo:
```
[
  {
    text: 'one',
    icon: 'info',
  },
  {
    text: 'two',
    icon: 'face',
  },
  {
    text: 'three',
    icon: 'warning',
  },
]
```

Before:
изображение

All items are highlighted when hovering.

Working on only if data items have `id` keys.

After:
изображение

-----
Also, I think it needs a deeper refactoring

## Types of changes


untael:
Looks good. Thanks.

PR #555, opened 2021-03-11
Auto black/white color based on sidebar color.

m0ksem:
## Description
Auto black/white color based on sidebar color.
изображение

## Types of changes


m0ksem:
Maybe I should not hardcode colors. What the best way to pass colors? In the theme config? Props?

m0ksem:
Also need something like this for VaBadge.
изображение

Maybe for VaButton, VaCard etc. 

It would be nice to add prop textColor for ColorMixin.

PR #560, opened 2021-03-16
IconsConfig

### LINKED ISSUE #538: Set va-icon class.
https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/vuestic-components/va-icon/IconMixin.ts
```js
const getClass = (iconConfig: any, font: string) => {
  let iconClass = ''
  if (isMaterialFont(font)) {
    iconClass = 'material-icons'
  } else if (isMaterialOutlinedFont(font)) {
    iconClass = 'material-icons-outlined'
  } else if (isFontAwesome4Font(font)) {
    iconClass = `fa fa-${iconConfig.code}`
  } else if (isFontAwesome5Font(font)) {
    iconClass = `${font} fa-${iconConfig.code} fa-fw`
  } else if (isIonicFont(font)) {
    iconClass = `icon ion-md-${iconConfig.code}`
  } else if (isIonicOutlineFont(font)) {
    iconClass = `icon ion-ios-${iconConfig.code}-outline`
  } else if (isWeLoveIconsFont(font)) {
    iconClass = `${font}-${iconConfig.code}`
  }

  return iconClass
}
````
The class is now empty.

I want to use icon in my config from scss file 
изображение
My icon class:
```scss
.vuestic-iconset-dashboard::before {
  content: '\0043';
}
```

but there is no option to set `iconClass`

I propose to use `iconConfig.code` as iconClass if it was not found or add `iconClass` property to config.
___

It would also be nice to make a mechanism so that any icons can be supported.

m0ksem:
## Description
closes #538

## Types of changes


packages/ui/src/components/vuestic-components/va-icon/VaIcon.vdemo.vue
```diff
@@ -1,25 +1,25 @@
...
     <VbCard title="MD">
-      <va-icon class="material-icons">
+      <i class="material-icons">
         home
-      </va-icon>
+      </i>
       <va-icon name="home" />
     </VbCard>
     <VbCard title="FA 4">
       <va-icon name="home" />
```
asvae, line 10: `<va-icon class="fa fa-home" />` - this defaults to md icons. Shouldn't be.

We want to prevent config resolution if there is no name at all.

packages/ui/src/services/GlobalConfigPlugin.ts
```diff
@@ -1,8 +1,9 @@
 import { ref, inject, App } from 'vue'
 
 import { getDefaultConfig } from '../components/vuestic-components/va-config/config-default'
-import { merge } from 'lodash'
+import { mergeWith } from 'lodash'
 import { colorsPresets } from './color-config/color-theme-presets'
+import { iconsPresets } from './../services/icon-config/presets'
```
asvae, line 6: Should include only: 
* vuestic aliases (close, arrows etc);
* md default

Everything else should be passed to plugin config or set dynamically.

asvae:
Looks good, but please try to connect to project and see if TS does not complain and types are accessible.

packages/ui/src/components/vuestic-components/va-icon/VaIcon.vdemo.vue
```diff
@@ -75,13 +75,21 @@
       <va-icon :component="getSvgComponent()" />
```
asvae, line 75: Can we add this to config as well? Similarly to other cases?

packages/ui/src/components/vuestic-components/va-icon/VaIcon.vdemo.vue
```diff
@@ -1,25 +1,25 @@
 <template>
   <VbDemo>
     <VbCard title="MD">
-      <va-icon class="material-icons">
+      <i class="material-icons">
```
asvae, line 4: Please use `va-icon` instead of `i` we're comparing direct access to config access.

packages/ui/src/services/icon-config/icon-config.ts
```diff
@@ -1,73 +1,43 @@
...
-      return {
-        componentProps: preset.componentProps(...regexGroups),
-        component: preset.component,
-      }
-    } else {
-      return { component: preset.component }
-    }
+export function getIconConfig (name: string, iconsConfig?: IconsConfig): IconConfig {
+  if (!iconsConfig) {
+    throw new Error(`Cant find config for name="${name}". You can create default config. Visit DOCS // TODO`)
```
asvae, line 33: Probably we don't need the check here. Having icon config is guaranteed.

packages/ui/src/services/icon-config/icon-config.ts
```diff
@@ -1,73 +1,59 @@
...
   }
 
-  const icon = alias?.to || name
-  const preset = iconsConfig.presets?.find((preset) => icon.match(preset.name)) || iconsConfig.defaultPreset
+  return executeConfigFunctions(config.to || name, config)
+}
 
-  const iconConfig = getCssIcon(icon, preset) || getLigatureIcon(icon, preset) || getComponentIcon(icon, preset)
-  return { ...iconConfig, ...defaultPropsFromPreset(preset) }
+export function createIconConfig (config: {
```
untael, line 46: So it will create iconConfig or iconsConfig?

packages/ui/src/services/icon-config/IconMixin.ts
```diff
@@ -1,13 +1,24 @@
...
+import { getIconConfig } from './icon-config'
+import { useGlobalConfig } from '../GlobalConfigPlugin'
 
 class IconMixinProps {
-  name = prop<string>({ type: String, default: '' })
+  name = prop<string>({ type: String, required: true })
   iconsConfig = prop<any>({ type: Object, default: {} })
 }
 
+function copyObjectWithoutUndefined (obj: any) {
```
untael, line 10: Move this to utils?

packages/ui/src/services/icon-config/icon-config.ts
m0ksem: According to config types value can be `undefined`. User can pass `undefined` by accident, so we need to check this.

PR #582, opened 2021-03-31
Refactor button. Fix button-group and button-toggle.

m0ksem:
## Description
Fix button-group gradient. Also clean up code for VaButton component.
изображение

Also docs VaButtonGroup and VaButtonToggle must be updated.

## Types of changes


m0ksem:
Need ButtonGroup style for squared buttons.

Also need to change round prop, to square. Like in other components.

PR #591, opened 2021-04-06
Now accorditon value two way binded with collapse children.

### LINKED ISSUE #589: Accordion do not update this value
изображение
Should update va-accordion value
изображение
I should update value manully

* **Other information**

m0ksem:
## Description
closes #589

There was a problem: accordion wasn't updating its modelValue when collapses.
This probleb was caused becouse `vue-class-component` doesn't provide fully this in setup function. So `this.valueComputed` in `onChildChange` was undefined. 
I rewrite this component to composition api without using `vue-class-component`. Now it's works fine.

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-accordion/VaAccordion.vue
+++ b/packages/ui/src/components/vuestic-components/va-accordion/VaAccordion.vue
@@ -5,19 +5,11 @@
 </template>
 
 <script lang="ts">
-import { provide } from 'vue'
-import { Options, Vue, mixins, prop, setup } from 'vue-class-component'
-import { StatefulMixin } from '../../vuestic-mixins/StatefulMixin/StatefulMixin'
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/vuestic-components/va-accordion/VaAccordion.vue
```diff
@@ -28,79 +20,62 @@ export type Accordion = {
...
+export default defineComponent({
+  emits: ['update:modelValue', 'hello'],
+  props: {
+    modelValue: { type: Array, default: () => [] },
+    multiply: { type: Boolean, default: false },
+    inset: { type: Boolean, default: false },
+    popout: { type: Boolean, default: false },
+  },
+  setup (props, ctx) {
+    const statefull = useStateful(props, ctx.emit)
```
untael, line 32: stateful with one 'l'

PR #703, opened 2021-05-07
Input and select restyle.

### LINKED ISSUE #618: Redesign inputs and select
image

m0ksem:
## Description
closes #618

- [x] Restyle
- [x] Keyboard navigation (focus select, open with space/enter, focus inside select dropdown)
- [x] Hinted search for select
- [x] Bug fixes
- [ ] Correct colors from our global theme
- [ ] CSS variables use global css variables.

Now select uses VaInput as the anchor. It means VaSelect styles relative to VaInput. 

I also dropped the chips value and replaced this with content slot API (as in route-link). In a new case, you can render any component relative to value, not only chips. This way is less complex and more flexible.

## Types of changes


asvae (approved):
Looks good. Thanks.

PR #745, opened 2021-05-14
Added Navbar docs. Added shape and text color props to VaNavbar. 

m0ksem:
## Description
I have to add textColor prop for navbar, which by default is undefined. If textColor is undefined then we use `getTextColor` from our `color-functions`.
I removed unexpected shift for background color, so for now user can be sure that if he passes `#ffde00` color as a prop, then exactly this color will be used.

Also, I added documentation for VaNavbar.

image
image

## Types of changes


asvae (approved):
Looks good. Thanks.

PR #771, opened 2021-05-16
Fixed margin and language dropdown

NikhilSharma03:
Screenshot from 2021-05-16 15-23-23

## Description
There was an issue with the margin around content and the language dropdown

## Markup:

<details>
I fixed margin and language dropdown
```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/docs/src/components/header/Header.vue
+++ b/packages/docs/src/components/header/Header.vue
@@ -116,9 +116,6 @@ export default class Header extends mixins(PropsMixin) {
       flex: 1 1 !important;
       justify-content: space-between;
 
-      & .mr-2{
-        padding-left:20px;
-      } 
 
       .language-dropdown{
         display:flex;
--- a/packages/docs/src/components/header/Header.vue
+++ b/packages/docs/src/components/header/Header.vue
@@ -116,6 +116,16 @@ export default class Header extends mixins(PropsMixin) {
       flex: 1 1 !important;
       justify-content: space-between;
 
+      & .mr-2{
+        padding-left:20px;
+      } 
+
+      .language-dropdown{
+        display:flex;
+        justify-content:flex-end;
+
+      }
+
       & .va-dropdown__anchor,
       .color-dropdown {
         display: flex;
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/header/Header.vue
```diff
@@ -116,6 +116,16 @@ export default class Header extends mixins(PropsMixin) {
       flex: 1 1 !important;
       justify-content: space-between;
 
+      & .mr-2{
```
untael, line 119: mr means margin-right so I think this css is wrong.

NikhilSharma03: Ok i have removed it. Thank you for telling me about the error.

packages/docs/src/layouts/default.vue
```diff
@@ -230,6 +230,13 @@ html {
 
     overflow-y: auto;
     overflow-x: hidden;
+
+    @media(max-width:670px){
+      & > :last-child {
```
m0ksem, line 235:
```suggestion
      & > .layout.gutter--xl {
```

Because of this, we will not have side effects, if we want to add something to the end. And it is just clearer for humans :).

packages/docs/src/layouts/default.vue
```diff
@@ -230,6 +230,13 @@ html {
...
+        padding:0em !important;
```
m0ksem, line 236:
```suggestion
        padding: 0rem;
```

You don't need an `!important` if you will use more priority selector (`.layout.gutter--xl` instead `:last-child`)

packages/docs/src/layouts/default.vue
```diff
@@ -230,6 +230,13 @@ html {
...
+        padding-bottom: 2em;
+      }
```
m0ksem, line 238:
```suggestion
      }
```

Calm down eslint.

m0ksem (changes requested):
And you lost Padding on the left, after that commit 230bb0e0f71d0391fbfa3b81716941a5de4c4130

NikhilSharma03:
@m0ksem Can you please check??

m0ksem (approved):

m0ksem:
@NikhilSharma03, Looks good, thanks!

NikhilSharma03:
Thank you!

PR #797, opened 2021-05-20
Fix #789: Dropdown refactor, closeOnContentClick prop, StatefulMixin are added

### LINKED ISSUE #789: Clicking on Add translation should close dropdown
image

Derranion:
## Description
Fix #789: Refactor Dropdown, closeOnContentClick prop, StatefulMixin are added
## Markup:
Now you could use closeOnContentClick as a flag to mark dropdown that should close its content on click.
Default behavior is set to close content on click.

VaButtonDropdown now uses model as described in our docs instead of not doing anything.

Also StatefulMixin is added to VaDropdown & VaButtonDropdown.

VaDropdown refactored - logic improved and simplified.
<details>

```vue
<va-button-dropdown
      color="primary"
      flat
      :label="currentLanguageName"
      :offset="[0, 25]"
      :isContentClosable="true"
    >
```
</details>

## Types of changes

### DIFF
--- a/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
+++ b/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
@@ -1,5 +1,4 @@
 <template>
-<div>
   <va-button-dropdown
     class="mr-2 mb-2"
     label="Will not close dropdown on click"
@@ -7,5 +6,4 @@
   >
     Content
   </va-button-dropdown>
-</div>
 </template>
--- a/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
+++ b/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
@@ -2,7 +2,7 @@
   <va-button-dropdown
     class="mr-2 mb-2"
     label="Will not close dropdown on click"
-    :close-on-click-inside="false"
+    :close-on-content-click="false"
   >
     Content
   </va-button-dropdown>
--- a/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
+++ b/packages/docs/src/examples/va-button-dropdown/ClickInside.vue
@@ -1,0 +1,11 @@
+<template>
+<div>
+  <va-button-dropdown
+    class="mr-2 mb-2"
+    label="Will not close dropdown on click"
+    :close-on-click-inside="false"
+  >
+    Content
+  </va-button-dropdown>
+</div>
+</template>
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -791,7 +791,7 @@
         "keepAnchorWidth": "Keeps anchor position the same.",
         "offset": "Sets the distance between dropdown and anchor.",
         "modelValue": "Sets a dropdown state.",
-        "closeOnClickInside": "Sets dropdown content on click behavior."
+        "closeOnContentClick": "Sets dropdown content on click behavior."
       },
       "events": {
         "click": "Emitted when user clicks on button.",
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -3056,7 +3056,7 @@
       },
       "clickInside": {
         "title": "Click inside",
-        "text": "You could set the dropdown to be closed or not after clicking on it's content."
+        "text": "You could set the dropdown to be closed or not after clicking on its content."
       }
     }
   },
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -3053,6 +3053,10 @@
       "events": {
         "title": "Events",
         "text": "The `va-dropdown-button` can handle click events."
+      },
+      "clickInside": {
+        "title": "Click inside",
+        "text": "You could set the dropdown to be closed or not after clicking on it."
       }
     }
   },
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -790,7 +790,8 @@
         "position": "Sets the position of the dropdown content.",
         "keepAnchorWidth": "Keeps anchor position the same.",
         "offset": "Sets the distance between dropdown and anchor.",
-        "modelValue": "Sets a dropdown state."
+        "modelValue": "Sets a dropdown state.",
+        "isContentClosable": "Sets dropdown content on click behavior."
       },
       "events": {
         "click": "Emitted when user clicks on button.",
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -3056,7 +3056,7 @@
       },
       "clickInside": {
         "title": "Click inside",
-        "text": "You could set the dropdown to be closed or not after clicking on it."
+        "text": "You could set the dropdown to be closed or not after clicking on it's content."
       }
     }
   },
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -791,7 +791,7 @@
         "keepAnchorWidth": "Keeps anchor position the same.",
         "offset": "Sets the distance between dropdown and anchor.",
         "modelValue": "Sets a dropdown state.",
-        "isContentClosable": "Sets dropdown content on click behavior."
+        "closeOnClickInside": "Sets dropdown content on click behavior."
       },
       "events": {
         "click": "Emitted when user clicks on button.",
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -183,12 +183,6 @@
     </VbCard>
     <VbCard title="Stateful">
       <va-button-dropdown
-        label="Stateful"
-        stateful
-      >
-        Content
-      </va-button-dropdown>
-      <va-button-dropdown
         label="Stateless without v-model"
         :stateful=false
       >
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -175,7 +175,7 @@
     </VbCard>
     <VbCard title="Click inside">
       <va-button-dropdown
-        label="Dropdown content will not be closed on click"
+        label="Dropdown content click will not close it"
         :closeOnClickInside="false"
       >
         Content
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -173,6 +173,28 @@
         Content
       </va-button-dropdown>
     </VbCard>
+    <VbCard title="Click inside">
+      <va-button-dropdown
+        label="Dropdown content will not be closed on click"
+        :closeOnClickInside="false"
+      >
+        Content
+      </va-button-dropdown>
+    </VbCard>
+    <VbCard title="Stateful">
+      <va-button-dropdown
+        label="Stateful"
+        stateful
+      >
+        Content
+      </va-button-dropdown>
+      <va-button-dropdown
+        label="Stateless without v-model"
+        :stateful=false
+      >
+        Content
+      </va-button-dropdown>
+    </VbCard>
   </VbDemo>
 </template>
 
@@ -192,6 +214,11 @@ export default {
       console.log('click', e)
     },
   },
+  data () {
+    return {
+      value: true
+    }
+  }
 }
 </script>
 
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -176,7 +176,7 @@
     <VbCard title="Click inside">
       <va-button-dropdown
         label="Open"
-        :closeOnClickInside="false"
+        :closeOnContentClick="false"
       >
         Clicking here won't close dropdown content
       </va-button-dropdown>
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -181,7 +181,7 @@
         Clicking here won't close dropdown content
       </va-button-dropdown>
     </VbCard>
-    <VbCard title="Stateful">
+    <VbCard title="Stateless">
       <va-button-dropdown
         label="Stateless without v-model"
         :stateful=false
--- a/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
+++ b/packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
@@ -175,10 +175,10 @@
     </VbCard>
     <VbCard title="Click inside">
       <va-button-dropdown
-        label="Dropdown content click will not close it"
+        label="Open"
         :closeOnClickInside="false"
       >
-        Content
+        Clicking here won't close dropdown content
       </va-button-dropdown>
     </VbCard>
     <VbCard title="Stateful">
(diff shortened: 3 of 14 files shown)


packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -51,6 +52,7 @@ class DropdownProps {
   closeOnClickOutside = prop<boolean>({ type: Boolean, default: true })
```
m0ksem, line 52: isContentClosable prop means that you actually have an opportunity to close the dropdown. So if it's false, you can not close the dropdown.

Looks like `close-on-click-inside` makes more sense.

m0ksem (commented):
Also, I think we need to close the dropdown after clicking exactly on links, buttons, etc. It would be great if we just bind the close method to the slot.

In this case, we can write simple code.

```vue
<va-dropdown>
	  <template #anchor> <va-button /> </template>
	  <template #default="{ close }">
		<route-link to="/"> // As I know route-link does not support @click
			<span @click="close">
				Move to another page
			</span>
 		<route-link/>
	    <va-button @click="close"> Close </va-button>
	  </template>
</va-dropdown>
```

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
Derranion: > isContentClosable prop means that you actually have an opportunity to close the dropdown. So if it's false, you can not close the dropdown.
> 
> Looks like `close-on-click-inside` makes more sense.
> 
> 🤔

Thanks, yeah - that sounds even clearer and easier to understand.
I'll fix it.

Derranion: As discussed docs have been added, prop is renamed.

If you don't mind I will create a different branch for on slot click behavior because it will probably need docs & description as well. And this issue looks solved already.

Thanks!

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -130,13 +130,21 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
     })
   }
 
-  onAnchorClick (): void {
-    this.$emit('anchor-click')
+  onElementClick (element: string): void {
```
asvae, line 133: should have proper types

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -15,7 +15,7 @@
         class="va-dropdown__content-wrapper"
         @mouseover="$props.isContentHoverable && onMouseOver()"
         @mouseout="onMouseOut()"
-        @click="$props.closeOnClickInside && onAnchorClick()"
+        @click="$props.closeOnClickInside && onElementClick('dropdown-click')"
```
asvae, line 18: This is not `dropdown-click`. Please think of better name.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -130,13 +130,21 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
-      return
+    if (element === 'anchor-click') {
+      this.$emit(element)
+      if (this.isClicked && !this.closeOnAnchorClick) {
+        return
+      }
+    }
+    if (element === 'dropdown-click') {
+      this.$emit(element)
+      if (this.isClicked && !this.closeOnClickInside) {
```
asvae, line 145: why would you check for `isClicked` here?

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -130,13 +130,21 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
     }
-    if (this.isClicked && !this.closeOnAnchorClick) {
-      return
+    if (element === 'anchor-click') {
+      this.$emit(element)
+      if (this.isClicked && !this.closeOnAnchorClick) {
+        return
+      }
+    }
+    if (element === 'dropdown-click') {
```
asvae, line 143: `click` is not element.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
Derranion: Removed this excessive check in the next commit, thanks!

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -130,23 +130,27 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
     })
   }
 
-  onElementClick (element: string): void {
+  handleInsideClick (signalName: string, toClose: boolean): void {
```
asvae, line 133: `emitName`?

packages/docs/src/locales/en/en.json
```diff
@@ -3032,6 +3033,10 @@
       "events": {
         "title": "Events",
         "text": "The `va-dropdown-button` can handle click events."
+      },
+      "clickInside": {
+        "title": "Click inside",
+        "text": "You could set the dropdown to be closed or not after clicking on it."
```
asvae, line 3039: ...on its content

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -82,20 +82,13 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
-      switch (this.trigger) {
-      case 'hover':
-        return this.isMouseHovered
-      case 'click':
-        return this.isClicked
-      case 'none':
-        return this.modelValue
-      default:
-        return false
+      if (['click', 'hover', 'none'].includes(this.trigger)) {
```
asvae, line 85: There is only 3 triggers, no? Why do you need to check for that?

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -41,34 +44,31 @@ class DropdownProps {
...
   closeOnClickOutside = prop<boolean>({ type: Boolean, default: true })
   closeOnAnchorClick = prop<boolean>({ type: Boolean, default: true })
   isContentHoverable = prop<boolean>({ type: Boolean, default: true })
   offset = prop<number | number[]>({ type: [Array, Number], default: () => [] })
-  trigger = prop<string | number | any[]>({
-    type: [Array, Number, String],
-    default: 'click',
-    validator: (trigger: string): boolean => ['click', 'hover', 'none'].includes(trigger),
-  })
+  trigger = prop<string>({ type: String, default: 'click' })
```
asvae, line 58: pretty sure we want to keep typing here

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -82,20 +82,13 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
-        return false
+      if (['click', 'hover', 'none'].includes(this.trigger)) {
+        return this.valueComputed
       }
+      return false
     }
 
     const triggeredValue = getTriggeredValue()
-
     this.$emit('trigger', triggeredValue)
```
asvae, line 92: Hmm. That's not how getters work.
There should never be the case where we emit from getter.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -105,10 +98,6 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
     // the idea is to emit 'trigger' in any case
     const triggeredValue = this.triggeredValue
```
asvae, line 99: Probably you can return immediately?

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -128,15 +117,34 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
     })
   }
 
+  handleClick (emitName: ClickType, toClose: boolean): void {
+    this.$emit(emitName)
+    if (toClose) {
```
asvae, line 122: `toClose` doesn't seem to be useful, you can just use `this.closeOnClickInside` here directly.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -128,15 +117,34 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
   onAnchorClick (): void {
-    this.$emit('anchor-click')
     if (this.disabled) {
       return
     }
-    if (this.isClicked && !this.closeOnAnchorClick) {
-      return
+    if (this.trigger === 'click') {
+      if (this.valueComputed) {
+        this.handleClick('anchor-click', this.closeOnAnchorClick)
```
asvae, line 141: Why is this event getting triggered only when dropdown is closed?
Wouldn't it make sense to trigger it on any anchor-click?

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -147,20 +155,24 @@ export default class VaDropdown extends mixins(DropdownPropsMixin) {
...
+      this.hoverOutDebounceLoader.reset()
+    }
   }
 
   onMouseOut (): void {
-    if (this.isContentHoverable) {
-      this.hoverOutDebounceLoader.run()
-    } else {
-      this.isMouseHovered = false
+    if (this.trigger === 'hover') {
```
asvae, line 168: You may generally want to use [early return](https://codeburst.io/code-js-in-stlye-early-return-3462323e30d1) instead of if/else.

packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.vue
```diff
@@ -5,8 +5,10 @@
       :disabled="disabled"
       :position="position"
       :offset="$props.offset"
-      @update:modelValue="toggleDropdown"
       :keep-anchor-width="keepAnchorWidth"
+      :close-on-click-inside="closeOnClickInside"
+      v-model="valueComputed"
```
asvae, line 10: Doesn't make sense to use stateful mixin here I think.
If I understood correctly, we only need one state for "shown" dropdown. And that state is already present in dropdown.

So it's valid to pass `:stateful` down, but you can use `valueProxy` getter/setter instead of `valueComputed`.

packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
```diff
@@ -173,6 +173,28 @@
...
+        label="Dropdown content will not be closed on click"
+        :closeOnClickInside="false"
+      >
+        Content
+      </va-button-dropdown>
+    </VbCard>
+    <VbCard title="Stateful">
+      <va-button-dropdown
+        label="Stateful"
+        stateful
```
asvae, line 187: You need only one demo. For whichever is not handled in `default`.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -15,6 +15,7 @@
         class="va-dropdown__content-wrapper"
```
asvae, line 15: Demo seems to completely broken.

`/#/demo/vuestic-components/va-dropdown/VaDropdown.demo.vue`

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
Derranion: We emit it when dropdown is opened as well - by "handleClick" method, not only when it's closed.

We do not emit only when it is disabled but if we need to emit in this case as well I can change it.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
Derranion: Thanks for the info, I've read the article and usually using this approach,

In this case however, if I understand it correctly, we don't need to return because we want to manage onMouseOut event by DebounceLoader functions.

packages/docs/src/examples/va-button-dropdown/ClickInside.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+<div>
```
asvae, line 2: Why do we need div here?

packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.vue
```diff
@@ -103,21 +106,23 @@ class ButtonDropdownProps {
...
-  emits: ['click', 'main-button-click'],
+  emits: ['click', 'main-button-click', 'update:modelValue'],
 })
 export default class VaButtonDropdown extends mixins(
   SizeMixin,
   ColorMixin,
   ButtonDropdownPropsMixin,
 ) {
-  showDropdown = false
+  // showDropdown = false
```
asvae, line 125: You can remove that :)

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -41,34 +44,36 @@ class DropdownProps {
...
   // If set to false - dropdown won't dodge outside container.
   preventOverflow = prop<boolean>({ type: Boolean, default: false })
+  closeOnClickInside = prop<boolean>({ type: Boolean, default: true })
   closeOnClickOutside = prop<boolean>({ type: Boolean, default: true })
   closeOnAnchorClick = prop<boolean>({ type: Boolean, default: true })
   isContentHoverable = prop<boolean>({ type: Boolean, default: true })
   offset = prop<number | number[]>({ type: [Array, Number], default: () => [] })
-  trigger = prop<string | number | any[]>({
-    type: [Array, Number, String],
+  trigger = prop<string>({
```
asvae, line 58:
```ts
type ClickTrigger = 'click' | 'hover' | 'none'

prop<ClickTrigger>
````

Also `: boolean` can be removed.

File: packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue (hunk 14 above)
Comment (asvae) [reply packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue:154]: Why not?

```ts
// We only care about mouseout for hover trigger
if (this.trigger !== 'hover') {
  return
}

if (this.isContentHoverable) {
  this.hoverOutDebounceLoader.run()
} else {
  this.valueComputed = false
}
this.hoverOverDebounceLoader.reset()
```

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
```diff
@@ -357,14 +359,14 @@ export default class VaSelect extends mixins(
   }
 
   selectOption (option: any): void {
-    const isSelected = this.checkIsOptionSelected(option)
-
     if (this.doShowSearchInput) {
       (this as any).$refs.searchBar.focus()
```
m0ksem, line 363: Simply remove this line. As you have an issue with focusing the search bar after an option selected, you need just remove the search bar focusing (on this line).

```
(this as any).$refs.searchBar.focus()
```

packages/ui/src/components/vuestic-components/va-select/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -172,6 +172,12 @@ export default class VaSelectOptionList extends mixins(
     } else {
       this.hoveredOptionComputed = null
     }
+    // We need this to select list items when VaSelect has searchBar
+    // Presumably because searchBar gets focus and after clicking on OptionList line 10: @focus="hoverFirstOption" gets triggered
+    // Looks like this results in registering event listeners for only first visible items in the OptionList so items lower in the list cant be selected
+    this.focus()
```
m0ksem, line 178: You don't need such a hack with refocusing. If you want to leave the focus on the options list, simply remove the focusing search bar after option selection.

m0ksem (changes requested):

packages/ui/src/components/vuestic-components/va-select/VaSelectOptionList/VaSelectOptionList.vue
Derranion: Thanks that's exactly what I wanted to discuss during the call, just a work in progress comment.

asvae (approved):
ThankS!

Issue #802, opened 2021-05-21
VaProgress invisible prop


m0ksem:
I want to hide the progress bar if there is nothing to load. This should not change my page layout.

Some prop that allows setting opacity to 0.

PR #806, opened 2021-05-23
feat: invisible prop for va-progress-bar and va-progress-circle (#802)

smellyshovel:
## Description

This PR adds an `invisible` prop to both the `va-progress-bar` and `va-progress-circle` components. The prop allows to hide/show a component it's applied to by toggling the `opacity` css property which allows to show/hide a corresponding element on a page without runtime-layout-corruption (other elements on the page jumping/moving).

Also provided demos for the `:book` mode and updated the docs' example sections for the both components, though couldn't managed to update the API sections for some reason, therefore **need help** with that.

## Types of changes

### DIFF
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -496,6 +496,7 @@
       "props": {
         "value": "The progress value",
         "indeterminate": "Create a indeterminate loading bar",
+        "invisible": "Make the `va-progress-bar` component invisible (sets its `opacity` css property to `0` which allows to keep the space allocated for it in the layout even when the component is hidden)",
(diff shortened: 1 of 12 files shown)


packages/docs/src/locales/en/en.json
```diff
@@ -496,6 +496,7 @@
       "props": {
         "value": "The progress value",
         "indeterminate": "Create a indeterminate loading bar",
+        "invisible": "Make the `va-progress-bar` component invisible (sets its `opacity` css property to `0` which allows to keep the space allocated for it in the layout even when the component is hidden)",
```
asvae, line 499: Why same translaction 2 times?

PR #812, opened 2021-05-26
Fix #768: List DOM is fixed

### LINKED ISSUE #768: List DOM is off
image

- [ ] Should be under the  same `ul`.

Derranion:
## Description
closes #768

Now we have the ability to use a list of translation strings in our documentation,

Example:
` DocsHelper.list(['installation.node','installation.npm'])`

I've updated our current documentation with newly added feature, where it was needed as well.
<details>

```ts
list: (translationStringList: TranslationString[]): ListBlock => {
  return {
    type: BlockType.LIST,
    translationStringList
  }
},
```
</details>

## Types of changes

### DIFF
--- a/packages/docs/src/components/DocsContent.vue
+++ b/packages/docs/src/components/DocsContent.vue
@@ -42,7 +42,7 @@
       <MarkdownView
         v-else-if="block.type === BlockType.LIST"
         :key="block.type + index"
-        :value="`${block.translationStringList.map(translationString=>$t(translationString))
+        :value="`${block.translationStringList.map(translationString=>'- ' + $t(translationString))
                                               .join('\n')}`"
       />
       <DocsTable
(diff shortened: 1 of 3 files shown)


packages/docs/src/components/DocsContent.vue
```diff
@@ -39,6 +39,12 @@
         :key="block.type + index"
         :value="`${$t(block.translationString)}`"
       />
+      <MarkdownView
```
asvae, line 42: Any reason to keep '-' in translactions?

packages/ui/src/services/api-docs/ManualApiOptions.ts
```diff
@@ -3,6 +3,7 @@
 // as strings.
 
 export type TranslationString = string
+export type TranslationStringList = string[]
```
asvae, line 6: TranslationString[]

asvae:
Thanks. Looks good!

asvae (approved):

PR #823, opened 2021-06-01
Remove va-medium-editor from vuestic-ui.

m0ksem:
## Description
This PR moves va-medium-editor from vuestic-ui to vuestic-admin repo. 
We don't need va-medium-editor in vuestic-ui, because of these breaks SSR.

There is no documentation about this component, so I think we can just remove it.

## Types of changes


asvae:
Thanks, looks good.

PR #824, opened 2021-06-01
fix: va-button clickable area

### LINKED ISSUE #822: Button cannot be clicked outside content
image

smellyshovel:
Fix #822.

The `v-on="inputListeners"` used to be attached to the contents block (which is basically the button's text (usually)) and not to the button element itself. Changed the behavior and it seems working as intended to.

Was it some kind of a mistake/typo or am I missing something?


smellyshovel:
What it used to be like:
usedtobe

How it works now:
havebecome

Issue #810, opened 2021-05-25
Miss `:focus` state in sidenav items


cawa-93:
* **What is the expected behavior?**
Link in focus is visually selected

* **What is the current behavior?**
Links in sidenav hasn't `:focus` state. 

* **Steps to reproduce, ideally also a screenshot or gif if it's a style issue**
Just try navigate with `tab`.

smellyshovel:
Current behavior:

![Peek 2021-06-03 12-33](https://user-images.githubusercontent.com/9782236/120622949-f2da4b00-c467-11eb-89a0-83c24fd308fb.gif)

You can see that it's actually possible to switch between items with the `tab` key, but the items aren't actually highlighted.
The other issue is that it's still possible to select items that are hidden with `accordion`/`collapse`, which, I believe, shouldn't be the case.

smellyshovel:
Solved in #828 
![Peek 2021-06-03 12-33](https://user-images.githubusercontent.com/9782236/120648476-5a05f880-c484-11eb-93bb-e234f1d08cd6.gif)

PR #828, opened 2021-06-02
fix: #810 #829 #830 #833 #836 sidebar items and va-collapse

smellyshovel:
## Description

Not sure how good this solution is, but I'm having issues with using mixins when defining components with `defineComponent` (and VaSidebarItem happens to be the one defined that way).

So I've deviced to rewrite the Keyboard Only Focus Mixin the way that it's became a composable: `useKeyboardOnlyFocusMixin`.

It uses a bit different approach that the old mixin does (in terms of defining which type of focus event happened: keyboard or mouse), but still it solves #810 .

```vue
// Your code
```
</details>

## Types of changes


smellyshovel:
Solves #810 #829 and #830 

Peek 2021-06-03 12-33

## Considering #810 

Items are now properly highlighted.

## Cosidering #829 

See details in #829 

## Cosidering #830

See details in #830

packages/ui/src/composables/useKeyboardOnlyFocus.ts
```diff
@@ -0,0 +1,32 @@
+// it's another implementation of KeyboardOnlyFocusMixin's functionality but for use with the Composition API
+
+import { ref } from 'vue'
+
+export default function useKeyboardOnlyFocus () {
+  const hasKeyboardFocus = ref(false)
+  let prevClicked = false
```
asvae, line 7: Please no abbreviations.

smellyshovel:
solves #833 as well now

Switches
VaSwitch

Tabs
VaTab

Checkboxes
VaCheckbox

Chips
VaChip

Collapses
VaCollapse

List Items
VaListItem

packages/docs/src/components/sidebar/Sidebar.vue
```diff
@@ -6,6 +6,7 @@
       <va-collapse
         v-for="(route, key) in navigationRoutes"
         :key="key"
+        :focusable="false"
```
asvae, line 9: What's the logic here?

packages/ui/src/components/vuestic-components/va-list/VaListItem.vue
```diff
@@ -44,23 +43,31 @@ const ListItemPropsMixin = Vue.with(ListItemProps)
...
   get computedClass () {
     return {
       'va-list-item--disabled': this.disabled,
-      'va-list-item--focus': this.isKeyboardFocused,
     }
   }
 
-  onFocus () {
-    (this as any).KeyboardOnlyFocusMixin_onFocus()
-    this.$emit('focus')
```
asvae, line 63: Maybe keep event?

smellyshovel: It's moved to the template.

packages/docs/src/components/sidebar/Sidebar.vue
smellyshovel: Removed.

smellyshovel: Actually, no. The prop itself is removed, but forgot this usage. TODO

smellyshovel: removed in ec0ed78 (will push it shortly)

smellyshovel:
Now when you use a custom header for a `va-collapse`:

```vue
<va-collapse>
  <template #header="{  changeValue }">
    <button @click="changeValue()">Click me!</button>
  </template>
</va-collapse>
```

it exposes the `changeValue` function via the scoped slots which when invoked opens/closes the va-collapse.

smellyshovel:
done

packages/docs/src/components/sidebar/Sidebar.vue
```diff
@@ -7,13 +7,13 @@
         v-for="(route, key) in navigationRoutes"
         :key="key"
       >
-        <template #header>
-          <va-sidebar-item :activeColor="activeColor">
+        <template #header="{ value }">
+          <va-sidebar-item :activeColor="activeColor" @keydown.stop>
```
asvae, line 11: Why `@keydown.stop`?

smellyshovel: otherwise the event would be triggered twice (on parent and on child) and the expand wil be immediately closed right after it's opened

PR #831, opened 2021-06-03
Fix #827. Fixed bug with trackby prop as functions.

### LINKED ISSUE #827: VaSelect key-by and track-by
Not very clear what key-by does if we have track-by.

m0ksem:
## Description
For now, valueBy returns value in `v-model`, but still shows text from textBy.

image

## Types of changes


asvae:
Looks legit. 👍

PR #845, opened 2021-06-11
fix: #844 don't use <a> for collapse custom headers in docs' sidebar

### LINKED ISSUE #844: Sidebar items (expands) for the docs require 2 tab-presses before focused
Now we need to press the tab key an additional time to focus next sidebar item (docs). Should be a single press.

Leftover from #828

smellyshovel:
closes #844 

This also fixes the keyboard-tab'ing issue.

## Types of changes

### DIFF
--- a/packages/docs/src/components/sidebar/Sidebar.vue
+++ b/packages/docs/src/components/sidebar/Sidebar.vue
@@ -9,10 +9,10 @@
       >
         <template #header="{ value, hasKeyboardFocus }">
           <div
-            class="collapse-custom-header"
+            class="sidebar__collapse-custom-header"
             :class="{
-              'collapse-custom-header--active': isRouteHasActiveChild(route),
-              'collapse-custom-header--keyboard-focused': hasKeyboardFocus
+              'sidebar__collapse-custom-header--active': isRouteHasActiveChild(route),
+              'sidebar__collapse-custom-header--keyboard-focused': hasKeyboardFocus
             }"
           >
             {{ $t(route.displayName) }}
@@ -127,7 +127,8 @@ export default class Sidebar extends Vue.with(Props) {
 <style lang="scss" scoped>
 @import "~vuestic-ui/src/components/vuestic-sass/resources/resources.scss";
 
-  .collapse-custom-header {
+.sidebar {
+  &__collapse-custom-header {
     padding: 1rem 1.2rem;
     display: flex;
     justify-content: space-between;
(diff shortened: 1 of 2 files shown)


smellyshovel:
#844

asvae (commented):
Perfect, thanks!

packages/docs/src/components/sidebar/Sidebar.vue
```diff
@@ -7,15 +7,17 @@
...
-          <va-sidebar-item :activeColor="activeColor" @keydown.stop>
-            <va-sidebar-item-content :class="{ 'va-sidebar-item--active': isRouteHasActiveChild(route) }">
-              <va-sidebar-item-title>
-                {{ $t(route.displayName) }}
-              </va-sidebar-item-title>
-              <va-icon v-if="route.children" :name="value ? 'expand_less' : 'expand_more'" />
-            </va-sidebar-item-content>
-          </va-sidebar-item>
+        <template #header="{ value, hasKeyboardFocus }">
+          <div
```
m0ksem, line 11: It would be great if you fix the issue with the VaSidebarItem and VaCollapse components instead of removing them from the docs page.

image

There are two ways:
- Set tabindex -1 for <a> and react to VaCollapse focus
- Remove tabindex for VaCollapse header, to prevent extra focusing.

smellyshovel: The thing here is that, as per @asvae , there should be no links (anchors) at all inside the va-collapse's header (and it actually makes sense since we don't want to redirect the user somewhere - instead, we only want to open the collapse). And since the va-sidebar-item implys the use of links (it uses the `<a>` tag), I have to stop using the va-sidebar-item for collapse's header at all.

Here's how the new markup looks like:
Screenshot_20210614_182450

smellyshovel: And yep, the collapse's header itself still needs to be `tabindex`able to make it possible to open/close collapses using the keyboard.

asvae: @m0ksem I'm frankly not seeing any point jamming sidebar-item into collapse header. Outside of styling similarity it doesn't extend any of its behavior.

packages/docs/src/components/sidebar/Sidebar.vue
```diff
@@ -7,15 +7,17 @@
...
+            class="sidebar__collapse-custom-header"
+            :class="{
+              'sidebar__collapse-custom-header--active': isRouteHasActiveChild(route),
+              'sidebar__collapse-custom-header--keyboard-focused': hasKeyboardFocus
+            }"
+          >
+            {{ $t(route.displayName) }}
+            <va-icon v-if="route.children" :name="value ? 'expand_less' : 'expand_more'" />
```
asvae, line 19:
```
v-if="route.children"
```
Can you check if this check is relevant? IMO if route doesn't have children - it shouldn't be collapse.

asvae:
Thanks. Looks good.

asvae (approved):

PR #852, opened 2021-06-14
chore(deps): move "vue" from deps to peer-deps and dev-deps

### LINKED ISSUE #838: Move vue to peer dependencies
We want to add vue as peer dependency. That prevents conflicts when different version is being used in app.

See implementations from other frameworks:
* [vuetify](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/package.json)
* [element-plus](https://github.com/element-plus/element-plus/blob/dev/package.json)

smellyshovel:
closes #838. Move the "vue" dependency from deps to peer-dependencies and dev-dependencies.

## Types of changes

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -40,7 +40,6 @@
     "normalize.css": "^8.0.1",
     "spa-http-server": "^0.9.0",
     "v-tooltip": "^2.1.2",
-    "vue": "^3.0.4",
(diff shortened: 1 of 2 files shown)


asvae:
Pretty sure this won't do anything useful.
Vue is going to be include in build either way.

asvae:
Thanks. Merged manually.

PR #856, opened 2021-06-15
Moved gravatar in peer dependencies.

### LINKED ISSUE #839: Move gravatar in peer dependencies
Quite a beefy dep for what it does and is not useful for many projects.

DieselTripleX:
## Description

closes #839

## Types of changes


m0ksem:
https://vuestic.dev/en/ui-elements/avatar#with-gravatar

asvae:
Just some early feedback, but we don't need the whole feature.

So removing that from docs, props, demos and deps would be a prudent action.

m0ksem (approved):

PR #859, opened 2021-06-16
Modal appended to document root via <teleport>.

### LINKED ISSUE #817: Modal should be appended to document root, not to parent.

```
<section class="content"></section>

<div class="modal"></div>
```

```
<section class="content">            
    <div class="modal"></div>
</section>
```

image

1. Go to http://localhost:8080/admin/ui/modals.
2. Open DevTools
2. Find `.app-layout__page`
3. Set this element z-index to 0.

* **Other information** 

If we set modal parent z-index then, this z-index will be applied to modals (that how it works). For now it works well in our docs and vuestic-admin, but modals can be easily broken if the user will set position: relative or z-index to his `app-content`.

DieselTripleX:
## Description

closes #817 

fix#817

## Types of changes

### DIFF
--- a/packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
@@ -1,5 +1,5 @@
 <template>
-  <teleport v-if='valueComputed' :to="teleportTarget">
+  <teleport v-if='valueComputed' :to="target">
     <div class="va-modal">
       <modal-element
         name="va-modal__overlay--transition"
@@ -94,7 +94,7 @@ import VaIcon from '../va-icon'
 
 class ModalProps {
   modelValue = prop<boolean>({ type: Boolean, default: false })
-  teleportTarget = prop<string>({ type: String, default: '#app' })
+  target = prop<string>({ type: String, default: '#app' })
   title = prop<string>({ type: String, default: '' })
   message = prop<string>({ type: String, default: '' })
   okText = prop<string>({ type: String, default: 'OK' })
--- a/packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
@@ -1,82 +1,84 @@
 <template>
-  <div class="va-modal">
-    <modal-element
-      name="va-modal__overlay--transition"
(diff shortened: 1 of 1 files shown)


DieselTripleX:
@asvae I accidentally broke formatting for the component, now it's all fixed. Look at it now, please :)

packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
```diff
@@ -92,6 +94,7 @@ import VaIcon from '../va-icon'
 
 class ModalProps {
   modelValue = prop<boolean>({ type: Boolean, default: false })
+  teleportTarget = prop<string>({ type: String, default: '#app' })
```
m0ksem, line 97: I don't think users will understand the meaning of the `teleportTarget` prop. Need a better name.

m0ksem (changes requested):

m0ksem (approved):

packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
```diff
@@ -220,6 +220,10 @@ export default class VaModal extends mixins(
       }
   }
 
+  get isTeleportDisabled () {
+    return this.parentNode === 'parent'
```
m0ksem, line 224: Need docs about this feature.

packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
```diff
@@ -94,7 +94,7 @@ import VaIcon from '../va-icon'
 
 class ModalProps {
   modelValue = prop<boolean>({ type: Boolean, default: false })
-  target = prop<string>({ type: String, default: '#app' })
+  parentNode = prop<string>({ type: String, default: 'body' })
```
m0ksem, line 97: You can declare type above 
```ts
declare type CSSSelector = string
```

```suggestion
  parentNode = prop<CSSSelector | 'parent'>({ type: String, default: 'body' })
```

This will be useless for typing, but it would be easier to understand this prop for humans.

m0ksem (commented):
Can be better. Also, need docs for this feature.

m0ksem:
Moved to https://github.com/epicmaxco/vuestic-ui/pull/965 as I can see.

PR #861, opened 2021-06-17
Fix #846 #854 #858 #868 #875 #736 Select issues, disabled state issues

### LINKED ISSUE #736: Test select
- [x] There are multiple issues which we want to test and fix.

-----------------
Unsorted:

- [x] Searchable - сёрчбар должен оставаться на месте и не скроллится
- [x] Клики на селект листают элементы вниз
- [x] Нажатие ↑ закрывает селект
- [ ] Неправильный вид курсора

But should be more

### LINKED ISSUE #846: Select issues
- [x] Disabled select should not be focusable
image

- [x] Remove outline from internal search, cursor should be the only indication of focus
image

- [x] Navigating up from search should not close dropdown
- [x] Navigating up and down shouldn't unfocus search (use state to manage internal cursor, not focus).

### LINKED ISSUE #854: When you open select - it should show selected item
Right now it just stays at beginning.

### LINKED ISSUE #858: VaInput disabled state issue
Disabled VaInput should not change cursor style on hover & its content should not be selectable.

Right now hover changes cursor style, text inside is selectable.

* **Screenshots:**

VaInput Disabled is Hoverable and selectable
VaInput Disabled selection

### LINKED ISSUE #868: VaSelect multiselect creates empty elements
If there are no elements selected inside Multi Select pressing Space or Enter should not add elements.

If there are no elements selected inside Multi Select pressing Space or Enter right now adds empty elements.

1. Open Multi Select without search
2. Press Space or Enter a few times

MultiSelectEmptyElements

### LINKED ISSUE #875: Select search bar is not visible inside a long list
Search bar should be visible after navigation back to the top when using keyboard.
If we scroll to the bottom of the long list navigation back to the top using keyboard will not make search bar visible.

1. Open any Select with search bar that has five or more items.
2. Press ArrowDown or Scroll down until the search bar will not be visible
3. Press ArrowUp to go up
4. Search bar will never become visible

searchIsNotVisible

### LINKED ISSUE #928: Select loses focus.
Select should stay focused after being closed by click.
Select loses focus after being closed by click.

1. Open Select by clicking on it
2. Click again on Select to close it
3. It will lose focus

Derranion:
## Description
close #846
close #858
close #854 
close #868
close #875
close #736
close #928 

VaSelect navigation should feel better now.

Also fixed most of the disabled state issues of our components.

Some of the changes:
- Disabled state fixes of the Input, Dropdown, Select
- Select navigation fix and enhanced user experience 
- Additional hotkey for Select (Esc)
- Improved search for Select
- Opening Select will hover and focus last selected item to improve usability
- Search bar is now always visible
- Navigation and selection could be used from Search bar
- Scroll animation is instant
- Select will stay focused after being closed by mouse click
- Tab moves to the next element right away
- Readability improvement and some performance tweaks

And some other bug fixes.

## Types of changes


Derranion:
Work in progress - other discussed bug fixes and usability improvements will be added soon.

Derranion:
Work in progress:
- Found new bugs will fix them as well, other updates are incoming.

packages/docs/src/locales/en/en.json
```diff
@@ -2381,7 +2381,7 @@
       },
       "allowCreate": {
         "title": "Allow create",
-        "text": "You can add options that initial options not contain"
+        "text": "You can add options that initial options do not have"
```
smellyshovel, line 2384: You can add options not listed among the initial ones (or smtn like)

packages/docs/src/locales/en/en.json
```diff
@@ -2391,7 +2391,7 @@
         "title": "Keyboard navigation",
         "moves": "By using `ARROW DOWN` and `ARROW UP` you can move between options",
         "selects": "By using `ENTER` and `SPACE` you can select current option",
-        "hints": "If you start typing on opened option list it will create a search buffer (will be reset when a new key is not typed for 1 second) that will be used to search in the options labels"
+        "hints": "If you type while option list is opened this will create a search buffer (will reset if you do not type for 1 second) that will be used to search in the options labels"
```
smellyshovel, line 2394: ...that will reset if you stop typing...

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
```diff
@@ -542,6 +578,11 @@ export default class VaSelect extends mixins(
       this.hintedSearchQuery += event.key
     }
 
+    if (this.doShowSearchInput) {
```
smellyshovel, line 581: quite a strange naming imho
maybe should better be "showSearchInput"?

smellyshovel (approved):

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
Derranion: Sounds good.

packages/docs/src/locales/en/en.json
```diff
@@ -2381,7 +2381,7 @@
       },
```
asvae, line 2381: Simple select still closes when I press `up` on first item. It shouldn't.
image

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -85,6 +88,17 @@ export default class VaDropdown extends mixins(
     }
   }
 
+  get stateClasses (): string[] {
```
asvae, line 91: I like to call this `classComputed`, but doesn't matter much.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -85,6 +88,17 @@ export default class VaDropdown extends mixins(
...
+    const baseClass = 'va-dropdown'
+    const classes: string[] = []
+
+    if (this.$props.disabled) {
+      classes.push(`${baseClass}_disabled`)
```
asvae, line 96: Looks very hacky. What the intention is here?
Generally computeds should not have side effects.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -285,6 +299,10 @@ export default class VaDropdown extends mixins(
   /* Solved the alignment problem (if we try to align inline and block elements) */
   line-height: var(--va-dropdown-line-height);
 
+  &_disabled {
```
asvae, line 302: `&__disabled`?

`block__element--modifier`

packages/ui/src/components/vuestic-components/va-input/VaInput.demo.vue
```diff
@@ -256,6 +256,7 @@
         :messages="messages"
       />
     </VbCard>
+<!--    TODO: Figure out why demo overrides user-select: none-->
```
asvae, line 259: What do you mean here?

packages/ui/src/components/vuestic-components/va-input/VaInput.vue
```diff
@@ -191,6 +191,10 @@ export default class VaInput extends mixins(
       classes.push(`${baseclass}_error`)
     }
 
+    if (this.$props.disabled) {
+      classes.push(`${baseclass}_disabled`)
```
asvae, line 195: Also super weird.

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
```diff
@@ -93,27 +92,30 @@
...
+          :id="$props.id"
+          ref="searchBar"
+          v-model="searchInputValue"
+          class="va-select__input"
+          placeholder="Search"
+          removable
+          :name="$props.name"
+          :tabindex="tabindex + 1"
+          :bordered="true"
+          @keydown.up.stop.prevent="hoverPreviousOption"
```
asvae, line 111: `="hoverPreviousOption()"` adding brackets is preferable.
Makes it explicit you don't pass anything to method.

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
```diff
@@ -436,6 +463,18 @@ export default class VaSelect extends mixins(
...
+
+  scrollToSelected () {
+    if (!this.valueComputed.length) {
+      return
+    }
+
+    const selected = this.valueComputed
+    const scrollTo = typeof selected === 'string' ? selected : selected[selected.length - 1]
+    this.hoveredOption = scrollTo
+    this.$nextTick(() => (this.$refs as any).optionList.scrollToOption(scrollTo))
```
asvae, line 477: Minor thing, but you can declare `@Ref('optionList')`, then you don't have to use `any`.

packages/docs/src/locales/en/en.json
```diff
@@ -2381,7 +2381,7 @@
...
       "allowCreate": {
```
asvae, line 2382: Single select shouldn't support array as value. We can throw an error or warn at least.

image

packages/docs/src/locales/en/en.json
Derranion: Sorry, wasn't able to reproduce it.

For me it stays opened and do not close.

I've checked our code as well:

- Key Up from these "Options list" in this case emits
`this.$emit('no-previous-option-to-hover')`


- And this is the only thing we do in parent
`@no-previous-option-to-hover="focusSearchBar"`

focusSearchBar just focuses search bar, nothing more.

That's why described behavior seems really strange, may be you could try it again I think problem was in something else.

packages/ui/src/components/vuestic-components/va-input/VaInput.demo.vue
Derranion: Disabled VaInput text is selectable in our demo even though user-select correctly set to none.

Disabled VaInput works as intended everywhere else.

Looks like our VbPage overrides `user-select` property.

It was a side note to figure it out,

Thanks I will remove it.

packages/docs/src/locales/en/en.json
Derranion: Yeah, thanks it's related to all single selects.

There is no check for this.

I'll fix it.

packages/ui/src/components/vuestic-components/va-select/VaSelect.vue
Derranion: Thanks.

I've tried to use our decorator for refs but if I do something like this:

- Declare up top `@Ref('optionList') optionListRef: any`
- `... this.$nextTick( this.optionListRef.scrollToOption(scrollTo) )`

It works but there are some cases where ref points to null.

We could declare it within the next tick as well but it kind of overcomplicates logic and kills the whole point of using it.

Correct me if I'm wrong, maybe I missing something.

asvae: we might not have a way to do without nexttick, my point was only about ref

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
Derranion: Completely agree that Computed and Getters should not have any side effects.

But I believe there was a misunderstanding because I do not see any side effects here.
```
  get classComputed (): string[] {
    const baseClass = 'va-dropdown'
    const classes: string[] = []

    if (this.$props.disabled) {
      classes.push(`${baseClass}--disabled`)
    }

    return classes
  }
```

It's just a local constant which is used for convinience.

Most of our components are using the same approach - binding classes list Object or Array.

I could change it to something like this:

```
get classComputed () {
    return { 
      va-dropdown--disabled: this.$props.disabled
    }
}
```

From my perspective though current version is better for a possible component expansion in the future.

PR #865, opened 2021-06-18
Added setHSLAColor in addition of shiftHSLAColor.

### LINKED ISSUE #841: shiftHSLAColor, but it is set instead of shift
image

It would be nice to have a function to set color, instead of shifting.

For example, I want to change the opacity (hsla alpha).

DieselTripleX:
## Description

closes #841

Added setHSLAColor function and unit-test for it, as requested.

## Types of changes

### DIFF
--- a/packages/ui/src/services/color-config/color-functions.ts
+++ b/packages/ui/src/services/color-config/color-functions.ts
@@ -45,6 +45,28 @@ export const shiftHSLAColor = (color: ColorInput, offset: { h?: number; s?: numb
   return result.HSLA
 }
 
+export const setHSLAColor = (color: ColorInput, newColor: { h?: number; s?: number; l?: number; a?: number }) => {
+  const result = new ColorTranslator(color)
+
+  if (newColor.h !== undefined) {
+    result.setH(newColor.h)
+  }
+
+  if (newColor.s !== undefined) {
+    result.setS(newColor.s)
+  }
+
+  if (newColor.l !== undefined) {
+    result.setL(newColor.l)
+  }
+
+  if (newColor.a !== undefined) {
+    result.setA(newColor.a)
+  }
+
+  return result.HSLA
+}
+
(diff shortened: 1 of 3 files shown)


packages/ui/src/services/color-config/color-functions.ts
```diff
@@ -45,6 +45,28 @@ export const shiftHSLAColor = (color: ColorInput, offset: { h?: number; s?: numb
   return result.HSLA
 }
 
+export const setHSLAColor = (color: ColorInput, newColor: { h?: number; s?: number; l?: number; a?: number }) => {
+  const result = new ColorTranslator(color)
+
+  if (newColor.h !== undefined) {
+    result.setH(newColor.h)
```
m0ksem, line 52: Would be nice to write this code in one line and decrease function length.

packages/ui/src/services/color-config/tests/color-functions.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { setHSLAColor } from '../color-functions'
+
+describe('ColorFunctions', () => {
+  describe('setHSLAColor', () => {
+    it('changeHEXhue', () => {
```
m0ksem, line 5: Rename test names to human language. Like "Like change hue in hex color".

m0ksem (changes requested):

packages/ui/src/services/color-config/tests/color-functions.spec.ts
DieselTripleX: Got it!

m0ksem:
This function should be used somewhere in the framework. It would be nice, if you add this to useColors hook.

Then a user can use it in his components.

```
packages/ui/src/services/color-config/color-config.ts:50```

m0ksem (approved):

PR #870, opened 2021-06-22
Fixed DOC's color palette and names.

### LINKED ISSUE #711: DOCS: Search placeholder color
Update search placeholder color to `secondary` which is `#767C88`.

DieselTripleX:
## Description

closes #711 

1. Fixed placeholder color as intended.
2. Added new colors to DOC's themes and adjusted names.
3. Fixed cases when used old color names with new theme.

## Types of changes

### DIFF
--- a/packages/docs/src/components/sidebar/Sidebar.vue
+++ b/packages/docs/src/components/sidebar/Sidebar.vue
@@ -1,5 +1,5 @@
 <template>
-  <va-sidebar class="sidebar" v-model="visible" :width="sidebarWidth">
+  <va-sidebar class="sidebar" v-model="visible" color="background" :width="sidebarWidth">
     <algolia-search />
 
     <va-accordion v-model="value" multiply>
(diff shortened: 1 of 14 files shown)


packages/docs/src/components/sidebar/Sidebar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <va-sidebar class="sidebar" v-model="visible" :width="sidebarWidth">
+  <va-sidebar class="sidebar" v-model="visible" color="background" :width="sidebarWidth">
```
m0ksem, line 2: If you set sidebar default color to the background you don't need to add this prop here.
```suggestion
  <va-sidebar class="sidebar" v-model="visible" :width="sidebarWidth">
```

packages/docs/src/examples/va-sidebar/Color.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="mb-4" style="height: 16rem; border: 2px solid var(--va-primary);">
-    <va-sidebar color="primary" text-color="secondary" minimizedWidth="64px" width="18rem">
+    <va-sidebar color="primary" text-color="white" minimizedWidth="64px" width="18rem">
       <va-sidebar-item>
```
m0ksem, line 4:
```suggestion
      <va-sidebar-item active-color="background">
```

The first sidebar item should also have an active-color prop, like other items. Hover color is relative to active color. You can check colored docs demos and see how different is hover color.

m0ksem (changes requested):

PR #880, opened 2021-06-23
Fixed vaDropdown overflowing issue.

### LINKED ISSUE #850: Use much complex solution for Modals, Dropdowns, Popovers
Use much complex solution for Modals, Dropdowns, Popovers to prevent bugs with overflow: hidden.

We need to make a fixed dropdown, that can be shown outside overflow-hidden card and divs.

There is some example: https://codepen.io/agop/pen/itbew

DieselTripleX:
## Description

closes #850

1. Fixed overflowing issue for Dropdown component. (It also fixed Popover component overflowing issue, because Popover essentially using Dropdown component)
2. Removed VaPopup from the project.

## Types of changes


m0ksem (approved):

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -41,15 +43,11 @@ class DropdownProps {
   position = prop<string>({ type: String, default: '' })
   hoverOverTimeout = prop<number>({ type: Number, default: 30 })
   hoverOutTimeout = prop<number>({ type: Number, default: 200 })
-  boundaryBody = prop<boolean>({ type: Boolean })
   modelValue = prop<boolean>({ type: Boolean, default: false })
   disabled = prop<boolean>({ type: Boolean })
-  // Makes no sense
-  // fixed = prop<boolean>({ type: Boolean })
+  teleportTarget = prop<string>({ type: String, default: '#app' })
```
m0ksem, line 48: Need a better name for this prop. We don't need to use 'teleport' work in this prop name.

m0ksem (changes requested):

m0ksem:
How about keyboard navigation with this teleport feature?

m0ksem (approved):

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -41,15 +43,11 @@ class DropdownProps {
   position = prop<string>({ type: String, default: '' })
   hoverOverTimeout = prop<number>({ type: Number, default: 30 })
   hoverOutTimeout = prop<number>({ type: Number, default: 200 })
-  boundaryBody = prop<boolean>({ type: Boolean })
   modelValue = prop<boolean>({ type: Boolean, default: false })
   disabled = prop<boolean>({ type: Boolean })
-  // Makes no sense
-  // fixed = prop<boolean>({ type: Boolean })
+  target = prop<string>({ type: String, default: '#app' })
```
asvae, line 48: What's this magical `#app`?
Maybe use `body` instead?

We likely want this component to work fine even if there is no `#app`.

packages/ui/src/components/vuestic-components/va-dropdown/VaDropdown.vue
```diff
@@ -11,17 +11,19 @@
...
-      <div
-        class="va-dropdown__content-wrapper"
-        @mouseover="$props.isContentHoverable && onMouseOver()"
-        @mouseout="onMouseOut()"
-        @click="onDropdownContentClick()"
-        ref="contentWrapper"
-      >
-        <div :style="$props.keepAnchorWidth ? anchorWidthStyles : ''">
-          <slot />
+      <teleport :to="target">
```
asvae, line 14: This demo seems to be broken.

image

There are generally 2 cases we want to cover:
1) User want dropdown to be global.
2) User wants to keep dropdown inside a wrapper (rare case).
If you think we don't have to consider 2nd case at all - please remove it from demo (mb some other places).

asvae:
Left a couple of points.
Thanks for removing old code!

PR #886, opened 2021-06-25
#883 added shadow prop for va-chip

### LINKED ISSUE #883: Remove shadow from Chip by default
Make chip shadow as an option.

Now it looks like:
image

By default it should looks like:
image

sergey-kravcov:
close #883 

## Description
Added shadow prop for va-chip.
Refactoring va-chip component.
Updated docs and vue-book for shadow of va-chip

## Types of changes


m0ksem (approved):

packages/docs/src/components/page-configs/ui-elements/va-chip/page-config.ts
```diff
@@ -123,6 +123,18 @@ export default [
     type: BlockType.EXAMPLE,
     component: 'va-chip/Link',
   },
+  {
```
asvae, line 126: doesn't matter much, but you can use helpers here.

asvae:
Thanks!

PR #888, opened 2021-06-25
Feat/Date-picker

### LINKED ISSUE #855: Datepicker component

m0ksem:
## Description
closes #855

VaDateInput prototype

#### Components list
- [x] VaDatePicker
- [x] VaDateInput - input and VaDatePicker in dropdown 
- [ ] 🤔 VaTimePicker - we can move this component to next and then add prop `show-time-picker`
##### Subcomponents - I don't think we want to allow users to use them, but, why not?
- [x] VaDayPicker
- [x] VaMonthPicker
- [x] 🤔 VaYearPicker 
- [x] VaDatePickerHeader

#### Form props
// Require https://github.com/epicmaxco/vuestic-ui/issues/910
- [x] modelValue - Date | Date[] | { start: Date, end: Date }
- [ ] name - string (?)
- [ ]  disabled - boolean
- [ ] readonly - boolean
- [ ] min / max - Im not sure we should use it, because we have allowedDates

#### Date picker props
- [x] mode - 'single' | 'multiple' | 'range' -  for now it's automatically relative to modelValue. But if the user wants to use multiple or range? I don't think there is a way to select a range or multiple dates at the same time. 
Value can be undefined so we can not get mode automatically. So we need this prop.
- [x] type - 'day' | 'month'
- [x] weekends - (d: Date) => boolean // Custom weekends like New Year etc.
// Maybe rename to disabled, not allowed 
- [x] 🤔 allowedDates - (d: Date) => boolean
- [x] 🤔 allowedMonths - (m: Date) => boolean
- [ ] 🤔 allowedYears - (y: Date) => boolean
- [ ] 🤔 displayFormat - string | function - with function we can provide example of using https://date-fns.org/
- [x] firstWeekDay - string - 'Monday' | 'Sunday'
- [x] weekDayNames - string[] - ['M', 'T', 'W', 'T', 'F', 'S', 'S']
- [x] monthNames - string[] - ['JUN', 'FAB', ..., 'DEC']
- [x] ~~dayName - _function - (d) =>  'prefix' + d~~ Make this using slot
- [x] ~~currentDate - slightly outline current date  
  / or  
  showCurrentDate - we can use new Date()~~
      highlightTodayDate - default: true
- [ ] 🤔 locale - (?) maybe on next, this way we can support non-gregorian calendars
- [x] view - 'day' | 'month' ~~| 'year'~~ - current view or current picker (?)
- [ ] 🤔 events - { date: Dates | (d: Date) => boolean, eventName: string }
- [x] showOtherMonths { default: false }

#### Dropdown
- [x] isOpen - boolean

#### Style
- [x] color - default 'primary'
- [x] weekendColor - default null - for example we can use 'danger' 
- [x] ~~prevMonthIcon~~ Replace with slots 👨‍🎓
- [x] ~~nextMonthIcon~~ Replace with slots 👨‍🎓
- [x] hideWeekDays - or we can pass clear array to weekDayNames to hide it

### Emits
- [x] update:modelValue
- [x] update:day, update:month, update:year
- [x] click, click:day, click:month, ~~click:year~~
- [x] hover, hover:day, hover:month, ~~hover:year~~
- [x] update:is-open
- [x] update:view

### Slots
- [x] ...all slots from input  
- [x] ...all slots from dropdown
- [x] day
- [x] month
- [x] header-text
- [x] year
- [x] button:prev
- [x] button:next

### Methods
// 🤔 I think we do not need it 

### Features
- [x] Range select
- [x] Multiple select 
- [x] Statefull
- [ ] 🤔 Parse dates from VaInput text. (Maybe next)
- [x] Keyboard navigation

### Css variables
- [x] Colors css variables
### Docs
- [ ] TODO


m0ksem:
highlightWeekends & weekendColor -> weekendColor { type: String, default: undefined }

packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
```diff
@@ -210,9 +210,9 @@ export default {
   },
```
asvae, line 210: We probably want to hide other months by default.

image

packages/ui/src/components/vuestic-components/va-button-dropdown/VaButtonDropdown.demo.vue
```diff
@@ -210,9 +210,9 @@ export default {
...
   data () {
```
asvae, line 211: No keyboard support so far. Maybe different issue, but we should consider that.

packages/ui/src/components/vuestic-components/va-datepicker/VaDatePicker.demo.vue
```diff
@@ -0,0 +1,137 @@
...
+      <va-date-input v-model="value" value-type="month" class="mb-4" />
+
+      <va-date-input v-model="range" value-type="month" view="year" class="mb-4" />
+
+      <va-date-input v-model="dates" value-type="month" view="year" class="mb-4" />
+    </VbCard>
+
+    <VbCard title="date picker">
+      Date
+      <va-date-picker v-model="value" class="mb-4" />
```
asvae, line 106: Please move to different demo. It's different component.

packages/ui/src/components/vuestic-components/va-datepicker/VaDatePicker.demo.vue
```diff
@@ -0,0 +1,137 @@
...
+      </va-date-input>
+    </VbCard>
+
+    <VbCard title="first weekday">
+      <va-date-input v-model="value" label="Sun (default)" class="mb-4" />
+
+      <va-date-input v-model="value" label="Monday" firstWeekday="Monday" class="mb-4" />
+    </VbCard>
+
+    <VbCard title="disable dates">
```
asvae, line 92: Maybe not yellow?
image

packages/ui/src/components/vuestic-components/va-datepicker/VaDatePicker.demo.vue
```diff
@@ -0,0 +1,137 @@
...
+      <va-date-input v-model="value" label="Disable all Tuesday and Thursday" :allowedDates="(date) => date.getDay() !== 2 && date.getDay() !== 4" />
+    </VbCard>
+
+    <VbCard title="select month view">
+      <va-date-input v-model="value" value-type="month" class="mb-4" />
+
+      <va-date-input v-model="range" value-type="month" view="year" class="mb-4" />
```
asvae, line 99: placing first item of range, then clicking outside of range should reset selection.

packages/ui/src/components/vuestic-components/va-datepicker/VaDatePicker.demo.vue
```diff
@@ -0,0 +1,137 @@
+<template>
```
asvae, line 1: would be nice to see stateful and form usage examples

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDatePickerHeader/VaDatePickerHeader.vue
```diff
@@ -0,0 +1,106 @@
...
+import { VaDatePickerView } from '../../types/types'
+
+export default defineComponent({
+  name: 'VaDatePickerHeader',
+
+  emits: ['update:year', 'update:month', 'update:view'],
+
+  props: {
+    // Use default here to make year and month not undefined type
+    year: { type: Number, required: true, default: () => 0 },
```
asvae, line 31: required and default are mutually exclusive

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDatePickerHeader/VaDatePickerHeader.vue
```diff
@@ -0,0 +1,106 @@
...
+    month: { type: Number, required: true, default: () => 0 },
+    monthNames: { type: Array, required: true },
```
asvae, line 33: pretty sure we do not want to use required here.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDatePickerHeader/VaDatePickerHeader.vue
```diff
@@ -0,0 +1,106 @@
...
+          syncMonth.value = 0
+        } else {
+          syncMonth.value = syncMonth.value + 1
+        }
+      } else if (view?.value === 'year') {
+        syncYear.value = syncYear.value + 1
+      }
+    }
+
+    const prev = () => {
```
asvae, line 57: Hmm. Maybe date-fns? A lot of things could go wrong with vanilla date, especially if you factor in timezones.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDayPicker/VaDatePickerCalendarHook.ts
```diff
@@ -0,0 +1,65 @@
...
+  const localizeWeekday = (weekdayNumber: number) => {
+    if (options && options.firstWeekday?.value === 'Monday') {
+      // Set Sunday as 7th day of the week and Monday as first day of the week.
+      return weekdayNumber === 0 ? 6 : weekdayNumber - 1
+    }
+
+    return weekdayNumber
+  }
+
+  const getPreviousDates = () => {
```
asvae, line 23: also pretty sure we want to use date-fns here.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDayPicker/VaDayPicker.vue
```diff
@@ -0,0 +1,268 @@
...
+import { isDatesArrayInclude, isDatesEqual } from '../../utils/date-utils'
+import { VaDatePickerModelValue } from '../../types/types'
+
+const isToday = (date: Date): boolean => date.toDateString() === new Date().toDateString()
+
+export default defineComponent({
+  name: 'VaDayPicker',
+
+  props: {
+    monthNames: { type: Array as PropType<string[]>, required: true, default: [] },
```
asvae, line 59: No config support, which might be fine for  first iteration. But we want to have it eventually.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaMonthPicker/VaMonthPicker.vue
```diff
@@ -0,0 +1,228 @@
...
+        font-size: 9px;
+        color: var(--va-secondary);
+        font-weight: bold;
+        height: $cell-height;
+        line-height: $cell-height;
+      }
+    }
+  }
+
+  .calendar {
```
asvae, line 151: Avoid using simple class names to avoid conflicts.

Our core feature is ability to use in context of existing app.

packages/ui/src/components/vuestic-components/va-datepicker/VaDatePicker.demo.vue
m0ksem: image

Fixed, but need design in Figma.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDayPicker/VaDayPicker.vue
m0ksem: VaDateHeader, VaDayPicker, and VaMonthPicker are private components, that VaDatePicker uses. If the user wants to change `month-names` he can use such components config:

```ts
...
VaDatePicker: { monthNames: [...] }
...
```

Almost all props from VaDateHeader, VaDayPicker, and VaMonthPicker are required and should be passed from the VaDatePicker component, there is why users need to configure only VaDatePicker.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDatePickerHeader/VaDatePickerHeader.vue
m0ksem: I don't use the Dates object here, because there will be a lot of issues with reactivity. I will make the code simpler to understand. But it will be much dirty working here with dates.

packages/ui/src/components/vuestic-components/va-datepicker/components/VaDayPicker/VaDatePickerCalendarHook.ts
m0ksem: I can see any date-fns helper that can be useful here.

DieselTripleX:
image
I think, there is no need of current chosen month name in year's picker.

m0ksem:
image

Need highlight today month.

m0ksem:
show-other-months -> display-other-months
first-weekday: `Monday` -> `Monday, Mon, monday, mon`

m0ksem:
- [ ] For range date picker. If clicked on start should change start, if clicked on end - should change end.

packages/ui/src/components/vuestic-components/va-date-picker/utils/child-props.ts
```diff
@@ -0,0 +1,52 @@
+import { ComponentOptionsBase, PropType, computed, ComputedRef, Prop } from 'vue'
+
+/**
+ * Accepts parent component props and return value only for child component props.
+ *
+ * Used to proxy child component props from parent.
+ */
+export const filterComponentProps = (propsValues: Record<string, any>, childProps: Record<string, any>): ComputedRef<Record<keyof typeof childProps, any>> => {
+  return computed(() => Object
```
Derranion, line 9: Maybe we could do it a bit faster :)

 ```
return Object.keys(childProps)
              .reduce((acc, propName) => {
                  acc[propName] = propsValues[propName]
                  return acc
              }, {})
}
```

Instead of:
```
return computed(() => Object
    .keys(childProps)
    .reduce((acc, propName) => {
      return { ...acc, [propName]: propsValues[propName] }
    }, {}),
  )
```

To not iterate object every time, if you would like.

asvae:
Thanks! Great work.
I didn't check everything, but functionality-wise things work exactly as the should.

So let's merge it after docs are up :).

asvae (approved):

m0ksem:
@m0ksem, rename click:prepend to click-prepend according to VaInput naming.

PR #893, opened 2021-06-28
Typography re-work without using tables.

### LINKED ISSUE #790: Typography is not mobile friendly
Use the same approach as in other pages.

image

DieselTripleX:
## Description

closes #790 

Typography page was re-worked without using tables, which provides better mobile compatibility.

## Types of changes


packages/docs/src/components/DocsCode.vue
```diff
@@ -9,6 +9,20 @@ import 'prismjs/components/prism-bash'
 import VuePrismComponent from 'vue-prism-component'
 
 export default {
+  data() {
```
m0ksem, line 12: Would be nice to have a comment on why exactly do you need to rerender VuePrismComponent on code prop change. Why not just pass code prop to VuePrismComponent?

m0ksem (changes requested):

packages/docs/src/components/DocsCode.vue
asvae: Yeah, doesn't seem appropriate.

packages/docs/src/components/DocsCode.vue
```diff
@@ -1,4 +1,5 @@
 <template>
+  <!-- Tricky way to update code rendering component via v-if and nextTick. Watch handler() code below -->
```
m0ksem, line 2:
```suggestion
  <!-- Tricky way to update code rendering component via v-if and nextTick. See `code` prop watcher code below -->
```

packages/docs/src/components/DocsCode.vue
```diff
@@ -18,7 +19,7 @@ export default {
     code: {
```
m0ksem, line 19: Move `watch` under `props`.

Follow this order:
```
Props
Data
Computed
Methods
Watch
```

m0ksem (changes requested):

packages/docs/src/components/DocsCode.vue
DieselTripleX: 👍

m0ksem (approved):

m0ksem:
@DieselTripleX, resolve conflicts, please.

m0ksem (approved):

Issue #869, opened 2021-06-21
Dev Tools freeze on the docs page.


m0ksem:
Lags on Firefox and Chrome.

DieselTripleX:
I would like to figure it out!)

m0ksem:
Check how it's working with a new project created with vue cli. Does it freeze?

asvae:
A lot of css variable rewrites might be an issue actually >.>.

m0ksem:
Looks like fixed.

PR #908, opened 2021-06-30
fixed warnings about reactive objects in docs

### LINKED ISSUE #898: Docs in dev mode is showing warnings about reactive components.
TODO: provide a screenshot.

For now, you can open any docs page and see a lot of warnings.

sergey-kravcov:
close #898 

## Description
Fixed warnings about reactive objects in doc pages.

It was:
Снимок экрана 2021-06-30 в 17 03 25

Now:
Снимок экрана 2021-06-30 в 17 03 05

## Types of changes


m0ksem (approved):
This PR fixes #869 as I can see.

asvae:
@sergey-kravcov can you explain a bit how that works (in comments)?

sergey-kravcov:
@asva There is a component in the documentation that is responsible for displaying examples of using the components from Vuestick. This component is DocsExample. It not only shows examples of use, but also the example code. Therefore, as a prop, it takes the name of the component of the example. After that, it imports the component itself in the created hook and also reads its code (dividing the code into three parts: template, style, script). To display a component of the example, it must be available for the template of DocsExample, so you need to place it in the variable from `this`. But this variable must not be reactive, otherwise that warning will appear.

asvae:
@sergey-kravcov I see the problem.

you can probably just 'v-if=component' instead of using additional variable

sergey-kravcov:
@asvae The `component` variable is not reactive and does not cause the template to be updated after the component is imported. So I used an additional variable.

asvae:
@sergey-kravcov and why it's not reactive? Maybe you can just use null instead? :P

sergey-kravcov:
@asvae This variable cannot be reactive because warnings will pop up again. I used the shallowRef variable instead of the reactive one and it did the trick.

alygee:
> * [x]  Same thing should be also fixed on landing
>   image

passed raw objects to buttons in footer to avoid these warnings

PR #933, opened 2021-07-07
Fixed clipboard button.

### LINKED ISSUE #919: The error appears after clicking the Copy code button
The code should be copied.

The error appears after clicking the Copy code button and code do not copy.

Снимок экрана 2021-07-05 в 11 03 04

Tested in google chrome

DieselTripleX:
## Description

Closes #919

Fixed clipboard button, using Clipboard API. Previously, v-clipboard plugin for nuxt was used for copy functional, but we rejected nuxt after moving from vue2 -> vue3.

## Types of changes

### DIFF
--- a/packages/docs/src/components/DocsNavigation.vue
+++ b/packages/docs/src/components/DocsNavigation.vue
@@ -71,10 +71,17 @@ export default {
     }
   },
   methods: {
+    updateCopyButton(icon, text) {
+      this.copyIcon = icon
+      this.copyText = text
+    },
     copy () {
-      this.$clipboard(this.code)
-      this.copyIcon = 'fa fa-check'
-      this.copyText = 'Copied'
+      try {
+        window.navigator.clipboard.writeText(this.code)
+        this.updateCopyButton('fa fa-check', 'Copied')
+      } catch(e) {
+        if (e.message === 'NotAllowedError') { this.updateCopyButton('fa fa-times', 'Permission failure!') }
+      }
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/DocsNavigation.vue
```diff
@@ -71,10 +71,17 @@ export default {
...
+    updateCopyButton(icon, text) {
+      this.copyIcon = icon
+      this.copyText = text
+    },
     copy () {
-      this.$clipboard(this.code)
-      this.copyIcon = 'fa fa-check'
-      this.copyText = 'Copied'
+      try {
+        window.navigator.clipboard.writeText(this.code)
```
m0ksem, line 80: According to specification https://w3c.github.io/clipboard-apis/#dom-clipboard-writetext, write text is an async function. You should use `await` here to catch errors.
```suggestion
        await window.navigator.clipboard.writeText(this.code)
```

Can you check if I correct? Maybe this `await` can break your code 😄

m0ksem (changes requested):
As I can see [Clipboard API](https://caniuse.com/mdn-api_clipboard_writetext) has partial support for all browsers. For security reasons, I think.

But, [Clipboard API WriteText](https://caniuse.com/mdn-api_clipboard_writetext) is supported everywhere, except IE 11.

A small change is required.

m0ksem (approved):

asvae:
Thanks!

PR #935, opened 2021-07-07
Composition API Select, refactor & new composable files

### LINKED ISSUE #924: New Select property 'Max'
Right now we have something similar in our demo examples:

maxDemo
Custom tag max code:
```
<va-select
        v-model="defaultMultiSelect.value"
        class="mb-4"
        label="Custom tag max (3)"
        :options="defaultMultiSelect.options"
        multiple
        :tag-max="3"
/>
```

But there is no such property as 'tag-max' so it doesn't work.

Would be nice to have this feature - to be able to limit the amount of selected options.

Could be named as "Max" or "Limit".

Derranion:
## Description

**Key changes:**

- **Select is fully reworked and uses Composition API now.**
- Select refactored ( improved readability and structure )
- Some unneeded anymore dependencies are removed.
- Vue Class Component in the Select is replaced by other internal Vue tools in order to fully support TypeScript while making code lighter and faster. ( TS does not support some of the VCC syntax e.g. @Options )
- **New composable files** are added including new feature **useMaxSelections**, description is below.

close #924

New feature for **Select** has been added - property "**max-selections**":
- It allows users to set the maximum amount of selected options.

We could add this feature to the multiple other components by new composable file **useMaxSelections**, to: 
- OptionsList
- Accordion
- Checkbox

I'll make new issues and could implement this next.

## Documentation example:

maxSelectionsExample
## Code example:
```
<va-select
   class="mb-4"
   label="Multiple select, Max 2 selections"
   :options="options"
   v-model="valueMultipleMax"
   multiple
   max-selections=2
/>
```

## Types of changes


Derranion:
As discussed - looks like we could re-use **max-selections** property in other components.

It could be wise to add it to our selectable mixin or use composition API for it.

Will implement this as well.

Derranion:
We want to move our Select from Options API to Composition API so it's better to add this pull request later because otherwise this syntax will need to be changed one more time.

Derranion:
Removed "**on hold**" status because this pull request could be updated now, after our [Select updates](https://github.com/epicmaxco/vuestic-ui/pull/861).

We have a few issues currently - TypeScript can't parse some of the Vue Class Component syntax, for example adding props by  `@Options` syntax.

Second one is that `setup` hook which we use for Composition API is triggered earlier than mixins.

That's why I will move Select to Composition API, remove Vue Class Component ( we will still have complete TS support ) and add new Composable methods instead of mixins to be able to work with them inside a setup.

asvae:
Looks good!

asvae (approved):

PR #937, opened 2021-07-08
add docs page about color themes

### LINKED ISSUE #871: Add docs about using CSS variables colors
We have colors CSS variables --va-primary, --va-secondary etc. Which reactive relative to ColorsConfig.

Need some docs about it.

sergey-kravcov:
close #871 

## Description
Added Russian and English versions of docs page about color CSS variables.
Снимок экрана 2021-07-08 в 09 26 52
Снимок экрана 2021-07-08 в 09 26 39

## Types of changes


packages/docs/src/examples/color-variables/DefaultColors.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div v-for="(value, key) in colors" :key="key" :style="`color: ${value};`">{{ key }} (--va-{{ key }}): {{value}}</div>
```
m0ksem, line 2: Would be nice to use here black text color and some colored shape with border. Like VaColorIndicator, but handmade.

image

packages/docs/src/examples/color-variables/ColorBorder.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+  <div style="border: 1px solid var(--va-success);">Use --va-success</div>
```
m0ksem, line 2: Maybe you can make something more interesting here.

image

```
<template>
  <div style="border: 1px solid var(--va-danger); background-color: var(--va-warning);">

    <i class="mx-2" style="color: var(--va-danger);">!</i>

    <span style="color: var(--va-white);">This is custom alert that uses CSS variables</span>
  </div>
</template>
```

m0ksem: Also, you can pass props to an example component, that will be used instead of "This is custom alert...". 

We don't have this option for now, but you are free to add it. You can add it to ExampleOptions if you want.

image

m0ksem (approved):
Looks good. I added some not-required suggestions.

sergey-kravcov:
The page after some changes. Снимок экрана 2021-07-12 в 14 36 53

packages/docs/src/components/page-configs/styles/colors/page-config.ts
```diff
@@ -0,0 +1,39 @@
+import { ApiDocsBlock } from '@/types/configTypes'
+import { DocsHelper } from '@/helpers/DocsHelper'
+
+const scheme = `
```
asvae, line 4: That'll be pretty confusing for end user.

Instead we can have a structure like this.
* Overview
* Setup
  * Usage example
  * Default colors
* In-app usage
  * CSS helpers + example
  * How to use CSS variables
  * How to override CSS variables

PR #954, opened 2021-07-15
Tree-shaking

### LINKED ISSUE #489: try to use modules build
see https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77

we should probably still add fallback, but module build should be a priority.

### LINKED ISSUE #840: Tree shaking (a la carte)
See [vuetify tree shaking](https://vuetifyjs.com/en/features/treeshaking/). You can also check other frameworks.

What's we're trying to achieve:
* User should be able to load all framework in one go (full bundle).
* User should be able to load only parts of framework (partial bundle). We want to focus primarily on components (services can wait for now).

### LINKED ISSUE #955: Tidy CSS
Separate our SCSS into three groups:
- Global: modules like grid, typography, reset, normalize.
- SCSS Utlis: Only SCSS mixins, functions, variables. these files should not be compiled into a CSS bundle.
- CSS bundle. One file, that requires all global modules. Will be used to creating UMD and IIFE builds.

We want to allow the user to chose if he needs grid, typography, or normalize modules.
`// vuestic-grid.css, vuestic-typography.css, vuestic-normalize.css`
Also, there should be an option to include all at one, so we need some vuestic.css files.
`// vuestic.css`

Also, you should look for components, that depend on Global modules. Components should be independent. But it's okay to use SCSS utils (now we call them recourses).

### Related issues
closes: https://github.com/epicmaxco/vuestic-ui/issues/950
related to: https://github.com/epicmaxco/vuestic-ui/issues/840
with tree-shaking, we can fix this issue: https://github.com/epicmaxco/vuestic-ui/issues/876.

m0ksem:
## Current state
This PR is in progress for now.

I created test project with vite. Then, I imported only VaInput component, without whole plugin.
image

## Todo
- [x] 🤔 Remove UUID node module from vuestic, as it depends on the built-in crypto module. Do we really need this in the UI library?
- [x] Remove unused rollup plugins :)
- [x] Need to test umd build with CDN (just include it using `script` tag)
- [x] ✔ Need to test how vite works with `withTrasportConfig` function. There were some errors. 
- [x] For ESM build: Check file structure. Why do we have the `src/components/vuestic-components` path? ]
- [x] Split `umd` to `iife` (for browser) and `cjs` (for SSR) formats.
image
It will hard to change this file structure using Rollup, so I think we need to simplify this structure in the SRC project.
#### We can not import ts file from js script, without extension:
```
// Wrong
<script>
import MyFile from 'index'

// Ok
<script lang="ts">
import MyFile from 'index'
// Or
<script>
import MyFile from 'index.ts'
```
- MAYBE: there can be some resolve plugin. Or can be hacked with allowJS.
// This problem is not important, because we import ts file from js only in two components. But it can be dangerous in feature, because IDE doesn't highlight such error.
#### For some reason CSS includes recourses, variables etc CSS.

image

## Vite
Also, I have tested it with vite. But, unfortunately, there is no way to use the `preserveModules` option. 
Only one way to create separate ESM build use `manualChunks: id => parse(id).name` rollup option. This way appear some strange problems like this. We need spread files in ESM build because this is how tree-shaking we can make.
image
Also this approach does not provide any tree-shaking
image
(Importing VaInput only)

closes #840
closes #955
closes #489


m0ksem:
I caught myself on the issue when I do not import the VaInput component because use the Global Declared one. Need some eslint rule for this issue or convention.

m0ksem:
Tree-Shaking API:

```ts
// main.ts (default)
import { VuesticPlugin } from 'vuestic-ui'
import 'vuestic-ui/dist/vuestic.css'

app.use(VuesticPlugin)
```

```ts 
// main.ts (tree-shaking)
import { VuesticPlugin } from 'vuestic-ui/dist/esm'
import 'vuestic-ui/dist/css/typography.css'
import 'vuestic-ui/dist/css/css-variables.css'

app.use(VuesticPlugin)
```

```vue
// App.vue (tree-shaking)
improt VaInput from 'vuestic-ui/components/VaInput'
```

vuestic.css:
 - grid
 - css variables
 - typography
 - etc.

m0ksem:
We checked the UMD bundle with fake CDN usage. Works fine:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <link href="./vuestic-ui.css" rel="stylesheet" />
</head>
<body>
  <div id="app">
    <va-input></va-input>
    <va-button></va-button>
  </div>
  <script src="https://unpkg.com/vue@next"></script>
  <script src="./vuestic-ui.js"></script>
  <script> 
    Vue.createApp({}).use(VuesticUI.VuesticPlugin).mount('#app')
  </script>
</body>
</html>
```

asvae:
Thanks. Great work!

Issue #952, opened 2021-07-13
The Data Table component


sergey-kravcov:
## Description
- The component is used for displaying tabular data. Features include sorting, searching, pagination and row selection.

## Features
- [ ] Visual feedback - When user interacts with component they should see result of their actions (:active, :hover, :focus selectors and ...).
- [ ] Keyboard navigation.
- [ ] Stateless support.

## Docs
- [ ] Do proper documentation page (#17).

## Testing
- [ ] Demos for all cases.
- [ ] Couple of tests for core functionality.

## Props
- [ ] `custom-sort` function` - Function used to sort items
- [ ] `expand-icon` string` - Icon used for expand toggle button.
- [ ] `expanded` array ` - Array of expanded items. Can be used with .sync modifier
- [ ] `headers` DataTableHeader[]` - An array of objects that each describe a header column
DataTableHeader: {
  label: string,
  key: string,
  align?: 'start' | 'center' | 'end',
  sortable?: boolean,
  class?: string | string[],
  cellClass?: string | string[],
  sort?: (a: any, b: any) => number
}
- [ ] `hide-default-header` Boolean` - Hide the default headers
- [ ] `item-class` string | function` - Property on supplied items that contains item’s row class or function that takes an item as an argument and returns the class of corresponding row
- [ ] `item-key` string ` - The property on each item that is used as a unique key
- [ ] `items` array ` - The array of items to display
- [ ] `items-per-page` number ` - Changes how many items per page should be visible.
- [ ] `loading` Boolean` - If true and no items are provided,
- [ ] `loading-text` string ` - Text shown when loading is true and no items are provided
- [ ] `multi-sort` Boolean` - If true then one can sort on multiple properties
- [ ] `no-data-text` string` - Text shown when no items are provided to the component
- [ ] `no-results-text` string` - Text shown when search prop is used and there are no results
- [ ] `page` number ` - The number of page
- [ ] `search` string` - Text input used to filter items
- [ ] `server-items-length` number` - Used only when data is provided by a server.
- [ ] `show-expand` Boolean` - Shows the expand toggle in default rows
- [ ] `show-select` Boolean` - Shows the select checkboxes in both the header and rows (if using default rows)
- [ ] `single-expand` Boolean` - Changes expansion mode to single expand
- [ ] `single-select` Boolean` - Changes selection mode to single select
- [ ] `sort-by` string | array` - Changes which item property (or properties) should be used for sort order.
[...]

PR #956, opened 2021-07-16
Added partial version of data table

sergey-kravcov:
Added part of data table functions.
You can see completed items here #952 

Снимок экрана 2021-07-16 в 14 31 15

## Types of changes


packages/ui/src/components/va-data-table/hooks/useFilterable.ts
```diff
@@ -0,0 +1,31 @@
+import { Ref, computed, watch } from 'vue'
+import { TableRow } from './useRows'
+
+export type TFilteringFn = (source: any) => boolean
+
+// TODO: `emit` must not be any!
```
smellyshovel, line 6: @RVitaly1978 please, take a look at what we can do with `emit: any` here and in other hooks. There must be a standard Vue declaration for emits, we should use it.

smellyshovel:
Generally approved, though I strongly recommend to get rid of `emit: any` in a couple of places.

packages/ui/src/components/va-data-table/hooks/useFilterable.ts
RVitaly1978: Done.

PR #965, opened 2021-07-21
Dropdown, Modal, Popover nesting

### LINKED ISSUE #850: Use much complex solution for Modals, Dropdowns, Popovers

DieselTripleX:
## Description

closes #850

First of all, merged #859 and #880.
1. Added teleport functional for dropdowns and modals to prevent overflowing.
  1.1 Added <attachElement> prop that responsible for teleport destination selector (Default: "body").
  1.2 Added <disableAttachment> prop that disables teleport feature (In rare cases, might be useful).
2. Removed popover component from the project.
3. Fixed select component styling issue, after teleporting dropdown to DOM root.
4. Changed demo's descripition for incoming fixes.

## Types of changes


packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
```diff
@@ -1,82 +1,84 @@
 <template>
```
asvae, line 1: Nested modal doesn't work correctly.

packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
```diff
@@ -92,6 +94,8 @@ import VaIcon from '../va-icon'
 
 class ModalProps {
   modelValue = prop<boolean>({ type: Boolean, default: false })
+  attachElement = prop<string>({ type: String, default: 'body' })
```
asvae, line 97: can we move this to composable file?

asvae:
Feel free to merge with squash after fix.

asvae (approved):

packages/ui/src/components/vuestic-components/va-modal/VaModal.vue
DieselTripleX: 👌 Got it!

m0ksem (changes requested):
Modal can not be closed. It causes an error inside Vue and then breaks vue-i18n.

image

m0ksem:
Remade modals a little bit, nested modals work fine now.

m0ksem:
Docs required here. Also, you don't specify attachment props for VaButtonDropdown and other components that using VaDropdown.

m0ksem:
I think we can provide `attachElement` prop using VaConfig in some rare cases, so we don't need a special prop to pass to VaDropdown. 

I suppose we can make some service that will allow us to pass props to specific child components later.
For example, we can create a config directive
```vue
<va-select ... v-components-config="{ VaDropdown: { 'close-on-click-outside': false } }" />
```

For now solution will be
```vue
<va-config :components="{ VaDropdown: { 'close-on-click-outside': false } }">
  <va-select ... />
</va-config>
```

Anyway passing props to VaDropdown in some components can be a rare case so we can handle it later.

asvae:
Thanks!

PR #976, opened 2021-07-26
Simplify props declaration in docs.

### LINKED ISSUE #972: Simplify props declaration in docs.
image

Too hard doing this 🐵. 

ApiDocs.vue component should be rewritten.

m0ksem:
## Description
closes #972

This PR:
- removes the obligation to write "local" in `api-options.ts` and automatically handles a translation string.
Likely you want to review this commit https://github.com/epicmaxco/vuestic-ui/commit/3ecb2f9848cedf2028d25ae035db6f9e798a3880 instead of all PR.
- removes old api-docs service from the vuestic-ui package.

---
I want to use it while I writing docs for DatePicker. It just speeds up this process.


asvae:
Can we also remove local configs? Not seeing too much value out of deprecation.

asvae:
Why do deprecation when you can just remove `local`?

m0ksem:
After this PR can be a lot of undocumented events shown to users.

image

So we need to check it out.

PR #977, opened 2021-07-27
Fix #971 Fix router anchor links in docs

### LINKED ISSUE #971: Following anchor links via router-link does not work correctly
  - when a user clicks on a router-link with an anchor, there should be a transition to the top of the element with an anchor

  - if the anchor is on the current page, then the jump to the anchor is not performed (nothing happens)
  - if the anchor is on another page, then the transition to the page top is performed and not to the anchor 

* **Other information**

RVitaly1978:
## Description
close #971

Some of the changes:
  - if the anchor is on the current page, then the jump to the anchor is performed
  - if the anchor is on another page, then the transition to the anchor is performed
  - if the link without anchor is on another page, then the transition to the page top is performed

## Types of changes


packages/docs/src/router/index.ts
```diff
@@ -25,4 +25,16 @@ export const router = createRouter({
       children: routes.map((r) => ({ ...r, name: undefined })),
     },
   ],
+  // Good solution, but not working
```
asvae, line 28: Remove this?

asvae (approved):

asvae:
Cool!

PR #999, opened 2021-08-10
Added the ability to localize examples in the docs

### LINKED ISSUE #951: Add the ability to localize examples in the docs
It would be nice to be able to pass text to examples, it would help with localization.
Снимок экрана 2021-07-12 в 15 46 26
It should be taken into account that the examples should remain adequate and the text localization code should not be very conspicuous in these examples.

mxmvshnvsk:
## Description
Added the ability to localize examples in the docs, removed `vue-prism-component`. Close #951 

## Types of changes


asvae:
Looks good. Thanks!

Issue #998, opened 2021-08-10
Move tslib to dependency


m0ksem:
For now `tslib` is a peer dependency which is wrong, because it is required for vuestic-ui to work. We need to move tslib to dependency that the user will install it automatically. 

Please, select a minimal version of `tslib` that we can require. You will see warnings while running the `build` command if you choose version 0.9.* for example.

Issue #1006, opened 2021-08-11
Fix vuestic-admin configs


m0ksem:
```
[20:37] Avantis: The latest vuestic-ui update seems to have broken vuestic-admin based projects. Most of the colors are wrong now and the default va-icon sizes arent right either.
The problem seems to be component settings from global-config dont do anything, any ideas on why that would be?
```

m0ksem:
![image](https://user-images.githubusercontent.com/23530004/129031461-0823599d-4589-48e3-b5b0-c5b69b2aed62.png)

Confirmed.

PR #1007, opened 2021-08-11
VA-998 move tslib to dependencies

alygee:
Moved the `tslib` from `devDependencies` to `dependencies` to remove the warning when installing the project.

## Description
Close https://github.com/epicmaxco/vuestic-ui/issues/998

Installed minimum version 2.2.0, same as in child library which requires this dependency ([asvae-executors](https://github.com/asvae/executors)).

image

## Types of changes

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -17004,15 +17004,15 @@ tslib@2.1.0, tslib@^2.0.3:
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.1.0.tgz#da60860f1c2ecaa5703ab7d39bc05b6bf988b97a"
   integrity sha512-hcVC3wYEziELGGmEEXue7D75zbwIIVUMWAVbHItGPx0ziyXxrOMQx4rQEVEV45Ut/1IotuEvwqPopzIOkDMf0A==
 
+tslib@2.2.0:
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.2.0.tgz#fb2c475977e35e241311ede2693cee1ec6698f5c"
+  integrity sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==
+
 tslib@^1.8.1, tslib@^1.9.0:
   version "1.14.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -17004,16 +17004,16 @@ tslib@2.1.0, tslib@^2.0.3:
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.1.0.tgz#da60860f1c2ecaa5703ab7d39bc05b6bf988b97a"
   integrity sha512-hcVC3wYEziELGGmEEXue7D75zbwIIVUMWAVbHItGPx0ziyXxrOMQx4rQEVEV45Ut/1IotuEvwqPopzIOkDMf0A==
 
+tslib@2.2.0:
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.2.0.tgz#fb2c475977e35e241311ede2693cee1ec6698f5c"
+  integrity sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==
+
 tslib@^1.8.1, tslib@^1.9.0:
```
smellyshovel, line 17012: Where does this one come from? `yarn why`?

alygee: tslib contains helper functions for TypeScript.

One of the our child libraries has this in `peerDependencies`, so our project must also have this in `**dependencies**` for everything to install correctly.

alygee: It is worth considering whether this child library is needed if it is only used in one place. It might be worth getting rid of this dependency (asvae-executors).

I encourage @asvae  to consider this issue.

m0ksem: In my opinion, tslib is actually a dependency, instead of peer dependency. This library is used in `asva-executors and it actually requires ts-lib as external in its rollup config. The easiest way is simply declare `tslib` as vuestic-ui dependency instead of updating this package.

m0ksem (approved):
Let's check how it works after merging this: https://github.com/epicmaxco/vuestic-ui/issues/1006

yarn.lock
asvae: We can make do without `asva-executors` I think, if some of it's functionality is needed it's easier to support that in context of vuestic-ui.

asvae:
@alygee can you try to compile a project, add it as a dependency to vue-cli ploject and see if it works?

asvae:
Thanks!

PR #1010, opened 2021-08-11
fix: sort api data props, events, slots, methods alphabetically

Lemonka93:
## Description
Sort documentation api data props, events, slots, methods alphabetically.

## Types of changes


smellyshovel (commented):
SortObjectByProp**s**Name -> SortObjectByPropName**s** makes more sense as per my understanding.

Lemonka93:
> SortObjectByProp**s**Name -> SortObjectByPropName**s** makes more sense as per my understanding.

Thanks, fixed.

packages/docs/src/helpers/SortObjectByPropNamesHelper.ts
```diff
@@ -0,0 +1,9 @@
+export const sortObjectByPropNames = (obj: any) => {
```
smellyshovel, line 1: No `any`s please. Better use `Record<string, unknown>`, or even something like this but in conjunction with generics.

packages/docs/src/helpers/SortObjectByPropNamesHelper.ts
```diff
@@ -0,0 +1,9 @@
...
+  const arrFromObj = Object.entries(obj);
```
smellyshovel, line 2: Consider using Object.keys, if it's possible. Should be though. We sorting objects by keys, so we shouldn't bother with values at all.

packages/docs/src/helpers/SortObjectByPropNamesHelper.ts
```diff
@@ -0,0 +1,9 @@
...
+
+  const sortedArr = arrFromObj.sort((element1: any, element2: any) => {
```
smellyshovel, line 4: Once again, no `any`s. It's gonna live like this for quite some time definitely and no one's gonna fix it if you don't do it now.

packages/docs/src/helpers/SortObjectByPropNamesHelper.ts
```diff
@@ -0,0 +1,9 @@
...
+    return element1[0].localeCompare(element2[0]);
+  });
+  const sortedObj = Object.fromEntries(sortedArr);
+  return sortedObj;
```
smellyshovel, line 8: Please, describe why you think it should be better to create a new Object instead of modifying the initial one in-place. I'm not saying that it's wrong, I'm just wondering if in-place modification could be more appropriate here.

smellyshovel (changes requested):

smellyshovel:
Actually, I might be mistaken, but as far as I remember the ES spec doesn't force implementations to respect the order the properties are specified in an object. So one browser can show those fields in the way they actually are and some another one may use a completely different algorythm.

The other solution might be to use built-in `Set`s or `Map`s. Please, research.

Derranion:
> Actually, I might be mistaken, but as far as I remember the ES spec doesn't force implementations to respect the order the properties are specified in an object. So one browser can show those fields in the way they actually are and some another one may use a completely different algorythm.
> 
> The other solution might be to use built-in `Set`s or `Map`s. Please, research.

That's right Object is just unordered collection.

Currently all major browsers loop over the properties of an object in the same order.

Though this order will still differ from declaration order.

For example you could try executing it in Google chrome console:
```
let obj = {
  "first":"first",
  "2":"2",
  "34":"34",
  "1":"1",
  "second":"second"
}


for (i in obj) { console.log(i); }
```
Result will look this way:
```
1
2
34
first
second
```

I could help you to update PR if you would like.

Just a side note: please use ESLint it will help with code style and let all project stay consistent.
( if you are using Webstorm: File -> Settings -> type "eslint" in search )

Derranion (changes requested):
Feel free to slightly modify your code according to the previous comments or just ask me to do so if you don't want.

Thanks for your work!

PR #1012, opened 2021-08-11
fix: tempfix for failing builds

smellyshovel:
The Vuestic won't serve in the book mode due to some issues with pointing to a 3d-party library that doesn't even seem to be a project's dependency.

This PR doesn't fix the broken WIP-color-pickers, though it allows getting past those issues and allows the dev-server to successfully serve the lib in the Book mode.


smellyshovel:
@kushich what is the rationale behide pointing to that `vue-color` dep (considering it isn't even listed in package.json)? Maybe I'm not aware of something crucial?

asvae:
@RVitaly1978 can you check this as well?

PR #1013, opened 2021-08-11
Added styles separately to dist

### LINKED ISSUE #1006: Fix vuestic-admin configs

m0ksem:
closed #1006

image
## Types of changes


asvae:
Thanks. I'll merge and check.

PR #1018, opened 2021-08-12
fix: sidebar links should respect changes to the primary color

### LINKED ISSUE #1017: Docs: sidebar navigation links don't respect changes to primary color
after

before

Open the docs, navigate to "Services -> Colors Config" and change the primary color. Sidebar links' colors should change respectively, but they don't.

* **Other information**

smellyshovel:
close #1017


asvae:
Seems good. Thanks.

PR #1029, opened 2021-08-16
Fix vite build

### LINKED ISSUE #1024: Vite build not recognize components
In vite dev, all works.

Vite ignores vuesitc-ui components in the build process. As I can see it is caused by preserve modules, some of js files are ignored by tree-shaking.

This can be fixed if we manually pass all components' index.ts files to rollup input. It also break esm build file structure since we need to use `manulChuks` here instead of `preservedModules`. preservedModules break .vue components to template, js and style files. 
As I can see it's a problem of rollup-plugin-vue and @vitejs/plugin-vue too (since vite extended rollup plugin).

Then we will have ugly output in esm bundle, but it works instead.
image

😭

I don't think it is important. 

Should be implemented after bundler tests will be done. https://github.com/epicmaxco/vuestic-ui/issues/1015

m0ksem:
For some reason, Vite can not build bundled lib using `preserve modules` option. So I decided to split esm bundle into chunks providing multiple inputs, that were generated using readdir of `/src/components` folder.

I checked this PR with `vite preview` command nad `vite dev`.

It also works with webpack (vue-cli).

I think this PR should be merged after tree-shaking tests is ready to prevent unexpected issues.
closes #1024

### DIFF
--- a/packages/ui/build/rollup/configs/esm.js
+++ b/packages/ui/build/rollup/configs/esm.js
@@ -7,19 +7,23 @@ import { nodeResolve as nodeResolvePlugin } from '@rollup/plugin-node-resolve'
 import typescriptDeclarationPlugin from '../plugins/rollup-typescript-declaration'
 import { terserPlugin } from '../plugins/rollup-teaser-preset'
 import { dependencies, peerDependencies } from '../utils'
+import { getInputs } from '../generate-rollup-inputs'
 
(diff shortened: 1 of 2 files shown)


m0ksem:
image

Tree-shaking also works well.

m0ksem:
image
Works fine with tests after
https://github.com/epicmaxco/vuestic-ui/pull/1037

m0ksem:
image

m0ksem:
I also tested vuestic-admin with this update, looks fine.

Issue #1035, opened 2021-08-17
ag grid styling


asvae:
https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/vuestic-ui?node-id=406%3A527

We want to provide a solution for vuestic user to

PR #1068, opened 2021-09-03
Styling simple ag-grid table

alygee:
related to https://github.com/epicmaxco/vuestic-ui/issues/1035

## Description
Added a styling that works in the context of vuestic variables/themes.   
The current solution contains styling a simple table and not separated from the bundle.  
I plan to provide a separated theme file that covers all of the grid functionality in **subsequent PRs**.

## Types of changes


packages/docs/package.json
```diff
@@ -70,10 +72,10 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-      "vue-cli-service lint --fix"
+      "vue-cli-service lint"
```
smellyshovel, line 75: Why get rid of "--fix"es?

packages/ui/src/styles/global/_typography.scss
```diff
@@ -192,6 +192,10 @@ ol.va-ordered {
   @include va-table();
 }
 
+.ag-theme-vuestic {
```
smellyshovel, line 195: Check the class name in the docs. It's `ag-grid-vuestic` there instead of `ag-theme-vuestic`

packages/docs/src/locales/ru/ru.json
```diff
@@ -3234,11 +3234,14 @@
...
     "textBlock": "Текстовый блок"
   },
-  "ag-grid-wrapper": {
-    "title": "AG Grid Wrapper",
-    "description": "Работа в процессе",
-    "related": "Связанные материалы",
-    "styledTablesLink": "Стилизованные таблицы"
+  "ag-grid": {
+    "title": "AG Grid",
+    "description": "Если нужен более гибкая настройка отображения таблиц, то мы рекомендуем воспользоваться библиотекой <a href='https://www.ag-grid.com/vue-data-grid/getting-started'>AG Grid</a>. Vuestic UI предостовляет готовую тему стилей для этой библиотеки.",
```
smellyshovel, line 3239: нужна*?

smellyshovel (changes requested):

packages/docs/package.json
alygee: because this linter has no `--fix` parameter, it starts formatting by default.  
There is only a `--no-fix` parameter  
https://cli.vuejs.org/core-plugins/eslint.html#injected-commands

packages/docs/src/locales/ru/ru.json
alygee: fixed

packages/ui/src/styles/global/_typography.scss
alygee: fixed

smellyshovel (approved):

asvae:
https://github.com/epicmaxco/vuestic-ui/pull/1068/files#diff-7760fc854ff24df74939253ca8251452539cf9d01accc0aff88bffb0b7ba8133R196 - that's incorrect. ag-grid in no way be included in main package.

Should be only something like that, most likely.
```
import `vuestic-ui/ag-grid.scss`
```

You may still use vuestic variables/mixins, so it's not independent, just one way dependency.

PR #1072, opened 2021-09-06
Add translations for code examples

alygee:
Added translation functionality in example's templates and for CodeSandbox.

## Description
I just copied the functionality into other components. I didn’t put it into the .ts file, because I think that the passing of the component's context at runtime smells bad.

## Types of changes


packages/docs/src/utilities/utils.ts
```diff
@@ -13,3 +13,18 @@ export const readComponent = async (fileName: string): Promise<any> => {
...
     `../examples/${fileName}.vue`)
 }
+
+/**
+* Find `$t` inside a code and replace it with translation by regex.
+* origin: `... $t('to replace') ...`
+* output: `... translated ...`
+*/
+export function applyTranslations (this: any, code: string) {
+  const replaces = code.match(/(?:\$t)\(.*?\)/) || []
```
RVitaly1978, line 23: here we need to add a flag `g`:
`const replaces = code.match(/(?:\$t)\(.*?\)/g) || []`

packages/docs/src/utilities/utils.ts
```diff
@@ -13,3 +13,18 @@ export const readComponent = async (fileName: string): Promise<any> => {
...
+
+  return replaces.reduce((acc, replaceSource) => {
+    const translation = replaceSource.replace(/(\$t|'|\(|\)|\[\d\])/gi, '')
+
+    return acc.replace(replaceSource, this.$t(translation))
```
RVitaly1978, line 28: here we need to replace as `string`, because our `code` prop is bound with `v-bind` directive:
``return acc.replace(replaceSource, `'${this.$t(translation)}'`)``

packages/docs/src/components/DocsExample.vue
```diff
@@ -21,12 +21,12 @@
         />
         <DocsCode
           language="markup"
-          :code="parsed.template"
+          :code="applyTranslations(parsed.template)"
```
RVitaly1978, line 24: we shoudn't apply `applyTranslations` here, because we applied it in component `DocsCode`
`:code="parsed.template"`

packages/docs/src/components/DocsExample.vue
```diff
@@ -21,12 +21,12 @@
...
           :class="[parsed.script ? 'docs-example__code--with-margin' : '']"
         />
         <DocsCode
           v-if="parsed.script"
-          :code="parsed.script"
+          :code="applyTranslations(parsed.script)"
```
RVitaly1978, line 29: same as above

packages/docs/src/components/DocsExample.vue
```diff
@@ -39,7 +39,7 @@
 // import VaContent from '../../ui/src/components/va-content/VaContent'
 import DocsCode from './DocsCode'
 import DocsNavigation from './DocsNavigation'
-import { readComponent, readTemplate } from '../utilities/utils'
+import { readComponent, readTemplate, applyTranslations } from '../utilities/utils'
```
RVitaly1978, line 42: `applyTranslations ` not needed in this component

packages/docs/src/components/DocsExample.vue
```diff
@@ -111,6 +111,7 @@ export default {
       const parsed = regex.exec(template) || []
       return parsed[1] || ''
     },
+    applyTranslations,
```
RVitaly1978, line 114: same as above

RVitaly1978 (commented):
Good job!

RVitaly1978 (approved):
Overall looks good, thanks.

packages/docs/src/components/tests/DocsCode.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      <div>
+        <some-component :prop="$t('passed prop')" />
+        <some-component>{{ $t('passed slot') }}</some-component>
+      </div>
+    </template>`
+
+    const OUTPUT_CODE = `<template>
+      <div>
+        <some-component :prop="'passed prop'" />
+        <some-component>{{ 'passed slot' }}</some-component>
```
asvae, line 42: possible to have `passed slot` in stead of `{{ 'passed slot' }}`? We discussed this before.

Issue #1069, opened 2021-09-03
Provide a AG Grid them separated from plugin


alygee:
Parent issue https://github.com/epicmaxco/vuestic-ui/issues/1035

At now Vuestic supports styling a simple table of `AG Grid` library.
But these styles are implemented as sass-mixin and are included in the plugin.

It is needs to provide a separate theme as separated file for the AG Grid (e.g. `dist/styles/ag-theme-vuestic.css`).
This will allow:
1) Set the minimum set of rendering parameters (rowHeight, headerHeight and minColWidth). They don't need to be set every time you use the plugin.
2) The theme can be easily extended to the rest of the library functionality, without the risk of breaking the plugin's behavior.

Useful links for creating a theme:
[Official docs](https://www.ag-grid.com/vue-data-grid/themes-customising)
[repo with examples](https://github.com/ag-grid/ag-grid-customise-theme)

Issue #1075, opened 2021-09-07
Removing node-sass as it is a duplicate dependency.


alygee:
Vuestic plugin and docs's project contain `sass` and `node-sass` dependencies. These libraries both do the same thing (`node-sass` is considered more faster, but `sass` considered more reliable).

However `node-sass` is [deprecated](https://sass-lang.com/blog/libsass-is-deprecated). Therefore possible compatibility and integration errors  with modern libraries. For example, it is not possible to use the functionality of the provided themes from AG Grid, when using `node-sass`.

I suggest to remove `node-sass` as a duplicate and unreliable library.

PR #1077, opened 2021-09-08
Provide separated theme for AG Grid

alygee:
close https://github.com/epicmaxco/vuestic-ui/issues/1069, https://github.com/epicmaxco/vuestic-ui/issues/1075

## Description
Provided `ag-theme-vuestic` theme, separated from the plugin.
Changed rollup config for style's build.
Fixed some scss warnings during build.
Removed node-sass dependency.

## Types of changes


packages/ui/package.json
```diff
@@ -75,7 +75,7 @@
     "@vue/compiler-sfc": "^3.1.4",
     "@vue/eslint-config-standard": "^6.0.0",
     "@vue/eslint-config-typescript": "^7.0.0",
-    "@vue/test-utils": "^2.0.0-beta.14",
+    "@vue/test-utils": "2.0.0-beta.14",
```
m0ksem, line 78: We need the `ag-grid-community` package as dependency or peer dependency if it is not required in vuestic-ui. 

For now, you have ag-grid-community in our docs package, and because it is monorepo it also available in the UI package.

m0ksem (changes requested):
image
Got this strange padding between columns in Firefox.

Also, can you make more examples. As I know there is a lot of feature that ag-grid has and its change visual as well. For example, multirow headers, resizable columns.

packages/ui/src/components/va-table/AgGrid.demo.vue
```diff
@@ -67,5 +65,6 @@ export default {
 </script>
 
 <style lang="scss">
-@import "ag-grid-community/dist/styles/ag-grid.css";
+@import "../../../../../node_modules/ag-grid-community/dist/styles/ag-grid.css";
+@import "vuestic-ui/dist/ag-theme-vuestic.css";
```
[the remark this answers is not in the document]
alygee: I changed path from `dist` to `src`, and made it relative. Also removed paddings between columns.

This PR is for providing a theme separated from the plugin.
Other examples with basic functionality will be in the future PR accoring with this issue  https://github.com/epicmaxco/vuestic-ui/issues/1078

packages/ui/package.json
```diff
@@ -93,7 +94,6 @@
     "jsdom": "^16.2.2",
     "lint-staged": "^11.1.2",
     "mini-css-extract-plugin": "^1.3.0",
-    "node-sass": "^5.0.0",
```
asvae, line 96: sure that's safe to remove?

asvae:
image

Can you check styling to be the same?
Right now colors are a bit off + font is not the same.

packages/ui/package.json
m0ksem: We've had a `sass` (dart-sass) and `node-sass`. As I know node-sass is faster, but deprecated from the recent past. `node-sass` doesn't support built-in modules such `sass:math`. 


We use webpack sass-loader plugin which should handle what package should we use. For now, they use both node-sass and sass.

Speaking about rollup, we handle sass with postcss plugin. 

For now, it is a safe remove, because we have this dependency in sass-loader. When it becomes deprecated we won't be needing this (and even handle this).

alygee:
> image
> 
> Can you check styling to be the same?
> Right now colors are a bit off + font is not the same.

The difference in pictures is due to different project settings.

packages/ui/package.json
alygee: Moved `ag-grid-community, ag-grid-vue3` dependencies to peerDependencies

packages/docs/src/examples/grid/AgGrid.vue
```diff
@@ -62,5 +60,6 @@ export default {
 </script>
 
 <style lang="scss">
-@import "ag-grid-community/dist/styles/ag-grid.css";
+@import "../../../../../node_modules/ag-grid-community/dist/styles/ag-grid.css";
```
m0ksem, line 63: I know we are not able to show the style tag for now. But I think this is important to be shown in this demo. It would be nice to use Webpack aliases there instead of ugly relative paths. We can do it later, but as far as I know, we want to make it so that users can copy this demo code, paste it into their own project and it should work fine.

```suggestion
@import "ag-grid-community/dist/styles/ag-grid.css";
@import "vuestic-ui/styles/ag-theme-vuestic/ag-theme-vuestic.scss";
```

m0ksem: I think it is not important for now. You can skip it.

packages/ui/src/styles/ag-theme-vuestic/_ag-theme-vuestic-default-params.scss
```diff
@@ -0,0 +1,28 @@
+@import "~ag-grid-community/dist/styles/ag-theme-base/sass/ag-theme-base-default-params";
+
+$ag-theme-vuestic-default-params: (
```
m0ksem, line 3: There is stylelint error. Can resolve it? I suggest removing turn off this rule because it works incorrectly with SCSS maps and it is not important for as.

m0ksem (changes requested):

packages/docs/src/locales/en/en.json
```diff
@@ -3236,7 +3236,7 @@
   },
   "ag-grid": {
     "title": "AG Grid",
-    "description": "If you need more flexible customization of the display of tables, we recommend using the <a href='https://www.ag-grid.com/vue-data-grid/getting-started'> AG Grid </a> library. Vuestic UI provides a ready-made style theme for this library.",
+    "description": "If you need more flexible customization of the display of tables, we recommend using the [AG Grid](https://www.ag-grid.com/vue-data-grid/getting-started') library. Vuestic UI provides a ready-made style theme for this library.",
```
RVitaly1978, line 3239: to make link on external source we adds `[[target=_blank]]` to markdown:
`[AG Grid](https://www.ag-grid.com/vue-data-grid/getting-started)[[target=_blank]]`

packages/docs/src/locales/ru/ru.json
```diff
@@ -3236,7 +3236,7 @@
   },
   "ag-grid": {
     "title": "AG Grid",
-    "description": "Если нужна более гибкая настройка отображения таблиц, то мы рекомендуем воспользоваться библиотекой <a href='https://www.ag-grid.com/vue-data-grid/getting-started'>AG Grid</a>. Vuestic UI предостовляет готовую тему стилей для этой библиотеки.",
+    "description": "Если нужна более гибкая настройка отображения таблиц, то мы рекомендуем воспользоваться библиотекой [AG Grid](https://www.ag-grid.com/vue-data-grid/getting-started'). Vuestic UI предостовляет готовую тему стилей для этой библиотеки.",
```
RVitaly1978, line 3239: same as above.
No quotation mark (`'`) is needed at the end of the path

PR #1089, opened 2021-09-13
Fix default collapse mode-value. Fix collapse unordered animation.

### LINKED ISSUE #1063: Collapse when mounted with v-model true is still closed
Go to `VaCollapse.demo.vue` and set

```
collapseValue: true,
```

Collapses are closed even though icon points upward. They should be open.
image

m0ksem:
## Description
closes #1063

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-collapse/VaCollapse.vue
+++ b/packages/ui/src/components/va-collapse/VaCollapse.vue
@@ -209,7 +209,7 @@ export default class VaCollapse extends mixins(
     this.setCollapseParams()
 
     this.mutationObserver = new MutationObserver(() => {
-      setTimeout(() => this.setCollapseParams(), 1000 / 60)
+      setTimeout(() => this.setCollapseParams(), 16)
     })
 
     this.mutationObserver.observe(this.body, {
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/va-collapse/VaCollapse.vue
```diff
@@ -207,14 +206,18 @@ export default class VaCollapse extends mixins(
   mounted () {
     this.getHeight()
 
+    this.setCollapseParams()
+
     this.mutationObserver = new MutationObserver(() => {
-      this.setCollapseParams()
+      setTimeout(() => this.setCollapseParams(), 1000 / 60)
```
asvae, line 212: Why 1000/60? Very peculiar timing.

asvae (approved):
One comment, but otherwise mergeable :).

packages/ui/src/components/va-collapse/VaCollapse.vue
m0ksem: 16ms. I thought 1000 / 60fps will be much readable than 16ms.

m0ksem, line 212:
```suggestion
      setTimeout(() => this.setCollapseParams(), 16)
```

PR #1095, opened 2021-09-16
Add style section to DocsCode

### LINKED ISSUE #1079: Missed styles block in code examples
is expected that example of code will render the entire vue-component.

At now code example are missing a block with styles. This can make the example incomplete and create ambiguity.

Screenshot from 2021-09-08 16-02-45

kushich:
closes #1079

## Types of changes


asvae:
See discussion in https://github.com/epicmaxco/vuestic-ui/issues/1079 before we continue with this MR.

For now, I would rather make it throw an error like ('please use helpers or inline styles instead of using `<styles>` in examples'), as I hardly see the case where `<styles>` are not a mistake.

m0ksem:
Can we make a flag if we want to show styles block? It can be also really useful in demos with CSS variables or Color Config. 

For example:
```ts
DocsHelper.example('colors-config/css-variable', { showStyleBlock: true }),
```

PR #1106, opened 2021-09-17
Add va-data-table docs

RVitaly1978:
## Description
Add docs (with translations and examples) to `va-data-table` component (see [PR](https://github.com/epicmaxco/vuestic-ui/pull/956))

## Types of changes


packages/docs/src/components/page-configs/ui-elements/va-data-table/api-options.ts
```diff
@@ -0,0 +1,36 @@
+import { defineManualApi } from '../../../DocsApi/ManualApiOptions'
```
asvae, line 1: image

asvae: Can you check if we can maybe decrease script size by moving data somewhere?

image

What comes to my mind is some simple API call. You can probably check how it's done for other frameworks. Seems unlikely to me that they include full data in each example.

packages/docs/src/components/page-configs/ui-elements/va-data-table/api-options.ts
```diff
@@ -0,0 +1,36 @@
...
+
```
asvae, line 2: Can you add a bit of text here? Looks pretty confusing to me.

packages/docs/src/locales/en/en.json
```diff
@@ -3242,9 +3289,43 @@
...
+        "title": "Basic usage",
+        "text": "The standard `va-data-table` will by default render your data as simple rows."
+      },
+      "slots": {
+        "title": "Slots",
+        "text": "Slots allow you to control the display of data and their views in different parts of the table. The examples below demonstrate some of these slots and what you can do with each of them."
+      },
+      "filtering": {
+        "title": "Filtering",
+        "text": "`va-data-table` allows to apply a single table-wise filter with the `filter` prop. To use a custom filtering function provide the `va-data-table` with the `filtering-fn` prop.\n\nThe following example shows the use of a custom filtering function (which looks for the exact match rather than the standard one which checks for substring-inclusion) when the respective checkbox is checked."
```
asvae, line 3306: table-wide?

packages/ui/src/components/va-data-table/VaDataTable.new.demo.vue
```diff
@@ -0,0 +1,321 @@
...
+          <col width="50">
+        </template>
+      </va-data-table>
+    </VbCard>
+
+    <VbCard title="Filtering">
+      <p>Number of filtered items: {{ filteredCount }}</p>
+      <input type="text" v-model="filter">
+
+      <input id="ucf" type="checkbox" v-model="useCustomFilteringFn">
```
asvae, line 54: Please don't use abbreviations without very good reason. Fixing it if another developer used them is also a good idea.
🤓

packages/ui/src/components/va-data-table/VaDataTable.new.demo.vue
```diff
@@ -0,0 +1,321 @@
...
+      <label for="ucf">Use custom filtering function (looks for an exact match)</label>
+
+      <va-data-table
+        :items="items"
+        :filter="filter"
```
asvae, line 59: I would suggest `filter-text` and `filter` (as function), instead of current naming.
Checking other solutions would make sense though. But `filter` implies for me we're dealing with function.

packages/ui/src/components/va-data-table/VaDataTable.new.demo.vue
```diff
@@ -0,0 +1,321 @@
...
+        :filtering-fn="customFilteringFn"
+        @filter="filteredCount = $event.length"
```
asvae, line 61: Having the same name for prop and event is not the greatest idea.
`@filtered` maybe?

packages/docs/src/examples/va-data-table/Filtering.vue
```diff
@@ -0,0 +1,179 @@
...
+        company: {
+          name: 'Keebler LLC',
+          catchPhrase: 'User-centric fault-tolerant solution',
+          bs: 'revolutionize end-to-end systems',
+        },
+      },
+    ]
+
+    return {
+      items: cloneDeep(users),
```
asvae, line 160: cloneDeep doesn't seem to be of any use?

packages/docs/src/components/page-configs/ui-elements/va-data-table/api-options.ts
```diff
@@ -0,0 +1,36 @@
...
+export default defineManualApi({
```
asvae, line 3: image

Can we use less bright color? :D

RVitaly1978: we use `--selected-color` calculated by `getFocusColor` (from our color-config functions) based on passed by user `selectedColor` prop. And the `focus-color` in this case is  quite a bit brighter than the `hover-color`. I think its a good decision.
Maybe, in this example I use less bright color in `selectedColor` prop.

packages/docs/src/components/page-configs/ui-elements/va-data-table/api-options.ts
RVitaly1978: I think it is more convenient for the user to see the data in the example code.
So, to decrease script size I removed redundant data

packages/docs/src/examples/va-data-table/Filtering.vue
RVitaly1978: Fixed this

packages/ui/src/components/va-data-table/VaDataTable.new.demo.vue
RVitaly1978: fixed this

packages/docs/src/locales/en/en.json
RVitaly1978: fixed this

packages/docs/src/components/page-configs/ui-elements/va-data-table/api-options.ts
RVitaly1978: I added a bit explain text to each example

PR #1119, opened 2021-09-22
Fix keyboard navigation for file upload.

### LINKED ISSUE #975: Validation on file input error
Seems like validate an input file doesn't work with standar validation in vuestic...it gives me this error ->  **TypeError: o.filter is not a function**.

This is my input:

`<va-file-upload v-model="avatar" type="gallery" file-types=".jpg,.png" dropzone />`

[CAPTION](http://i.imgur.com/WAonLsz.png)

m0ksem:
## Description
dKbs6vC3RQ

At first, I thought we have a problem with VaFileUpload because of vue-class-component, so I rewrite VaFileUpload to composition API. But, later I realized that there is no problem at all. So in this PR I fixed broken keyboard navigation and some small bugs.

close #975

### DIFF
--- a/packages/ui/src/components/va-file-upload/VaFileUpload.vue
+++ b/packages/ui/src/components/va-file-upload/VaFileUpload.vue
@@ -22,13 +22,14 @@
         Upload file
       </va-button>
       <input
-        ref="fileInput"
+        ref="fileInputRef"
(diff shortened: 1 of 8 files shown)


asvae:
Thanks!

PR #1149, opened 2021-10-04
Fix#882 Fix dateInput paddings

### LINKED ISSUE #822: Button cannot be clicked outside content

zvenigorodskaia:
## Description

close #822

Added width:100% to .va-dropdown class

Before:
image

After:
image

## Types of changes
- [ ]  New feature (non-breaking change which adds functionality)


packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -296,4 +296,8 @@ export default class VaDropdown extends mixins(
     z-index: var(--va-dropdown-content-wrapper-z-index);
   }
 }
+
+.va-dropdown:not(.va-popover) {
```
asvae, line 300: If you want to create and exception for popover (dropdown width shouldn't be 100%), then it's better to be done in context of Popover, as dropdown is Popover dependency.
Dropdown shoudn't even know about Popover existance.

packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -296,4 +296,8 @@ export default class VaDropdown extends mixins(
     z-index: var(--va-dropdown-content-wrapper-z-index);
```
asvae, line 296: Button dropdown looks kinda broken :).

http://localhost:8080/#/demo/va-button-dropdown/VaButtonDropdown.demo.vue

asvae:
Looks good. Thanks!

PR #1218, opened 2021-11-07
Fix #1217: Can't select checkbox with space button

### LINKED ISSUE #1217: Can't select checkbox with space button
http://vuestic-ui-demo.sub.asva.by/#/demo/va-checkbox/VaCheckbox.demo.vue

When I select with keyboard
and I press Space
Then Checkbox is being selected twice
But it only should be selected once

d4rkm3z:
## Description
close: #1217
Added stop propagation to the checkbox for click events

## Types of changes


asvae:
Thanks! That works.

PR #1232, opened 2021-11-12
Implements closeOnAnchorClick on va-select (issue epicmaxco/vuestic-admin#699)

Gabriel-scodiero:
close epicmaxco/vuestic-admin#699

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.vue
+++ b/packages/ui/src/components/va-select/VaSelect.vue
@@ -7,6 +7,7 @@
     :max-height="$props.maxHeight"
     :fixed="$props.fixed"
     :close-on-content-click="closeOnContentClick"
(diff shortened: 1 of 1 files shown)


asvae:
Thanks!

We'd need to provide some docs to this though.

Derranion (changes requested):
Thanks,

We need to change this a bit though because it will not work while `Dropdown` `trigger` property is set to `none` inside our `Select`.

Most likely this is done because we handle `Dropdown` visibility inside the `Select` that's why `Dropdown` logic is blocked by 'none' trigger - would be great to check this and change code to compensate.

Should be possible to move this part of logic away from `Select` and just use `Dropdown` internal logic.

vaDropdownTriggerNone

m0ksem:
This looks fixed.

Issue #1205, opened 2021-11-04
Pagination looks weird


asvae:
![image](https://user-images.githubusercontent.com/5394573/140294309-7bf5d812-3698-4642-a201-8e6fecc8fbec.png)

In my opinion, outline should look the same, only icons could have decreased opacity.

@xx13

PR #1262, opened 2021-11-24
Fix#1205 Directly changed the opacity for buttons with disabled state

ishpartko:
## Description

Directly changed the opacity for buttons with disabled state.

close: #1205

## Types of changes


asvae (approved):
👍 congrats on first mr!

PR #1264, opened 2021-11-25
Fix#1204 Add VaDataTable events

### LINKED ISSUE #1204: Datatable click events
Original post:
> Looks great! Is there a way to make the rows clickable, so that clicking on a row goes to a link or runs a method?

We probably need two events:
* row click
* row double click

RVitaly1978:
## Description
close #1204 

Changes:
- [x] - add `clickable` prop
- [x] - add table row events: `click`, `dblclick`, `contextmenu`
- [x] - update `sorted` and `filtered` emit arguments (add array of items initial indexes)

chrome-capture

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/data-table/api-options.ts
+++ b/packages/docs/src/page-configs/ui-elements/data-table/api-options.ts
@@ -34,6 +34,15 @@ export default defineManualApi({
     'update:sortingOrder': {
       types: '`() => TSortingOrder`',
     },
+    'row:click': {
+      types: '`() => RowClickEvent`',
+    },
+    'row:contextmenu': {
+      types: '`() => RowClickEvent`',
+    },
+    'row:dblclick': {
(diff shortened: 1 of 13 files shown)


packages/docs/src/page-configs/ui-elements/data-table/api-options.ts
```diff
@@ -20,7 +20,7 @@ export default defineManualApi({
   },
   events: {
     filtered: {
-      types: '`() => Number`',
+      types: '`() => Object`',
```
asvae, line 23: Can't we be more specific here?

asvae (approved):
Looks perfect! Thanks.

packages/docs/src/page-configs/ui-elements/data-table/api-options.ts
RVitaly1978: fixed it

Issue #1260, opened 2021-11-24
Search doesn't work properly in docs


untael:
Hints are not shown.
![image](https://user-images.githubusercontent.com/11652328/143233764-8c0e6533-36f8-4390-a08d-d383f8a7d83b.png)
Сursor moves to the end after deleting symbols.
![VV8juYx](https://user-images.githubusercontent.com/11652328/143234373-864b8c85-49bd-4d1b-a3c9-0153a0d4ac85.gif)

PR #1272, opened 2021-11-26
Fix#1260 Fix VaInput id attr and algolia-search input selector

RVitaly1978:
## Description
close: #1260

Changes:
- [x] - fix `va-input` id attr
- [x] - fix  `algolia-search` input selector

chrome-capture

## Types of changes

### DIFF
--- a/packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
+++ b/packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
@@ -49,7 +49,7 @@ export default class Search extends Vue {
         apiKey: 'be3528055c92da2ea5133b93ed548c6d',
         indexName: 'vuestic',
       }, {
-        inputSelector: '#algolia-search-input .va-input__content__input',
+        inputSelector: '#algolia-search-input',
         debug: true,
       }))
     })
--- a/packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
+++ b/packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
@@ -2,6 +2,7 @@
   <form id="search-form">
     <va-input
       id="algolia-search-input"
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
```diff
@@ -48,7 +49,7 @@ export default class Search extends Vue {
         apiKey: 'be3528055c92da2ea5133b93ed548c6d',
         indexName: 'vuestic',
       }, {
-        inputSelector: '#algolia-search-input',
+        inputSelector: '#algolia-search-input .va-input__content__input',
```
m0ksem, line 52: image

id should be removed from input-wrapper. This may cause errors in future for end user.

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -137,6 +137,7 @@ export default defineComponent({
       ...validationListeners,
       class: attrs.inputClass,
       style: attrs.inputStyle,
+      id: attrs.inputId,
```
m0ksem, line 140: Better use id and wrapperId, because usually we want to set id to input tag instead of va-input root element.

m0ksem (changes requested):

packages/docs/src/components/sidebar/algolia-search/AlgoliaSearch.vue
RVitaly1978: fixed it

packages/ui/src/components/va-input/VaInput.vue
RVitaly1978: fixed: set all `attrs` (except `class` and `style`) to `input` tag. Only `class` and `style` `attrs` are set to `va-input`

m0ksem (approved):

PR #1280, opened 2021-11-29
Fix#1279 displayed in the local copy of the documentation

### LINKED ISSUE #1279: Component examples are not displayed in the local copy of the documentation
The components are displayed as in the production.

image

ishpartko:
## Description
closes #1279 

## Types of changes

### DIFF
--- a/packages/docs/src/utilities/utils.ts
+++ b/packages/docs/src/utilities/utils.ts
@@ -1,19 +1,26 @@
 // File for documentation helper functions
 
+function getUIElementNameFromPath (path: string): string {
+  // 'src\page-configs\ui-elements\component-name' --> 'component-name'
+  // we need to use folder names instead of full paths to the component
+  return path.match(/[\\/]([\w-_]+)$/)?.[1] || path || ''
+}
+
 export const readTemplate = async (path: string, fileName: string): Promise<any> => {
+  const uiElementName = getUIElementNameFromPath(path)
   return await import(
     /* webpackChunkName: "examples" */
-    /* webpackMode: "lazy-once" */
-    `!raw-loader!../${path}/examples/${fileName}.vue`)
-    // `!raw-loader!../page-configs/${path}/examples/${fileName}.vue`)
+    /* webpackMode: "lazy" */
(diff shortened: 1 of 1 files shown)


packages/docs/src/utilities/utils.ts
```diff
@@ -1,25 +1,31 @@
 // File for documentation helper functions
 
+function getUIElementNameFromPath (path: string): string {
+  // 'src\page-configs\ui-elements\component-name' --> 'component-name'
```
asvae, line 4: Can you make it explicit that's a windows issue?
Pretty important in my opinion.

ishpartko:
@asvae I wrote a comment, I hope it is exhaustive, but if not, I can add more :)

packages/docs/src/utilities/utils.ts
```diff
@@ -1,25 +1,39 @@
 // File for documentation helper functions
 
+function getUIElementNameFromPath (path: string): string {
```
asvae, line 3: Very nitpicky, but you can use `const getUIElementNameFromPath = () => {}` here 🤓

packages/docs/src/utilities/utils.ts
```diff
@@ -1,25 +1,39 @@
...
+  * work's example:
+  * 'src\page-configs\ui-elements\component-name' --> 'component-name'
+  * why:
+  * it is only a guess that the path written through the relative operator "../"
+  * works differently in unix and windows and for fully identical operation
+  * in webpack we have to pass a more specific path as specified in the documentation
+  * https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import
+   * TODO: To remove this hack, we need to pass the name of the folder from which the webpack should import in each case instead of path
+  **/
+  return path.match(/[\\/]([\w-_]+)$/)?.[1] || path || ''
```
asvae, line 14: Just wondering, if we can maybe do nothing here, and webpack will still import everything just in case (as it likely won't be able to determine what's getting loaded).

But up to you if you want to spend anymore time on that (curiosity), as the issue is low prio (almost no one uses windows).

asvae (approved):

asvae:
Thanks! lgtm

Issue #1302, opened 2021-12-03
VaSelect width is too big


m0ksem:
https://vuestic.dev/en/ui-elements/form#trigger-validation

![image](https://user-images.githubusercontent.com/23530004/144670723-940bca40-880b-4127-bf2a-f7fe91b7b3e9.png)

Issue #1291, opened 2021-12-01
Select should use useValidation hook


m0ksem:
`VaForm` do not validate `VaSelect` when `VaForm`'s validate method called.

In example below in console first boolean is `validate` method result.

As you can see, select gets error state after blur, but don't get it after `validate` method call.

![WRMS6pMTvF](https://user-images.githubusercontent.com/23530004/144214165-356b68c6-1962-4a43-8974-217a0a1588a0.gif)

Issue #1285, opened 2021-11-30
Select not working in docs


m0ksem:
Select not working

https://vuestic.dev/en/ui-elements/form#trigger-validation

m0ksem:
@RVitaly1978, please consider to have a look at that issue. 

As I can see, stateful prop in select broke dropdown.

Issue #1281, opened 2021-11-29
Form validatation method don't validate VaSelect


untael:
$refs.form.validate() doesn't validate VaSelect in way as VaInput

PR #1307, opened 2021-12-06
Fix#1281 #1302 #1291 Fix VaSelect validation

RVitaly1978:
## Description
close: #1281
close: #1302
close: #1285
close: #1291

Changes:
- [x] - update `useValidation` and `useFormComponent` hooks
- [x] - fix `va-form` and `va-input` doc examples

chrome-capture

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/VaInput.vue
+++ b/packages/ui/src/components/va-input/VaInput.vue
@@ -122,7 +122,7 @@ export default defineComponent({
       listeners: validationListeners,
       computedError,
       computedErrorMessages,
-    } = useValidation(props, emit, () => reset())
+    } = useValidation(props, emit, () => reset(), () => focus(), () => blur())
 
     const canBeCleared = computed(() => {
       return props.clearable && ![null, undefined, ''].includes(props.modelValue as any)
@@ -158,6 +158,14 @@ export default defineComponent({
       emit('cleared')
     }
 
+    const focus = () => {
+      input.value?.focus()
+    }
+
+    const blur = () => {
+      input.value?.blur()
+    }
+
     return {
       input,
       textareaProps: filterComponentProps(props, VaTextareaProps),
@@ -177,12 +185,15 @@ export default defineComponent({
       computedInputAttributes,
       fieldListeners: createFieldListeners(emit),
       reset,
+      focus,
+      blur,
     }
   },
 
+  // we will use this while we have 'withConfigTransport' and problem with 'expose' method in 'setup' func
   methods: {
-    focus () { this.input?.focus() },
-    blur () { this.input?.blur() },
+    _focus () { this.input?.focus() },
+    _blur () { this.input?.blur() },
   },
 })
 </script>
--- a/packages/ui/src/components/va-input/VaInput.vue
+++ b/packages/ui/src/components/va-input/VaInput.vue
@@ -114,8 +114,21 @@ export default defineComponent({
 
   inheritAttrs: false,
 
-  setup (props, { emit, attrs }) {
+  setup (props, { emit, attrs, expose }) {
     const input = ref<HTMLInputElement | HTMLTextAreaElement>()
+
+    const reset = () => {
+      emit('update:modelValue', '')
+      emit('cleared')
+    }
+
+    const focus = () => {
+      input.value?.focus()
+    }
+
+    const blur = () => {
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -177,12 +185,15 @@ export default defineComponent({
...
+      focus,
+      blur,
     }
   },
 
+  // we will use this while we have 'withConfigTransport' and problem with 'expose' method in 'setup' func
   methods: {
-    focus () { this.input?.focus() },
-    blur () { this.input?.blur() },
+    _focus () { this.input?.focus() },
```
m0ksem, line 195: Better use `_focus` in setup and define allow user to use focus method using ref, instead of _focus.

image
image

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -158,6 +158,14 @@ export default defineComponent({
       emit('cleared')
     }
 
+    const focus = () => {
```
m0ksem, line 161: Minor. This should be on top of useValidation, because it used as callbacks here.

Usually, it's better to define function on top of its usage.

packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -499,7 +506,8 @@ export default defineComponent({
     }
 
     const focusSearchBar = () => {
-      searchBar.value?.focus()
+      // we will replace '_focus' with 'focus' when will resolve 'withConfigTransport' problem with 'expose' in 'setup' func
+      searchBar.value?._focus()
```
m0ksem, line 510: As you can see, it  bit weird to use _focus here.

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -101,14 +102,28 @@ export default defineComponent({
...
+    }
+
+    const input = ref<InstanceType<typeof VaInput>>()
+
+    const focus = (): void => {
+      // we will replace '_focus' with 'focus' when will resolve 'withConfigTransport' problem with 'expose' in 'setup' func
+      input.value?._focus()
+    }
+
+    const blur = (): void => {
```
m0ksem, line 116: These methods are not exposed because of withTransportConfig bug. You should use here the same approach with Options API methods.

packages/ui/src/composables/useFormComponent.ts
```diff
@@ -125,6 +118,19 @@ export function useFormComponent (props: Record<string, any>, context: any) {
...
+  const formContext = {
+    resetValidation,
+    validate,
+    reset,
+    hasError,
+    focus,
+    blur,
+  }
+
+  onMounted(() => formProvider?.onChildMounted?.(formContext))
```
m0ksem, line 130: Double check if we need both hooks useValidation and useFormComponent after this fix, please.

m0ksem (changes requested):
Please, expose focus and blur for all components and replace "_focus" with "focus" in exposed methods.

👍

packages/ui/src/components/va-input/VaInput.vue
RVitaly1978: Fixed it

packages/ui/src/components/va-input/VaInput.vue
RVitaly1978: Done

packages/ui/src/components/va-select/VaSelect.vue
RVitaly1978: Fixed it

packages/ui/src/components/va-time-input/VaTimeInput.vue
RVitaly1978: Done

packages/ui/src/composables/useFormComponent.ts
RVitaly1978: In this hotfix, I will leave both hooks. But they do the same, so we need to choose one of them.

asvae (approved):
All good apart from this small thing:

image

m0ksem (approved):
useValidation hook makes validation imidiately.

image


Can you, please, also add here imidiateValidation prop and set it false buy default?

image

Issue #1282, opened 2021-11-29
Ability to emit created values by VaSelect


untael:
`allow-create` prop should also emit created value

PR #1315, opened 2021-12-07
Fix#1282 Add VaSelect 'create-new' emit

RVitaly1978:
## Description
close: #1282

Changes:
- [x] - add `create-new` emit
- [x] - add translations
- [x] - update doc and demo examples

chrome-capture

## Types of changes


packages/docs/src/locales/en/en.json
```diff
@@ -758,7 +758,8 @@
       },
```
asvae, line 758: On mouse down:

image

Normal:

image

You can see outline on click. That shouldn't happen.

packages/docs/src/locales/en/en.json
```diff
@@ -758,7 +758,8 @@
...
       "events": {
```
asvae, line 759: When i focus select and i press tab - I expect next select to be focused.
Instead focus is lost.

packages/docs/src/locales/en/en.json
```diff
@@ -758,7 +758,8 @@
...
         "clear": "Emitted if select value has been cleared",
```
asvae, line 760: Pressing Enter should clear search text when allow-create flag is true.

image

packages/docs/src/locales/en/en.json
```diff
@@ -758,7 +758,8 @@
...
-        "updateSearch": "Emitted if search value has changed"
```
asvae, line 761: When I press UP button - cursor goes directly to "five" option.
It should not ignore the option in-between even though text value is the same (id is not!)

image

asvae:
Found a couple of things - let's see if we can fix these quickly. If no - let's move fixing these to another PR (apart from shadow things are not critical)

packages/docs/src/locales/en/en.json
RVitaly1978: Fixed it

packages/docs/src/locales/en/en.json
RVitaly1978: Fixed it

packages/docs/src/locales/en/en.json
RVitaly1978: Let's move fixing these to another PR

packages/docs/src/locales/en/en.json
RVitaly1978: I have this behavior when press `tab`
chrome-capture

packages/docs/src/locales/en/en.json
RVitaly1978: fixed it:
chrome-capture

asvae:
Thanks 🤗

PR #1320, opened 2021-12-07
Fix #1226 Fix VaTimeInput bug when passed text is not valid

### LINKED ISSUE #1226: Timepicker breaks down after writing text
CU34Ef0

Also there is a question: should we have possibility to write text?

RVitaly1978:
## Description
close: #1226 

Changes:
- [x] - fix `va-time-input` bug when passed text (prop `manual-input` enable) is not valid
- [x] - tweak regexp in `useTimeParser` hook

chrome-capture

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -15,7 +15,7 @@
         :readonly="readonly || !manualInput"
         :error="!isValid || inputProps.error || computedError"
         :error-messages="computedErrorMessages"
-        @change="onInputTextChanged"
+        @change="onInputTextChanged($event.target.value)"
       >
         <template v-for="(_, name) in $slots" v-slot:[name]="bind">
           <slot :name="name" v-bind="bind"></slot>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -129,6 +129,12 @@ export default defineComponent({
       blur,
     })
 
+    watch(modelValueSync, () => {
+      if (!isValid.value && valueText.value === '') {
+        isValid.value = true
+      }
+    })
```
m0ksem, line 136:
```suggestion
    watch(modelValueSync, () => {
      isValid.value = true
    })
```

This condition looks useless and makes modelValue required not to be null.

1. `if (isValid == false) isValid = true`
same as 
`isValid = true`

2. `valueText.value == ''` when `isValid is false` or `modelValue is null`
By default, modelValue is not required and validation of required rule controlled by VaInput, and it's text.


Double check if I correct.

m0ksem (changes requested):

packages/ui/src/components/va-time-input/VaTimeInput.vue
RVitaly1978: fixed it

m0ksem (approved):

PR #1344, opened 2021-12-13
Fix behavior of select dropdown when tab key pressed

RVitaly1978:
## Description
Changes:
- [x] - set focus to select when tab key pressed (when dropdown open)

chrome-capture

## Types of changes


m0ksem (approved):
Thanks!

PR #1348, opened 2021-12-14
Fix #1308 Fix VaInput textarea height and resize

### LINKED ISSUE #1308: Textarea's row limits don't work without an autogrow prop
**Vuestic-ui version:** 1.3.1

**Steps to reproduce**
Screenshot 2021-12-06 at 14 30 03
Screenshot 2021-12-06 at 14 30 14

Limits for rows without an autogrow prop.

RVitaly1978:
## Description
close: #1308 

Changes:
- [x] - fix textarea height
- [x] - update resize logic

chrome-capture

## Types of changes


packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
```diff
@@ -46,18 +48,22 @@ export default defineComponent({
...
+        rowHeight.value = calculateRowHeight()
+      }
     }
 
     const updateHeight = () => {
-      if (!props.autosize || !textarea.value) { return }
-
-      height.value = textarea.value?.scrollHeight || 0
+      if (isResizable.value) {
+        height.value = calculateHeight() || 0
```
m0ksem, line 65: UnTG7PDWeP

Here is a bug in height calculation.

m0ksem (changes requested):

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
```diff
@@ -82,10 +88,11 @@ export default defineComponent({
 
     return {
       textarea,
-      computedHeight: computed(() => ({
+      computedStyle: computed(() => ({
         minHeight: rowHeight.value * props.minRows + 'px',
-        maxHeight: rowHeight.value * props.maxRows + 'px',
+        maxHeight: props.maxRows && (rowHeight.value * props.maxRows + 'px'),
```
Derranion, line 93: We can move logic away from `return`:

```
return {
  ...
  computedStyle,
  ...
}
```

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
```diff
@@ -82,10 +88,11 @@ export default defineComponent({
...
         height: height.value + 'px',
+        resize: props.autosize && 'none',
```
Derranion, line 95: Would be great if you can add prop description to [our docs](https://vuestic.dev/en/ui-elements/input#api) as well.

Derranion (approved):

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
RVitaly1978: fixed it

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
RVitaly1978: Probably we have some problems with our doc app, because in translations we have this description:
Screenshot 2021-12-28 at 14 23 03

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
RVitaly1978: fixed it

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
m0ksem: @RVitaly1978, try to add `name: 'VaInput'` to component options. That should make a trick. 

Vue class component auto generates component name from its class. Define component not able to make that, so we need to provide component name manually. Would be great to have component name as ESLint rule.

m0ksem (approved):

packages/ui/src/components/va-input/components/VaTextarea/VaTextarea.vue
RVitaly1978: fixed it in `va-input` and its inner components

PR #1349, opened 2021-12-15
Fix useValidation hook

RVitaly1978:
## Description
Changes:
- [x] - fix `validate` method in `useValidation` hook
- [x] - use `useValidation` hook in `VaSelect`
- [x] - update docs examples

chrome-capture

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.vue
+++ b/packages/ui/src/components/va-select/VaSelect.vue
@@ -139,7 +139,7 @@
 import { defineComponent, PropType, ref, computed, watch, nextTick } from 'vue'
 
 import { useSelectableList, useSelectableListProps } from '../../composables/useSelectableList'
-import { useFormComponent, useFormComponentProps } from '../../composables/useFormComponent'
+import { useValidation, useValidationProps, useValidationEmits } from '../../composables/useValidation'
 import { useLoadingProps } from '../../composables/useLoading'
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -239,11 +239,11 @@ export default defineComponent({
...
 
     const {
-      validate,
       isFocused,
-      computedErrorMessages,
+      validate,
       computedError,
-    } = useFormComponent(props, () => reset(), () => focus(), () => blur())
+      computedErrorMessages,
+    } = useValidation(props, context.emit, () => reset(), () => focus(), () => blur())
```
m0ksem, line 246: Here we can make:
```ts
setup (props, { emit, expose }) {
  // ...
}
```

```suggestion
    } = useValidation(props, emit, () => reset(), () => focus(), () => blur())
```

(You can ignore that comment)

m0ksem (approved):
Cool. Thanks of a lot of quality fixes.

Would be nice to see error messages from VaForm in a demo that you made, but it is a different task and not important.

packages/ui/src/components/va-select/VaSelect.vue
RVitaly1978: Fixed it

Issue #1360, opened 2021-12-27
VaSelect with error - add d-flex for parent block


zvenigorodskaia:
**Vuestic-ui version:** 1.3.3

**What is the expected behavior?**
![image](https://user-images.githubusercontent.com/35342240/147472561-1def61aa-ab06-4d4f-968b-a985c763670d.png)

**What is the current behavior?**
![image](https://user-images.githubusercontent.com/35342240/147472530-aadb664d-4766-4922-87cb-5324af9d8360.png)

Issue #1354, opened 2021-12-20
va-card__inner don't have height 100%


m0ksem:
**Vuestic-ui version:** 1.3.3

Setting 

```css
.va-card {
  height: 100%;
}
```

don't work

Issue #1353, opened 2021-12-19
VaSelect inner label can be selected by user


m0ksem:
**Vuestic-ui version:** 1.3.3

For now, it is unusable if user want to copy this text and looks ugly after click.
![Q7YznemN3b](https://user-images.githubusercontent.com/23530004/146662355-952580af-151d-446f-a81d-76b5a64e45f8.gif)

zvenigorodskaia:
@m0ksem You think we need to disable text highlighting here?

m0ksem:
Yap. Would be nice to remove text highlighting on double click if it is possible. Or remove text highlighting completely.

PR #1363, opened 2021-12-27
VaSelect and VaCard ui bugfixes

zvenigorodskaia:
close: #1353 close: #1360 close: #1354
 
## Description
 - VaSelect inner label - disable text highlighting
 - VaSelect with error - add d-flex for parent block
 - Added height for va-card__inner

## Types of changes


packages/ui/src/components/va-switch/VaSwitch.vue
```diff
@@ -332,6 +332,11 @@ export default class VaSwitch extends mixins(
   #{&}__track-label {
     color: $white;
     margin: auto 0.5rem auto 2rem;
+    user-select: none;
+    -webkit-user-select: none;
```
asvae, line 336: Pretty sure prefixes are auto-added (no need to provide them in css, `user-select: none`, should be enough).

packages/docs/src/page-configs/ui-elements/switch/examples/Error.vue
```diff
@@ -1,5 +1,5 @@
 <template>
```
asvae, line 1: yarn.lock doesn't seem to need any change.
So keeping it the way it was would make sense from my perspective.

asvae:
Looks good. Thanks!

Issue #1347, opened 2021-12-13
Infinite scroll component calls load callback no matter how far from the container's end


smheidrich:
**Vuestic-ui version:** 1.3.3

**Steps to reproduce**

Just scroll up and down for a bit in the `va-infinite-scroll` examples in the docs: https://vuestic.dev/en/ui-elements/infinite-scroll

**What is the expected behavior?**

From what I understand, the `load` callback should normally only be called if the current scroll position is within `offset` pixels from the end of the container. The default is 500 which should be exceeded after scrolling up and down in the examples a bit.

**What is the current behavior?**

Scrolling up or down anywhere within the containers causes the `load` callback to be called and more elements to be loaded, regardless of how close to the bottom of the container we are.

**Other information**

Happens in both Firefox and Chrome on my system.

PR #1369, opened 2021-12-28
Fix#1347 Fix VaInfinite scroll

RVitaly1978:
## Description
close: #1347

Changes:
- [x] - rewrite component on Composition API
- [x] - add events: `onload` and `onerror`
- [x] - update examples

chrome-capture

chrome-capture (1)

## Types of changes


packages/ui/src/components/va-infinite-scroll/hooks/useScroll.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (props.reverse) {
+      element.value.scrollTop = element.value.scrollHeight
+    }
+
+    addListener()
+  })
+
+  onBeforeUnmount(() => {
+    if (props.disabled) { return }
+
```
m0ksem, line 43:
```suggestion
```

I think there is no need for this check.

packages/ui/src/components/va-infinite-scroll/hooks/useScroll.ts
```diff
@@ -0,0 +1,51 @@
...
+    removeListener()
+  })
+
+  return {
+    addListener,
+    removeListener,
```
m0ksem, line 49: Please, mark it as Scroll Listener.

```suggestion
    addScrollListener,
    removeScrollListener,
```

packages/docs/src/page-configs/ui-elements/infinite-scroll/examples/Default.vue
```diff
@@ -6,6 +6,7 @@
     >
       <va-infinite-scroll
         :load="appendRecordsAsync"
+        :offset="40"
```
m0ksem, line 9: Why do we need to offset in default demo? I think the default demo should work without additional props.

As I user, I want components "just work".

```suggestion
```

packages/docs/src/page-configs/ui-elements/infinite-scroll/examples/Reverse.vue
```diff
@@ -6,6 +6,7 @@
     >
       <va-infinite-scroll
         :load="appendRecordsAsync"
+        :offset="40"
```
m0ksem, line 9: Let's better have additional demo with offset.

```suggestion
```

packages/ui/src/components/va-infinite-scroll/VaInfiniteScroll.vue
```diff
@@ -1,30 +1,26 @@
 <template>
   <component
-    :is="$props.tag"
+    :is="tag"
```
m0ksem, line 3:
```suggestion
    :is="$props.tag"
```

$props can help us separate props, computed and data properties in template. Do you see any reasons to remove this $props (Maybe some issues with types or smth like this)?

m0ksem (changes requested):

packages/ui/src/components/va-infinite-scroll/hooks/useScroll.ts
RVitaly1978: fixed it

packages/ui/src/components/va-infinite-scroll/hooks/useScroll.ts
RVitaly1978: fixed it

packages/ui/src/components/va-infinite-scroll/VaInfiniteScroll.vue
RVitaly1978: returned everything back ))

packages/docs/src/page-configs/ui-elements/infinite-scroll/examples/Default.vue
RVitaly1978: returned everything back ))

packages/docs/src/page-configs/ui-elements/infinite-scroll/examples/Reverse.vue
RVitaly1978: returned everything back ))

m0ksem (approved):

packages/ui/src/components/va-infinite-scroll/VaInfiniteScroll.vue
Derranion: Just wanted to add that IDE highlights props \ data \ computed with a different color but yeah adding `props` may make it easier to review.

Issue #1355, opened 2021-12-23
Vuestic input component - @change and @input events do not work


alfred-low:
**Vuestic-ui version:** 1.3.3

**Steps to reproduce**
Implement via:
<va-input
  type="textarea"
  autosize
  :label="narrative.title"
  v-model="narratives[narrative.title.toLowerCase()]"
  :min-rows="2"
  @change="stageIsComplete"
>
</va-input>

**What is the expected behavior?**
stageIsComplete method is called when there is user input into the text field.

**What is the current behavior?**
stageIsComplete method is NOT called.

**Other information**
As far as I can determine, despite documentation (https://vuestic.dev/en/ui-elements/input), the Vuestic input component does not appear to support the change or input events.  The only event that seems to work is @click.

PR #1379, opened 2021-12-30
Feat/fix vainput behavior

VasylHryha:
## Description

close:  https://github.com/epicmaxco/vuestic-ui/issues/1355

</details>
Changes:

- [x] Fix VaInput documentation and update it
- [x] Fix issue with overriding between validation onFocus, onBlur and EmitProxy onFocus, onBlur
- [x] Attach inputEvents to textArea

## Types of changes


packages/ui/src/composables/useEmitProxy.ts
```diff
@@ -11,7 +11,7 @@ export const useEmitProxy = <T extends string>(events: T[]) => {
   const createListeners = (emit: (event: T, ...args: any[]) => void) => {
     return events.reduce((acc, key) => ({
       ...acc, [eventToListenerName(key)]: (...args: any[]) => emit(key, ...args),
-    }), {} as Record<T, any>)
+    }), {} as Record<string, any>)
```
Derranion, line 14: Looks like Maksim will not like this :D

Derranion (approved):
Great.

Just a side note - we usually using rebase, if possible, instead of merge in order to have readable history.

packages/ui/src/composables/useEmitProxy.ts
VasylHryha: So actually Record can't contain  `T` type since here 
```javascript
[eventToListenerName(key)]
```
it changes, example: 
 ```javascript
const event = 'focus'
cons result = eventToListenerName(event)
console.log(result) // onFocus
```
As we could see, output values are not the same as input values

sergey-kravcov (approved):

asvae:
Great work. Thanks!

PR #1408, opened 2022-01-12
Rewrite VaOptionList.vue in composition api

### LINKED ISSUE #1377: Move OptionList to Composition API
This will be a bit easier to do because we already have `useFormComponent` ([link](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/composables/useFormComponent.ts)) that can replace `FormComponentMixin` that is used as the main part of `SelectableListMixin` inside the `OptionList`.

As a **complete example** you could check our [VaSelect](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-select/VaSelect.vue) that uses most of the Composition API features.

LighthouseKeeperYN:
## Description

close: #1377
## Types of changes

### DIFF
--- a/packages/ui/src/components/va-option-list/VaOptionList.vue
+++ b/packages/ui/src/components/va-option-list/VaOptionList.vue
@@ -2,50 +2,50 @@
   <VaMessageListWrapper
     :error="computedError"
     :error-messages="computedErrorMessages"
-    :error-count="errorCount"
+    :error-count="$props.errorCount"
   >
     <ul
       class="va-option-list__list"
-      :id="String(id)"
+      :id="String($props.id)"
     >
       <li
-        v-for="(option, index) in options"
+        v-for="(option, index) in $props.options"
         :key="getKey(option)"
       >
         <slot
           :props="{
             option,
             isDisabled,
-            name,
-            color,
-            leftLabel,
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-option-list/VaOptionList.vue
```diff
@@ -2,61 +2,61 @@
   <VaMessageListWrapper
     :error="computedError"
     :error-messages="computedErrorMessages"
-    :error-count="$props.errorCount"
+    :error-count="errorCount"
```
asvae, line 5: Discussed with @RVitaly1978,
there is no much need to remove $props, as they serve as indication you're using props directly, and not computed of sorts.

asvae:
Looks good. Thanks!

Feel free to merge after the fix :).

packages/ui/src/composables/useValidation.ts
```diff
@@ -18,7 +18,7 @@ interface ValidationProps {
 }
 
 export const useValidationProps = {
-  modelValue: { },
+  modelValue: { required: false },
```
asvae, line 21: Why do you need `required: false` here? It's false by default.

PR #1414, opened 2022-01-13
Deprecate useFormComponent

LighthouseKeeperYN:
Discussed with @RVitaly1978 and decided to deprecate useFormComponent in favor of useValidation.

## Types of changes


packages/ui/src/composables/useValidation.ts
```diff
@@ -42,7 +42,7 @@ export const useValidation = (
   emit: (event: any) => any,
   reset: () => any,
   focus: () => any,
-  blur: () => any,
+  blur?: () => any,
```
asvae, line 45: What is blur doing in this method at all?
I don't see it in `FormChild` declaration.

asvae:
Looks good!
Merge after fix :).

asvae (approved):

packages/ui/src/composables/useValidation.ts
RVitaly1978: It is my mistake. This can be safely removed after fixes in the components `VaInput`, `VaSelect`, `VaTimeInput`

RVitaly1978 (approved):

Issue #1413, opened 2022-01-13
Have no way to configure column width for component va-datatable


wossig:
**Vuestic-ui version:** 1.3.3

**Steps to reproduce**
<img width="693" alt="image" src="https://user-images.githubusercontent.com/22833218/149355996-f175f845-9298-442a-b890-67fa00207db5.png">

**What is the expected behavior?**

**What is the current behavior?**

**Other information**

PR #1422, opened 2022-01-18
Fix#1413 Add columns configuration to `VaDataTable`

RVitaly1978:
## Description
close: #1413 

changes: 
- [x] - add `width`, `classes` and `style` props to columns configuratiion
- [x] - update examples, locales and page-config

chrome-capture

## Types of changes


packages/ui/src/components/va-data-table/hooks/useStyleable.ts
```diff
@@ -2,7 +2,24 @@ import { TableCell } from './useRows'
 import { getFocusColor, getHoverColor } from '../../../services/color-config/color-functions'
 import { getColor } from '../../../services/color-config/color-config'
 import { computed, Ref } from 'vue'
-import { TableColumn } from './useColumns'
+import { TableColumn, TClassesOptions, TStyleOptions } from './useColumns'
+
+function getClasses (classes: TClassesOptions | undefined): string[] {
+  if (!classes) { return [] }
```
LighthouseKeeperYN, line 8: I think it's worth using default value instead of checking for undefined
`function getClasses (classes: TClassesOptions | undefined = []): string[]`

packages/ui/src/components/va-data-table/hooks/useStyleable.ts
```diff
@@ -2,7 +2,24 @@ import { TableCell } from './useRows'
...
+  if (typeof classes === 'function') {
+    const computedClasses = classes()
+    return (typeof computedClasses === 'string') ? [computedClasses] : computedClasses
+  }
+
+  return (typeof classes === 'string') ? [classes] : classes
+}
+
+function getStyles (styles: TStyleOptions | undefined): Record<string, any> {
+  if (!styles) { return {} }
```
LighthouseKeeperYN, line 19: Same as above

packages/ui/src/components/va-data-table/hooks/useStyleable.ts
RVitaly1978: Thanks! Fixed it.

m0ksem (approved):
Few suggestions:
- Third column width select is missing. 
- You can make width selects clearable, since they don't have initial value.

packages/docs/src/page-configs/ui-elements/data-table/examples/Styling.vue
```diff
@@ -12,6 +12,12 @@
         label="First heading vertical align"
         :options="verticalAlignOptions"
       />
+      <va-select
+        class="mt-2"
+        v-model="columns[0].width"
+        label="First heading width"
+        :options="widthOptions"
```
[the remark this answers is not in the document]
RVitaly1978: Added width select to the third column and made all selects clearable.

Issue #1407, opened 2022-01-12
Datatables row edit


arnaudnjita:
**Is your feature request related to a problem? Please describe.**

Yes because I am trying to devise a way of editing the data of specific rows in the datatables section of the documentation.

**Describe the solution you'd like**

I'll love that feature of editing data in the table rows of the datatables should be included in the documentation. 

**Describe alternatives you've considered**

I am trying to see how I can extract the id of that specific row in the datatable, with the id I can work a way around in using it to edit the information

asvae:
We surely can add an example for that.

asvae:
Something like [that](https://vuetifyjs.com/en/components/data-tables/#external-sorting):

![image](https://user-images.githubusercontent.com/5394573/149319188-936f00dd-869c-4b97-9541-adba4fb44d11.png)

PR #1428, opened 2022-01-20
Fix#1407 Add example of `va-data-table` with CRUD

RVitaly1978:
## Description
close: #1407

changes:
- [x] - minor update component (default `verticalAlign`, expand data to `cell(column)` slot)
- [x] - add example with **CRUD** in docs

chrome-capture
chrome-capture (1)
chrome-capture (2)

## Types of changes


LighthouseKeeperYN (approved):

packages/docs/src/page-configs/ui-elements/data-table/examples/CRUD.vue
```diff
@@ -0,0 +1,144 @@
...
+      { key: 'username', sortable: true },
+      { key: 'email', sortable: true },
+      { key: 'actions', width: 130 },
+    ]
+
+    return {
+      items,
+      columns,
+
+      addNameModel: '',
```
asvae, line 70: FIrst off - I would suggest using an object for create and edit.
I.e. 
```
editedUser: null
editedUser: {id: '', name: '', surname: '', email: ''}
```
Same for created user. So resets updates would look differently because of different data structure.

packages/docs/src/page-configs/ui-elements/data-table/examples/CRUD.vue
```diff
@@ -0,0 +1,144 @@
...
+    },
+    editItem () {
+      this.items = [
+        ...this.items.slice(0, this.editItemId),
+        this.updateItem(),
+        ...this.items.slice(this.editItemId + 1),
+      ]
+      this.resetEditData()
+    },
+    openModalToEditItemById (id) {
```
asvae, line 122: Normally for edit you would just clone `{...likeThis}` an object.

asvae:
Found a couple of things, and maybe we can discuss it briefly is something doesn't look clear. 😉

packages/docs/src/page-configs/ui-elements/data-table/examples/CRUD.vue
RVitaly1978: Thanks! Fixed it.

packages/docs/src/page-configs/ui-elements/data-table/examples/CRUD.vue
RVitaly1978: Thanks! Fixed it.

Issue #1397, opened 2022-01-06
Create a list of components we need to convert to composition API


asvae:
closes #567 after all components refactored.

LighthouseKeeperYN:
- [ ] packages\docs\src\components\DocsAlert.vue
- [ ] packages\docs\src\components\DocsAnchor.vue
- [ ] packages\docs\src\components\DocsApi\ApiDocs.vue
- [ ] packages\docs\src\components\DocsApi\ApiDocsPropsRow.vue
- [ ] packages\docs\src\components\DocsCode.vue
- [ ] packages\docs\src\components\DocsComponent.vue
- [ ] packages\docs\src\components\DocsContent.vue
- [ ] packages\docs\src\components\DocsExample.vue
- [ ] packages\docs\src\components\DocsHeadline.vue
- [ ] packages\docs\src\components\DocsLink.vue
- [ ] packages\docs\src\components\DocsNavigation.vue
- [ ] packages\docs\src\components\DocsSubtitle.vue
- [ ] packages\docs\src\components\DocsTable\DocsTable.vue
- [ ] packages\docs\src\components\header\components\ColorDropdown.vue
- [ ] packages\docs\src\components\header\components\HeaderSelector.vue
- [ ] packages\docs\src\components\header\components\LanguageDropdown.vue
- [ ] packages\docs\src\components\header\components\VersionDropdown.vue
- [ ] packages\docs\src\components\header\components\VuesticDocsLogo.vue
- [ ] packages\docs\src\components\header\components\VuesticLogo.vue
- [ ] packages\docs\src\components\header\Header.vue
- [ ] packages\docs\src\components\landing\Admin.vue
- [ ] packages\docs\src\components\landing\ColorTab.vue
- [ ] packages\docs\src\components\landing\Customize.vue
- [ ] packages\docs\src\components\landing\Footer.vue
- [ ] packages\docs\src\components\landing\Header.vue
- [ ] packages\docs\src\components\landing\icons\IconAdmin.vue
- [ ] packages\docs\src\components\landing\icons\IconEpicmax.vue
- [ ] packages\docs\src\components\landing\icons\IconSpinners.vue
- [ ] packages\docs\src\components\landing\OpenSource.vue
- [ ] packages\docs\src\components\landing\SeamlessIntegration\SeamlessIntegration.vue
- [ ] packages\docs\src\components\landing\StarsButton.vue
[...]

PR #1438, opened 2022-01-24
refactor: rewrite VaAlert in composition api

LighthouseKeeperYN:
#1397 

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-alert/VaAlert.vue
+++ b/packages/ui/src/components/va-alert/VaAlert.vue
@@ -66,106 +66,55 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, computed, PropType, ref, Ref, nextTick, onMounted, onBeforeUnmount } from 'vue'
+import { defineComponent, computed, PropType } from 'vue'
 
-import { getTextColor } from '../../services/color-config/color-functions'
-import { useColors } from '../../services/color-config/color-config'
 import { useStateful } from '../../mixins/StatefulMixin/cStatefulMixin'
-
+import { useAlertStyles } from './useAlertStyles'
 import VaIcon from '../va-icon'
 
 export default defineComponent({
   name: 'VaAlert',
   components: { VaIcon },
   props: {
-    color: ({ type: String as PropType<string>, default: 'primary' }),
-    title: ({ type: String as PropType<string>, default: '' }),
-    description: ({ type: String as PropType<string>, default: '' }),
-    icon: ({ type: String as PropType<string>, default: '' }),
-    closeText: ({ type: String as PropType<string>, default: '' }),
-    closeable: ({ type: Boolean as PropType<boolean>, default: false }),
-    dense: ({ type: Boolean as PropType<boolean>, default: false }),
-    outline: ({ type: Boolean as PropType<boolean>, default: false }),
-    center: ({ type: Boolean as PropType<boolean>, default: false }),
-    borderColor: ({ type: String as PropType<string>, default: '' }),
-    border: ({
+    modelValue: { type: Boolean as PropType<boolean>, default: true },
+    color: { type: String as PropType<string>, default: 'primary' },
+    title: { type: String as PropType<string>, default: '' },
+    description: { type: String as PropType<string>, default: '' },
+    icon: { type: String as PropType<string>, default: '' },
+    closeText: { type: String as PropType<string>, default: '' },
+    closeable: { type: Boolean as PropType<boolean>, default: false },
+    dense: { type: Boolean as PropType<boolean>, default: false },
+    outline: { type: Boolean as PropType<boolean>, default: false },
+    center: { type: Boolean as PropType<boolean>, default: false },
+    borderColor: { type: String as PropType<string>, default: '' },
+    border: {
       type: String as PropType<string>,
       default: '',
       validator: (value: string) => ['top', 'right', 'bottom', 'left', ''].includes(value),
-    }),
-    modelValue: ({
-      type: Boolean as PropType<boolean>,
-      default: true,
-    }),
+    },
   },
   setup (props, { slots, emit }) {
-    const dark = 'var(--va-dark)'
-
-    const { getColor } = useColors()
-    const colorComputed = computed(() => getColor(props.color))
-
-    const hasIcon = computed(() => props.icon || slots.icon)
-    const hasTitle = computed(() => props.title || slots.title)
-    const borderClass = computed(() => `va-alert__border--${props.border}`)
-    const closeIcon = computed(() => props.closeText || 'close')
-    const alertStyle = computed(() => {
-      let background = colorComputed.value
-      let boxShadow = 'none'
-      if (props.outline) {
-        background = 'transparent'
-      }
-      if (props.border) {
-        background = '#ffffff'
-        boxShadow = 'var(--va-alert-box-shadow)'
-      }
-      return {
-        border: props.outline && `1px solid ${colorComputed.value}`,
-        padding: props.dense && '0.25rem 0.75rem',
-        background,
-        boxShadow,
-      }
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -68,127 +66,110 @@
...
-    }
-  }
+  props: {
+    color: ({ type: String as PropType<string>, default: 'primary' }),
+    title: ({ type: String as PropType<string>, default: '' }),
+    description: ({ type: String as PropType<string>, default: '' }),
+    icon: ({ type: String as PropType<string>, default: '' }),
+    closeText: ({ type: String as PropType<string>, default: '' }),
+    closeable: ({ type: Boolean as PropType<boolean>, default: false }),
+    dense: ({ type: Boolean as PropType<boolean>, default: false }),
```
m0ksem, line 87:
```suggestion
    dense: { type: Boolean as PropType<boolean>, default: false },
```

Extra brackets.

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -201,12 +182,14 @@ export default class VaAlert extends mixins(
   margin: var(--va-alert-margin-y) auto;
   display: var(--va-alert-display);
   align-items: var(--va-alert-align-items);
-  border: var(--va-alert-border-width, var(--va-control-border)) solid transparent;
+  border:
+    var(--va-alert-border-width, var(--va-control-border)) solid
+    transparent;
```
m0ksem, line 187:
```suggestion
  border: var(--va-alert-border-width, var(--va-control-border)) solid transparent;
```

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -68,127 +66,110 @@
...
+    const dark = 'var(--va-dark)'
+
+    const { getColor } = useColors()
+    const colorComputed = computed(() => getColor(props.color))
+
+    const hasIcon = computed(() => props.icon || slots.icon)
+    const hasTitle = computed(() => props.title || slots.title)
+    const borderClass = computed(() => `va-alert__border--${props.border}`)
+    const closeIcon = computed(() => props.closeText || 'close')
+    const alertStyle = computed(() => {
```
m0ksem, line 111: Can you make more extra new lines between methods?

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -68,127 +66,110 @@
...
+        boxShadow = 'var(--va-alert-box-shadow)'
+      }
+      return {
+        border: props.outline && `1px solid ${colorComputed.value}`,
+        padding: props.dense && '0.25rem 0.75rem',
+        background,
+        boxShadow,
+      }
+    })
+    const contentStyle = computed(() => {
```
m0ksem, line 128: It would be cool to have styling logic hidden in some useAlertStyle hook… It's not a requirement, but hiding this logic will make this component even more simpler.

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -68,127 +66,110 @@
...
+    const borderStyle = computed(() => ({
+      backgroundColor: props.borderColor
+        ? getColor(props.borderColor)
+        : colorComputed.value,
+    }))
 
-  get borderClass () {
-    return `va-alert__border--${this.border}`
-  }
+    const { valueComputed } = useStateful(props, emit)
```
m0ksem, line 157: valueComputed has not been returned in setup function.

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -68,127 +66,110 @@
...
+    const contentStyle = computed(() => {
+      let color = getTextColor(colorComputed.value)
+      if (props.outline) {
+        color = colorComputed.value
+      }
+      if (props.border) {
+        color = dark
+      }
+      return {
+        alignItems: props.center && 'center',
```
m0ksem, line 137: Would be nice to add here better typing. `contentStyle` should return something assignable to `CSSProperties` type so Volar wouldn't mark error. 
image

m0ksem (changes requested):
Great. Review my suggestion and fix `valueComputed`.

packages/ui/src/components/va-alert/VaAlert.vue
LighthouseKeeperYN: Fixed

packages/ui/src/components/va-alert/VaAlert.vue
LighthouseKeeperYN: Done

packages/ui/src/components/va-alert/VaAlert.vue
LighthouseKeeperYN: Fixed

packages/ui/src/components/va-alert/VaAlert.vue
LighthouseKeeperYN: Fixed

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -2,6 +2,7 @@
   <transition
     v-if="valueComputed"
     name="fade"
+    ref="alert"
```
m0ksem, line 5: Why do we need this?

LighthouseKeeperYN: It's a a rudiment from debugging, I'll remove it.

packages/ui/src/components/va-alert/useAlertStyles.ts
```diff
@@ -0,0 +1,92 @@
...
+  const alertStyle = computed(() => {
+    let background = colorComputed.value
+    let boxShadow = 'none'
+
+    if (props.outline) {
+      background = 'transparent'
+    }
+
+    if (props.border) {
+      background = '#ffffff'
```
RVitaly1978, line 37: I think, it's better to use css-variables (like `const dark = 'var(--va-dark)'`) instead of constants.
And below (where `padding`) is the same story.

RVitaly1978 (approved):

packages/ui/src/components/va-alert/useAlertStyles.ts
LighthouseKeeperYN: done

packages/ui/src/components/va-alert/useAlertStyles.ts
```diff
@@ -20,6 +20,8 @@ type AlertStyleProps = {
 
 export const useAlertStyles = (props: AlertStyleProps) => {
   const dark = 'var(--va-dark)'
+  const white = '#ffffff'
```
RVitaly1978, line 23: I mean, that we have css-variable `--va-white`, so
`const white = 'var(--va-white)'`

packages/ui/src/components/va-alert/useAlertStyles.ts
```diff
@@ -20,6 +20,8 @@ type AlertStyleProps = {
...
+  const paddingDense = '0.25rem 0.75rem'
```
RVitaly1978, line 24: And here we can add local css-variable in `_variables.scss`.
something like this: `--va-affix-padding-dense: '0.25rem 0.75rem';`

packages/ui/src/components/va-alert/useAlertStyles.ts
LighthouseKeeperYN: done

packages/ui/src/components/va-alert/useAlertStyles.ts
LighthouseKeeperYN: done

PR #1452, opened 2022-01-25
refactor: rewrite avatar to composition api

LighthouseKeeperYN:
#1397

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-avatar/VaAvatar.vue
+++ b/packages/ui/src/components/va-avatar/VaAvatar.vue
@@ -1,15 +1,14 @@
 <template>
   <div
     class="va-avatar"
-    ref="avatar"
     :style="computedStyle"
   >
     <slot>
       <va-progress-circle
-        v-if="loading"
-        indeterminate
+        v-if="$props.loading"
         :size="sizeComputed"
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -24,47 +23,51 @@
...
+    fontSize: { type: String as PropType<string>, default: '' },
+  },
+  setup (props) {
+    const { getColor } = useColors()
+    const { sizeComputed, fontSizeComputed } = useSize(props, 'VaAvatar')
+
+    const colorComputed = computed(() => getColor(props.color))
+
+    const computedStyle = computed(() => ({
+      color: getColor(props.textColor, '#ffffff'),
```
m0ksem, line 55:
```suggestion
      color: getColor(props.textColor, 'white'),
```

Use white from colors config. For dark theme we can set white to `#000000` and VaAvatar should depend on this color.

m0ksem (changes requested):
Everything else look good, except white color.

RVitaly1978 (approved):

PR #1455, opened 2022-01-26
refactor: rewrite backtop to composition api

LighthouseKeeperYN:
#1397

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-backtop/VaBacktop.vue
+++ b/packages/ui/src/components/va-backtop/VaBacktop.vue
@@ -15,96 +15,89 @@
 </template>
 
 <script lang="ts">
-import { Options, mixins, prop, Vue } from 'vue-class-component'
+import { defineComponent, computed, PropType, ref, onMounted, onBeforeMount, Ref } from 'vue'
 
 import VaButton from '../va-button'
 
-class Props {
-  target = prop<Element | string | undefined>({
-    type: [Object, String],
-    default: () => window,
-  })
-
-  visibilityHeight = prop<number>({ type: Number, default: 300 })
-  speed = prop<number>({ type: Number, default: 50 })
-  verticalOffset = prop<string>({ type: String, default: '1rem' })
-  horizontalOffset = prop<string>({ type: String, default: '1rem' })
-  color = prop<string>({ type: String, default: '' })
-  horizontalPosition = prop<string>({
-    type: String,
-    default: 'right',
-    validator: (value: string) => {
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/va-backtop/VaBacktop.vue
```diff
@@ -15,96 +15,89 @@
...
-      [this.horizontalPosition]: this.horizontalOffset,
+    const computedStyle = computed(() => ({
+      [props.verticalPosition]: props.verticalOffset,
+      [props.horizontalPosition]: props.horizontalOffset,
+    }))
+
+    const targetElement: Ref<Element> = ref(document.body)
+    const getTargetElement = () => typeof props.target === 'string'
+      ? document.querySelector(props.target) || document.body
+      : props.target
```
RVitaly1978, line 59: Maybe we can do this:
`const targetElement = computed<Element>(() => typeof props.target === 'string' ? document.querySelector(props.target) || document.body : props.target)`
And maybe instead of `document.body` we have to use **parent of component**?

RVitaly1978 (changes requested):

packages/ui/src/components/va-backtop/VaBacktop.vue
LighthouseKeeperYN: > Maybe we can do this:
That's what I tried initially but the target is not always mounted when the computed is running which results in target being null.
>And maybe instead of document.body we have to use parent of component?
I don't mind but the docs says default target is body.

RVitaly1978 (approved):

packages/ui/src/components/va-backtop/VaBacktop.vue
```diff
@@ -15,96 +15,89 @@
...
-  get computedStyle () {
-    return {
-      [this.verticalPosition]: this.verticalOffset,
-      [this.horizontalPosition]: this.horizontalOffset,
+    const computedStyle = computed(() => ({
+      [props.verticalPosition]: props.verticalOffset,
+      [props.horizontalPosition]: props.horizontalOffset,
+    }))
+
+    const targetElement: Ref<Element> = ref(document.body)
```
asvae, line 56: This looks very wrong to me.

Pretty sure
```
ref(document.body)
```
will make the whole document into reactive data structure. Which could be CPU intensive and also not particularly useful.

LighthouseKeeperYN: fixed

Issue #1436, opened 2022-01-23
It should have a way to  use toast component in specific method, service easier under Typescript


wossig:
**Vuestic-ui version:** 1.3.3

**Steps to reproduce**
I'm trying to use taost component after some async operation (request api) did success, but from the document, it seems have no simple way to simply use global variable `$vaToast` under typescript, i checked existing issue regarding toast component, there's no regarding's.

**What is the expected behavior? **
Add global variable which other Vue ui framwork did like 
[https://element-plus.org/en-US/component/notification.html#with-types](url)
```javascript
import { ElNotification } from 'element-plus'

ElNotification({
  title: 'Success',
  message: 'This is a success message',
  type: 'success',
})
```

**What is the current behavior?**

Currently i used toast like this in my code, but not good style for typescript:
```javascript
function toast(config: any) {
      if (app) {
        const vaToast = app.appContext.config.globalProperties.$vaToast
        vaToast.init(config)
      }
    }
```

**Other information**

asvae:
@m0ksem can you elaborate on this? Not sure if it's up for implementation.

PR #1464, opened 2022-01-27
Feat/toast composition hook

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1436

Now `VaToast` is independent of Vue app context and can be used in any method.

Added `useToast` hook and test demo.

## Example
```ts
const { init, close, closeAllCreatedInThisHook, closeAll } = useToast()
```

## Breaking Changes

`closeAllCreatedInThisHook` - used if user want to close all toasts that actually created in SetupContext. So if multiple toasts shown from different components (I mean setup context) it would close toasts from other components. Can be useful. Can you help with method name, or you find it ok?

`closeAll` by default closes toasts only created in app, where hook is used. It accepts boolean, if it true it will close toasts for any vue app created on page. I'm not sure about default behavior, if you find it good - let's keep it. 

## Types of changes


RVitaly1978 (approved):

packages/ui/src/components/va-toast/VaToast.demo.vue
```diff
@@ -34,22 +34,63 @@
...
     const { setGlobalConfig } = useGlobalConfig()
-
     setGlobalConfig(config => ({
       ...config,
       VaToast: {
         color: 'red',
       },
     }))
+
+    const { init, close, closeAllCreatedInThisHook, closeAll } = useToast()
```
LighthouseKeeperYN, line 57: I think we should either change variable names or never destructure toast object. Standalone "init" function looks too vague.

packages/ui/src/components/va-toast/VaToast.demo.vue
```diff
@@ -34,22 +34,63 @@
...
+
+          init({ message: 'Now we try to close all toasts in 3 seconds' })
+          init({ message: 'Now we try to close all toasts in 3 seconds' })
+          init({ message: 'Now we try to close all toasts in 3 seconds' })
+          init({ message: 'Now we try to close all toasts in 3 seconds' })
+          init({ message: 'Now we try to close all toasts in 3 seconds' })
+
+          setTimeout(() => {
+            closeAll()
+          }, 3000)
```
LighthouseKeeperYN, line 86:
```suggestion
          setTimeout(closeAll, 3000)
```

packages/ui/src/components/va-toast/useToast.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Closes all VaToast instances in current App instance.
+   * @param allApps if you have multiple vue apps on page, set allApps to true to close it for all apps.
+   */
+  const closeAll = (allApps = false) => closeAllNotifications(allApps ? undefined : appContext)
+
+  /** Use this method if you need to close only toasts that created with this hook */
+  const closeAllCreatedInThisHook = () => {
+    createdInThisSetupContext.forEach((id) => closeById(id))
```
LighthouseKeeperYN, line 31:
```suggestion
    createdInThisSetupContext.forEach(closeById)
```

LighthouseKeeperYN (changes requested):

packages/ui/src/components/va-toast/VaToast.demo.vue
m0ksem: User can change it's name
```ts
const { init: initToast, close, closeAllCreatedInThisHook, closeAll } = useToast()
```
or simply

```ts
const toast = useToast()
```


I just left it as it was in options API `$vaToast.init()`

packages/ui/src/components/va-toast/useToast.ts
m0ksem: Actually, passing id argument is much readable. Also, if we will add some additional argument to closeById method, this forEach can cause some bugs. I would prefer to use a new arrow function as callback.

LighthouseKeeperYN (approved):

asvae:
Thanks!
Probably needs some docs as well, so that users will be able to use it.

asvae:
> if it true it will close toasts for any vue app created on page
Interesting approach. Could be useful is someone attempts [micro frontends](https://micro-frontends.org/). We might want to enforce this in architecture notes.

m0ksem:
My bad. I've forgotten about docs.

m0ksem:
I added some docs here.
image
image

PR #1472, opened 2022-02-01
Fix#1278 Fix input docs

### LINKED ISSUE #1278: Missed translations in va-input docs api
chrome-capture

Required:
- [x] - add trsnslations
- [ ] - add example with attrs

RVitaly1978:
## Description
close #1278

changes:
- [x] add missed descriptions
- [x] add examples with `attrs` to docs and demo
- [x] remove redundant `expose` in **setup** fuction
- [x] update `va-form` validation

## Types of changes


packages/ui/src/components/va-form/VaForm.vue
```diff
@@ -79,9 +79,13 @@ export default defineComponent({
...
     const validate = () => { // NOTE: temporarily synchronous validation
-      const formValid = nestedFormElements.value
+      let formValid = true
+      nestedFormElements.value
         .filter(({ validate }) => validate)
-        .every((child) => child.validate())
+        .forEach((child) => {
+          const isValidChild = child.validate()
+          if (!isValidChild && formValid) { formValid = false }
+        })
```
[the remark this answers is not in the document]
RVitaly1978: In the case with `every`, validation in the form stops after the first `invalid` child.

LighthouseKeeperYN (approved):

packages/ui/src/components/va-form/VaForm.vue
```diff
@@ -78,10 +78,12 @@ export default defineComponent({
       }
     }
 
+    // validation for every nested child
     const validate = () => { // NOTE: temporarily synchronous validation
       const formValid = nestedFormElements.value
         .filter(({ validate }) => validate)
-        .every((child) => child.validate())
+        .map((child) => child.validate()) // more readable than with 'forEach'
```
m0ksem, line 85: It's not forEach. It runs child.validate until any of result is false.

RVitaly1978: We want to validate each child element to show the user their validation status. So we use `map` and `every` method (for the human-readable reason we use these instead of `forEach`)

PR #1473, opened 2022-02-01
Refactor/checkbox

LighthouseKeeperYN:
#1397

Thoughts on value type changes are deeply appreciated.

## Types of changes


packages/ui/src/composables/useSelectable.ts
```diff
@@ -0,0 +1,143 @@
+import { PropType, computed, SetupContext, Ref } from 'vue'
```
m0ksem, line 1: How about to move this hook to checkbox if it used only in VaCheckbox? See TimePicker.

m0ksem (approved):

RVitaly1978 (approved):

Issue #1184, opened 2021-10-30
Refused to apply style from <URL>~normalize.css


amar-ox:
I got this error with normalize.css after updating to vuestic-ui 1.3.0
And va-date/time input are not rendered correctly

<img width="305" alt="vuestic-issue" src="https://user-images.githubusercontent.com/24830847/139522666-e00efc70-a75e-4958-980e-8c202c533853.PNG">

With Vue3 and @vue/cli 5.0.0-beta.4

Current package-lock.json: 
"vuestic-ui": {
      "version": "1.3.0",
      "resolved": "https://registry.npmjs.org/vuestic-ui/-/vuestic-ui-1.3.0.tgz",
      "integrity": "sha512-FMwt9vZsmxhuVBiXuOSViFWwDzUBmbzvoSSldwyvmlktq9ig/beHIAIWSVGZ5cDFLI7bNrdK7c5DSR4cPGhK7Q==",
      "requires": {
        "@popperjs/core": "^2.9.2",
        "asva-executors": "^0.1.26",
        "cleave.js": "^1.6.0",
        "colortranslator": "^1.7.1",
        "detect-browser": "^5.2.0",
        "element-resize-detector": "^1.2.1",
        "flatpickr": "4.6.9",
        "lodash-es": "^4.17.21",
        "normalize.css": "^8.0.1",
        "tslib": "^2.2.0",
        "vue-flatpickr-component": "^9.0.3"
      }
    }

m0ksem:
Hi, @amar-ox. 

Add following code to your vite config to resolve this issue. 
```js
  resolve: {
    alias: [
      { find: '~normalize.css', replacement: 'normalize.css'}
    ]
  }
```

amar-ox:
@m0ksem  this does not seem to work. Is this the right way to add the code?

`module.exports = {
  lintOnSave: true,
  configureWebpack: {
    resolve: {
      alias: [
        { find: '~normalize.css', replacement: 'normalize.css'}
      ]
    },  
    resolve: {
      alias: {
        'vue-i18n': 'vue-i18n/dist/vue-i18n.cjs.js',
      },
    },
  },`

m0ksem:
```js
  configureWebpack: {
    resolve: {
      alias: {
        'vue-i18n': 'vue-i18n/dist/vue-i18n.cjs.js',
        '~normalize.css': 'normalize.css'
      },
    },
  },
```

[...]

PR #1483, opened 2022-02-04
Feat/vue cli plugin vite

m0ksem:
- Added vite template generator.
- Added ag-grid prompt.
- Moved vue cli installation on top of docs.

This allows users to use `vue add vuestic-ui` to vite project.

Plugin fixes issue with normalize.css and adds required `~` alias to vite config. Related to this issue: https://github.com/epicmaxco/vuestic-ui/issues/1184. Because we have ~ alias for webpack, we need to add this alias to vite project.

I moved vue cli installation on top of docs, so it's now recommended way to install vuestic-ui.

image


m0ksem:
In this PR I added possibility to look on user's package.json. We can also make the same feature for nuxt3.

RVitaly1978 (approved):

Issue #1487, opened 2022-02-07
Doesn't work with nuxt 3


asvae:
See [Issue in nuxt 3 repo](https://github.com/nuxt/framework/discussions/1869).

We want to focus on making it work in context of nuxt 3, not getting all components functional (which we can do in context of #650)

m0ksem:
Removing `vue-property-decorator` should make a trick. https://github.com/epicmaxco/vuestic-ui/issues/1397

```ts
@Ref() readonly htmlInput!: HTMLInputElement
```

compiles to 

```ts
n.__decorate([Ref("htmlInput"),n.__metadata("design:type",HTMLInputElement)]
```

I have tried to change type to any and the issue is gone. But, there are a lot of `@Ref` usages

m0ksem:
closed in https://github.com/epicmaxco/vuestic-ui/pull/1501

PR #1501, opened 2022-02-11
Feat nuxt3

m0ksem:
Lots of changes here 😁

### List of changes
- Use mjs file format in esm build. (Required for nuxt3)
- Remove `@popperjs`, `tslib` and `asva-executors` dependecies. (There were some issues with esm builds)
- Create `usePopover`, `useDomRect`, `useDebounce` and `useDebounceFn` hooks.
`useDomRect` can also replace `element-resize-detector` dependency. But, because it used in VaPopup, we can remove this dependency along with VaPopup component (it's deprecated) 

- Refactored `VaDropdown`, `VaPopover` components to composition API.
- Changed `HTMLElement`, `Element` and `HTMLInputElement` types to `any` in decorators. (https://github.com/epicmaxco/vuestic-ui/issues/1487#issuecomment-1031349406)
- Created `createVuesticNuxtPlugin` and `createVuesticNuxtPluginWithoutComponents` (for treeshaking) plugins.
I don't know if we want to have nuxt3 as peer dependency, so for now I hard-coded `NuxtPlugin` types.
- Refactored vuestic-plugins, so I can understand why do we need them and where to find them.
Now, each plugin is placed in `[componentName|serviceName]/plugin` folder.
- Removed `installPlatform` (this was used in VaPopup)

## Things to discuss
- Do we need `VaPopup`? How about to remove it?
- Can we use nuxt as peer dependency? Likely not, there was issue with webpack types... I suggest we just let it hard-coded.
- Tests? Need to add nuxt to `bundlers-tests` packages.
- QA for `usePopover`. I don't use `popperjs` because of issues on nuxt build (`m.default` is undefiend ?😁). `floating-ui` (new `popperjs`'s name) uses async function to work with `react-native`. For me, async function causes freezes. So I made `usePopover ` hook based on popperjs code.
- Should we rename `usePopover` to `useDropdown`🤔 ?
- What about nuxt-bridge? Should we care for now?

## Things to do
- Docs about nuxt plugin.
- Remake docs to nuxt3 (I think this should be done before release, but this is quite hard to do)


packages/ui/src/composables/usePopover.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (autoPlacement) {
+      placement = getAutoPlacement(placement, anchorDomRect.value, contentDomRect.value, offset || 0)
+    }
+
+    let coords = calculateCoords(placement, anchorDomRect.value, contentDomRect.value)
+
+    if (offset) {
+      const offsetCoords = calculateOffsetCords(placement, offset)
+
+      coords = mapObject(coords, (coord, key) => coord + offsetCoords[key as 'top' | 'left' | 'right' | 'bottom'])
```
LighthouseKeeperYN, line 205:
```suggestion
      coords = mapObject(coords, (coord, key) => coord + offsetCoords[key as PlacementPosition])
```

packages/ui/src/composables/usePopover.ts
```diff
@@ -0,0 +1,196 @@
...
+      x: anchor.right,
+    }
+  }
+
+  // if position === 'bottom'
+  return {
+    x: calculateContentAlignment(align, anchor.left, anchor.width, content.width),
+    y: anchor.bottom,
+  }
+}
```
LighthouseKeeperYN, line 59:
```suggestion
const calculateContentCoords = (placement: Placement, anchor: DOMRect, content: DOMRect) => {
  const { position, align } = parsePlacement(placement)

  const alignmentX = calculateContentAlignment(align, anchor.left, anchor.width, content.width)
  const alignmentY = calculateContentAlignment(align, anchor.left, anchor.width, content.width)

    const positions = {
    top: { x: alignmentX, y: anchor.top - content.height },
    bottom: { x: alignmentX, y: anchor.bottom },
    left: { y: alignmentY, x: anchor.left - content.width },
    right: { y: alignmentY, x: anchor.right },
  }

  return positions[position]
}
```

packages/ui/src/composables/usePopover.ts
```diff
@@ -0,0 +1,196 @@
...
+  if (position === 'right') {
+    return { y: cross, x: main }
+  }
+
+  if (position === 'left') {
+    return { y: cross, x: -main }
+  }
+
+  return { y: main, x: cross }
+}
```
LighthouseKeeperYN, line 78:
```suggestion
const calculateOffsetCoords = (placement: Placement, offset: Offset): Coords => {
  const { position } = parsePlacement(placement)
  const { main, cross } = parseOffset(offset)

  const positions = {
    top: { y: -main, x: cross },
    bottom: { y: main, x: cross },
    left: { y: cross, x: -main },
    right: { y: cross, x: main },
  }

  return positions[position]
}
```

packages/ui/src/composables/usePopover.ts
```diff
@@ -0,0 +1,196 @@
...
+    bottom: Math.max((coords.y + content.height) - yMax, 0),
+    left: Math.max(xMin - coords.x, 0),
+    right: Math.max((coords.x + content.width) - xMax, 0),
+  }
+}
+
+const clamp = (min: number, v: number, max: number) => Math.max(Math.min(v, max), min)
+
+const calculateClipToEdge = (coords: Coords, content: DOMRect, anchor: DOMRect, root: DOMRect) => {
+  const { top, bottom, left, right } = calculateContentOverflow(coords, content, root)
```
LighthouseKeeperYN, line 98: I'd rather not use destructuring here since those names are used very commonly and can be misinterpreted in below expression in a quick glance.

packages/ui/src/composables/usePopover.ts
```diff
@@ -0,0 +1,196 @@
...
+  const { position, align } = parsePlacement(placement)
+  const { top, bottom, left, right } = calculateContentOverflow(coords, content, root)
+
+  if (top && position === 'top') { return ['bottom', align].join('-') as Placement }
+  if (bottom && position === 'bottom') { return ['top', align].join('-') as Placement }
+  if (right && position === 'right') { return ['left', align].join('-') as Placement }
+  if (left && position === 'left') { return ['right', align].join('-') as Placement }
+
+  return placement
+}
```
LighthouseKeeperYN, line 121:
```suggestion
const getAutoPlacement = (placement: Placement, coords: Coords, content: DOMRect, root: DOMRect): Placement => {
  const { position, align } = parsePlacement(placement)
  const overflow = calculateContentOverflow(coords, content, root)

  const positions = {
    top: ['bottom', align].join('-') as Placement,
    bottom: ['top', align].join('-') as Placement,
    right: ['left', align].join('-') as Placement,
    left: ['right', align].join('-') as Placement,
  }

  return (overflow[position] && positions[position]) || placement
}
```

LighthouseKeeperYN (approved):

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -88,7 +88,7 @@ import VaTimePicker from '../va-time-picker/VaTimePicker.vue'
 import VaInput from '../va-input/VaInput.vue'
 import VaIcon from '../va-icon/VaIcon.vue'
 import VaDropdown from '../va-dropdown/VaDropdown.vue'
-import VaDropdownContent from '../va-dropdown/VaDropdownContent/VaDropdownContent.vue'
+import VaDropdownContent from '../va-dropdown/'
```
RVitaly1978, line 91: is `VaDropdownContent` default import?
redundant `/`

RVitaly1978 (approved):

packages/ui/src/composables/useDomRect.ts
```diff
@@ -0,0 +1,121 @@
...
+    window.addEventListener('resize', cb)
+  })
+
+  onBeforeUnmount(() => {
+    window.removeEventListener('scroll', cb)
+    window.removeEventListener('resize', cb)
+  })
+}
+
+// TODO: investigate if we need to observe scroll on phones.
```
m0ksem, line 15: Remove these comments.

packages/ui/src/components/va-time-input/VaTimeInput.vue
m0ksem: Nice catch! Didn't notice that.

PR #1507, opened 2022-02-15
Fix#1212 VaCounter component

### LINKED ISSUE #429: va-input-number redesign
#9 
Need to redesign va-input-number 
Screenshot 2020-09-29 at 19 56 26

[iView](http://iview.talkingdata.com/#/components/input-number-en)
[Ant Design](https://ant.design/components/input-number/)
[Quasar](https://quasar.dev/vue-components/input#Input-of-number-type)
[Element](https://element.eleme.io/#/en-US/component/input-number#inputnumber)

### LINKED ISSUE #1212: Component idea - stepper
Not ready for implementation, just an idea.

image

https://framework7.io/docs/stepper.html

RVitaly1978:
## Description
close #1212
close #429

changes:
- [x] add component
- [x] add demo and docs examples

chrome-capture
chrome-capture (1)
chrome-capture (2)
chrome-capture
chrome-capture (1)

## Types of changes


packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+    width: { type: String, default: '154px' },
+    color: { type: String, default: 'primary' },
+    outline: { type: Boolean, default: undefined },
+    bordered: { type: Boolean, default: undefined },
+    // icons & buttons
+    increaseIcon: { type: String, default: 'add' },
+    decreaseIcon: { type: String, default: 'remove' },
+    buttons: { type: Boolean, default: false },
+    margins: { type: [String, Number], default: '4px' },
+    flat: { type: Boolean as PropType<boolean>, default: true },
```
LighthouseKeeperYN, line 135: Not a fun of default true values. Could be named "filled" which defaults to false so you can simply pass "filled" flag instead of 'flat="false"' expression

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+  { listen: 'click-prepend-inner', emit: 'click:decrease-icon' },
+  { listen: 'click-append-inner', emit: 'click:increase-icon' },
+])
+
+export default defineComponent({
+  name: 'VaCounter',
+
+  components: { VaInputWrapper, VaIcon, VaButton },
+
+  props: {
```
LighthouseKeeperYN, line 112: Please cast PropType to all props.

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+
+    const onManualInput = (counterValue: string) => {
+      valueComputed.value = counterValue
+    }
+
+    const getRoundDownWithStep = (value: number | string) => {
+      return props.min + props.step * Math.floor((+value - props.min) / props.step)
+    }
+
+    const onManualChange = (counterValue: number | string) => {
```
LighthouseKeeperYN, line 164: Not a fun of 'onEvent' names. There's no way to guess what this handler does.

LighthouseKeeperYN (changes requested):

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+    ...createFieldEmits(),
+    ...useFocusEmits,
+  ],
+
+  inheritAttrs: false,
+
+  setup (props, { emit, attrs }) {
+    const input = ref<HTMLInputElement | HTMLDivElement | undefined>()
+
+    const { isFocused, focus, blur } = useFocus(input, emit)
```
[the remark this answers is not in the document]
RVitaly1978: These methods will be useful when we resolve the problem with `withConfigTransport`

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: done

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: done

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+
+    <input
+      v-if="!$slots.content"
+      class="va-input__content__input"
+      ref="input"
+      type="number"
+      v-bind="{ ...computedInputAttributes, ...inputListeners }"
+      :value="valueComputed"
+      @input="onManualInput($event.target.value)"
+      @change="onManualChange($event.target.value)"
```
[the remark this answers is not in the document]
RVitaly1978: done

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
...
+    </template>
+
+    <input
+      v-if="!$slots.content"
+      class="va-input__content__input"
+      ref="input"
+      type="number"
+      v-bind="{ ...computedInputAttributes, ...inputListeners }"
+      :value="valueComputed"
+      @input="onManualInput($event.target.value)"
```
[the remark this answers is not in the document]
RVitaly1978: done

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,368 @@
+<template>
+  <VaInputWrapper
+    class="va-counter"
+    v-bind="fieldListeners"
+    :class="[$attrs.class, computedClass]"
+    :style="{ ...$attrs.style, ...computedStyle }"
```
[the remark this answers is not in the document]
RVitaly1978: done

LighthouseKeeperYN (approved):

xx13:
@RVitaly1978 

the component looks great. here are some comments related to docs:
- outlined input should also have ghost buttons by default (not outlined); it should be fixed in `buttons` and `widhts`
- change `times` to `copies` in `slots`
- merge 2 lines of notes into one, or use a bullet style for every line with `• ` prefix
- have the same button styles within each variant in `slots`: both outlined buttons in the 2nd variant, and both filled buttons in 3rd one.
- `custom icons and colors`: apply the [following icons and colours](https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/vuestic-ui?node-id=1214%3A3135), the last section

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+      blur,
+    } = useFocus(input, emit)
+
+    const { valueComputed } = useStateful(props, emit)
+
+    const setCountInput = (event: Event) => {
+      valueComputed.value = +(event.target as HTMLInputElement)?.value
+    }
+
+    const setCountChange = (event: Event) => {
```
m0ksem, line 167: Two handlers for same user-event? How about to have debounced `calculateCounterValue`? Or, we can just remove input handler.

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+    const computedStyle: ComputedRef<Partial<CSSStyleDeclaration>> = computed(() => {
+      const style = attrs.style || {}
+      return Object.assign(style, {
+        width: typeof props.width === 'number' ? `${props.width}px` : props.width,
+      })
+    })
+
+    const computedMargins = computed(() => (
+      typeof props.margins === 'number' ? `${props.margins}px` : props.margins
+    ))
```
m0ksem, line 274: It is a computedMargin (left and right), not margins.

PX can be changed with FN, same for width. It can be a function from utils folder.

```suggestion
    const computedMargin = computed(() => safeCSSLength(props.margins))
```

m0ksem (changes requested):
Extra controls on firefox.

image

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
+<template>
+  <VaInputWrapper
+    class="va-counter"
+    v-bind="fieldListeners"
+    :class="computedClass"
+    :style="computedStyle"
+    :color="$props.color"
+    :readonly="$props.readonly"
+    :disabled="$props.disabled"
+    :messages="$props.messages"
```
m0ksem, line 10: How about to use `extractComponentProps` function? Also, how about rules? Do we respect validation in this component?

m0ksem (changes requested):

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+      isFocused,
+      // will be useful when we resolve problem with 'withConfigTransport'
+      focus,
+      blur,
+    } = useFocus(input, emit)
+
+    const { valueComputed } = useStateful(props, emit)
+
+    const setCountInput = (event: Event) => {
+      valueComputed.value = +(event.target as HTMLInputElement)?.value
```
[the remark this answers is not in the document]
RVitaly1978: Done

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: Done

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: 1. I decided to use lodash `pick` function to choose props for `VaInputWrapper`.
2. We discussed and decided that this component can be without validation. For validation, it has props `min`, `max` and `step`

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: The `setCountInput` handler is required for manually entering values (props `manual-input`)

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+      name: props.increaseIcon,
+      ...(!disabledIncreaseAction.value && { onClick: increaseCount }),
+    }))
+
+    const squareCorners = computed(() => (
+      (typeof props.margins === 'string' ? parseFloat(props.margins) : props.margins) === 0
+    ))
+
+    const decreaseButtonProps = computed(() => ({
+      ...pick(props, ['outline', 'rounded', 'color', 'textColor']),
```
m0ksem, line 237: Why decreaseButtonProps and increaseButtonProps are different?

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+    }))
+
+    const increaseIconProps = computed(() => ({
+      class: { 'va-counter__icon--inactive': disabledIncreaseAction.value },
+      color: colorComputed.value,
+      name: props.increaseIcon,
+      ...(!disabledIncreaseAction.value && { onClick: increaseCount }),
+    }))
+
+    const squareCorners = computed(() => (
```
m0ksem, line 232: isSquareCorners 😶

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+
+    const isMaxReached = computed(() => {
+      if (!props.max) { return false }
+
+      return props.step
+        ? Number(valueComputed.value) > (props.max - props.step)
+        : Number(valueComputed.value) >= props.max
+    })
+
+    const disabledDecreaseAction = computed(() => (
```
m0ksem, line 198: doDisableDecreaseAction or isDecreaseActionDisabled. 😶

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+    }))
+
+    const classComputed = computed(() => ([
+      attrs.class,
+      { 'va-counter--input-square': squareCorners.value },
+    ]))
+
+    const styleComputed: ComputedRef<Partial<CSSStyleDeclaration>> = computed(() => {
+      const style = attrs.style || {}
+      return Object.assign(style, {
```
m0ksem, line 268:
```suggestion
      return { width: safeCSSLength, ...style }
```

This will create new object, style object will not be modifed.
Style.width will override width from props.

packages/docs/src/pages/ui-elements/counter.vue
```diff
@@ -0,0 +1,15 @@
...
+  <docs-content :config="config" />
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+import DocsContent from '../../components/DocsContent.vue'
+import config from '../../page-configs/ui-elements/counter/page-config'
+
+export default defineComponent({
+  name: 'DocsAvatar',
```
m0ksem, line 11: > DocsAvatar :)

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    const calculateCounterValue = (counterValue: number) => {
+      if (props.min && counterValue < props.min) {
+        valueComputed.value = props.min
+        return
+      }
+
+      if (props.max && (counterValue > props.max)) {
+        valueComputed.value = getRoundDownWithStep(props.max)
```
m0ksem, line 178:
```suggestion
        valueComputed.value = props.max
```

Can't it be just props.max?

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -0,0 +1,391 @@
...
+      valueComputed.value = Number(target?.value)
+    }
+
+    const setCountChange = ({ target } : { target: HTMLInputElement }) => {
+      calculateCounterValue(Number(target?.value))
+    }
+
+    const getRoundDownWithStep = (value: number) => {
+      if (!props.min || !props.step) { return value }
+      return props.min + props.step * Math.floor((value - props.min) / props.step)
```
m0ksem, line 168: Do we really need props.min here?

```suggestion
      return props.step * Math.floor((value) / props.step)
```

m0ksem (approved):
Good work!

packages/docs/src/pages/ui-elements/counter.vue
RVitaly1978: fixed :)

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: They have different computed `disabled` attr.
Refacrored a bit

packages/ui/src/components/va-counter/VaCounter.vue
m0ksem: Definitely we need this. props.min is used as offset here. So if 1 is min, then all values can be only 1 + step * n.

packages/ui/src/components/va-counter/VaCounter.vue
m0ksem: > They have different computed `disabled` attr. Refacrored a bit

Oh. I see you forget to remove 'outline' from pick, because it used a bit down

```ts
      outline: props.flat && props.outline,
```

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: refactored

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: Not always, since the `step` may not be a multiple of the `max - min`

packages/ui/src/components/va-counter/VaCounter.vue
RVitaly1978: Yes. If the user enters a value manually, then we must round it to the nearest valid value, taking into account the initial value (`min`) and the step size (`step`).
I have added comments to the code

packages/ui/src/components/va-counter/VaCounter.vue
m0ksem: So if component has value 24 with step=4 and max=25 there is no way to go upper more than 24. Make sense.
Okay.

m0ksem:
Unable to select text with manual input prop set to true.

m0ksem (changes requested):
https://github.com/epicmaxco/vuestic-ui/pull/1507#issuecomment-1088714507

m0ksem (approved):

PR #1515, opened 2022-02-17
Feat/required mark

### LINKED ISSUE #1497: Could we add asterisk after label for form control if we want controls be required.

With the current version, if we want to mark the specific control required, we can add prop `message` to achieve it, but let us think about that, in most scenarios, we just want to let user know the specific control is required without hint message. 

We should let control have ability add asterisk after label if we want the control to be required.
image

LighthouseKeeperYN:
close #1497

Adds red asterisk to the input label when input is required

## Types of changes


packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -36,6 +36,12 @@
                 :style="{ color: colorComputed }"
               >
                 {{ label }}
+                <span
```
m0ksem, line 39: Would be nice to have a slot here. I want required-mark to be `🐱` instead of `*` :)

packages/ui/src/composables/useForm.ts
```diff
@@ -3,13 +3,15 @@ import { computed, PropType } from 'vue'
...
   readonly: boolean;
+  required?: boolean;
   id?: string | number;
   name?: string | number;
 }
 
 export const useFormProps = {
   disabled: { type: Boolean, default: false },
   readonly: { type: Boolean, default: false },
+  required: { type: Boolean, default: false },
```
m0ksem, line 14: We need `show-required-mark` prop. For example, we have a form where all inputs required.

packages/ui/src/components/va-input/VaInput.demo.vue
```diff
@@ -269,6 +269,13 @@
         </template>
       </va-input>
     </VbCard>
+    <VbCard title="Required">
```
m0ksem, line 272: Would be nice to have same demo on select, time-input etc. Just to be sure that it works here too.

m0ksem (changes requested):
Here are a few suggestions. 

Also, I think `required` prop should also add required validation rule automatically. This rule should be customizable and `required` message must support i18n.

packages/ui/src/composables/useForm.ts
Derranion: From my point of view we need this mark to just highlight for the user that this field is required.

Derranion (approved):
Well done, would be great to have slot with default value as " * " instead, as @m0ksem [suggested](https://github.com/epicmaxco/vuestic-ui/pull/1515#discussion_r809325274).

Could you please add a few words to our [notion description](https://www.notion.so/Vuestic-ui-10e698cacd5846688647e1d8482277c9?p=7b60ebf22352445d80c9f307b6c1234a) in the end.

Thanks.

PR #1523, opened 2022-02-21
Fix#1493 Fix `VaDataTable` selection behavior, add `sticky-header`

### LINKED ISSUE #1492: [va-datatable] should support fixed header which AG grid supported
**Vuestic-ui version:** 1.3.4

### LINKED ISSUE #1493: va-data-tabel loose selected items on filtering, when those are not in the result
**Vuestic-ui version:** 1.3.4

**Steps to reproduce**
go to https://vuestic.dev/en/ui-elements/data-table#pagination

Pagination example has selection and filter too:

at the example for pagination select on and filter for a different line (e.g. select Antonette and  type "Sam" in the filter field)
you will loose the selection BUT you wont loose selection if you type "Ant" - so the behaviour isn't really be intended, or is it?

selected items is a v-model array and should be manipulated via script only 

on filtering, selected item should stay selected, please!

selection of item, that are not in the filtered result are lost

**Other information**
Peek 2022-02-09 10-32
Peek 2022-02-09 10-33

RVitaly1978:
## Description
close #1492
close #1493

changes:
- [x] update selectable behavior
- [x] add sticky-header behavior
- [x] update demo and docs examples

chrome-capture
chrome-capture (1)
chrome-capture

## Types of changes


packages/ui/src/components/va-data-table/hooks/useColumns.ts
```diff
@@ -71,25 +31,22 @@ export const buildTableColumn = (source: TTableColumnSource, initialIndex: numbe
...
-const buildColumnsFromItems = (rawItems: Ref<ITableItem[]>) => {
-  return Object.keys(merge({}, ...rawItems.value)).map(buildTableColumn)
+const buildColumnsFromItems = (items: Ref<ITableItem[]>) => {
+  return Object.keys(merge({}, ...items.value)).map(buildTableColumn)
 }
 
-const buildNormalizedColumns = (rawColumns: Ref<TTableColumnSource[]>) => {
-  return rawColumns.value.map(buildTableColumn)
+const buildNormalizedColumns = (columns: Ref<TTableColumnSource[]>) => {
+  return columns.value.map(buildTableColumn)
```
m0ksem, line 39:
```suggestion
const buildNormalizedColumns = (columns: TTableColumnSource[]) => {
  return columns.map(buildTableColumn)
```

How about remove Ref from these methods. This will remove overwhelming `toRef`.

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -252,100 +293,38 @@ export default defineComponent({
...
-    },
-    striped: {
-      type: Boolean,
-      default: false,
-    },
+    columns: { type: Array as PropType<TTableColumnSource[]>, default: () => [] as TTableColumnSource[] },
+    items: { type: Array as PropType<ITableItem[]>, default: () => [] as ITableItem[] },
+    modelValue: { type: Array as PropType<ITableItem[]> }, // selectedItems
+    sortingOrder: { type: String as PropType<TSortingOrder> }, // model-able
+    sortBy: { type: String as PropType<string> }, // model-able
```
m0ksem, line 303: This must work without PropType...

m0ksem (changes requested):
How about a sticky header demo? Would be nice to have different PRs for bug fixes and new features.

packages/ui/src/components/va-data-table/hooks/useColumns.ts
RVitaly1978: done

packages/ui/src/components/va-data-table/VaDataTable.vue
RVitaly1978: done

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -429,6 +430,7 @@ export default defineComponent({
 
   &--sticky {
     overflow-y: auto;
+    height: var(--sticky-table-height);
```
m0ksem, line 433: Can't we set height if the table is not sticky? 🤔

m0ksem (changes requested):

packages/ui/src/components/va-data-table/VaDataTable.vue
RVitaly1978: Added

m0ksem (approved):

PR #1530, opened 2022-02-23
feat/optional services

m0ksem:
Do this after https://github.com/epicmaxco/vuestic-ui/pull/1501

TODO: 
- [x] easy way to import CSS - ? - We could try to move to vite, then we can create a css inject plugins. 
   Solved by https://github.com/epicmaxco/vuestic-ui/pull/1678. Exports allow us to easy import style `import "vuestic-ui/styles/essential.css"`
- [x] docs
image

- [x] Types for vue context?

Here a list of services we have:
- Components config (props config) 
For now it's required, but we can easily make it none required. It has no sense to be remove for now, since we have withTransportConfig. Later, if we move to composable, this will has no sense at all, because GlobalConfig will not be a plugin anymore.
- ColorsConfigPlugin is very small plugin, a lot of components can use CSS variables instead overwhelming computeds.
- Icons config - we have a extra presets and aliases. We can remove them from 'createVuesticEssential'.
- VaDropdown and VaToastPlugins can be also removed if user will not use Options API.
- CSS Plugins - its a thing that we would like to have, but I'm not sure how can we implement this.
We have a few css libs, lke typography, grid, etc. They are totally optional. Would be nice to have CSS in JS plugins so user can simply import them. This can be done using style-inject, but I don't know how to deal with HRM.


packages/ui/src/components/va-toast/plugin/index.ts
```diff
@@ -1,21 +1,29 @@
...
+  close (id: string) {
+    closeById(id)
+  },
 
-      close (id: string) {
-        closeById(id)
-      },
+  closeAll (allApps = false) {
+    closeAllNotifications(allApps ? undefined : app?._context)
+  },
```
LighthouseKeeperYN, line 16: closeAllNotifications doesn't seem to accept any arguments. What the point of this expression?

m0ksem: My bad. Took this function from [useToast hook PR#1464](https://github.com/epicmaxco/vuestic-ui/pull/1464/files#diff-d6cb9ad66be91ffcf23cbdbc6e22f8f1449d8ad2b871c8b811973ea6b9b94f98R126)

Please, take a look on idea, but not on implementation. This is important architecture rework, we want it to be flexible as maximum.

m0ksem:
GlobalConfigPlugin could be renamed into VaGlobalConfigPlugin same as VaToastPlugin.

Also, GlobalConfigPlugin do not proved any methods in vue context. This could be helpful for debugging, because now I need to add useGlobalConfig into some setup function...

packages/docs/src/locales/en/en.json
```diff
@@ -1729,7 +1729,7 @@
   "toast": {
     "title": "Toast",
     "summaryText": "Component for sending a notification to the user. To make it easier to create toast we have a shorthands. For options API it is `$vaToast`. For composition api we have `useToast` hook. Hook and global object have similar API.",
-    "optionsAPI": "In options API you can use ToastPlugin (automatically installed with VuesticPlugin). `$vaToast` have a few methods: init, close, closeAll.",
+    "optionsAPI": "In options API you can use VaToastPlugin (automatically installed with cresteVuestic). `$vaToast` have a few methods: init, close, closeAll.",
```
asvae, line 1732:
```suggestion
    "optionsAPI": "In options API you can use VaToastPlugin (automatically installed with createVuestic). `$vaToast` have a few methods: init, close, closeAll.",
```

packages/docs/src/locales/en/en.json
```diff
@@ -3038,18 +3038,15 @@
       "title": "Tree shaking",
       "description": "If you plan to use only several components in your application and thus reduce the size of your bundle, you can use the Vuestic ESM build. Webpack (vue-cli) and Vite do that automatically.",
       "example": {
-        "title": "First, you don't need to use `VuesticPlugin` since it registers all the `vuestic` components globally. Instead, you can import specific components from the `vuestic-ui` package.",
-        "footer": "Alternatively, you can import them directly in your `.vue` files."
+        "title": "First, you don't need to use `createVuestic` since it registers all the `vuestic` components globally. We have `createVuesticEssential` instead, which register only essential plugins. You can specify components to declare globally. Or you can import them later.",
```
asvae, line 3041: core vs essentials vs full we discussed before.
We can leave essentials and full in docs, but all 3 should be depicted in notion so that our strategy is clear.

packages/docs/src/locales/en/en.json
```diff
@@ -3038,18 +3038,15 @@
...
-        "ToastInstall": "ToastInstall - used by `VaToast` component.",
-        "ColorHelpersPlugin": "ColorHelpersPlugin - used to create reactive CSS variables. Requires `GlobalConfigPlugin`. [Read more](/services/colors-config).",
-        "DropdownPopperSubplugin": "DropdownPopperSubplugin - used by `VaDropdown` component.",
-        "example": {
-          "title": "All of these plugins are also available through the `VuesticPluginsWithoutComponents`:"
-        }
+        "title": "Here is a list of plugins that you can use with Vuestic:",
+        "GlobalConfigPlugin": "GlobalConfigPlugin (essential) - used for `VaConfig` component and global props reassignment. [Read more](/services/global-config).",
+        "ColorHelpersPlugin": "ColorHelpersPlugin (essential) - used to create reactive CSS variables. Requires `GlobalConfigPlugin`. [Read more](/services/colors-config).",
+        "ToastInstall": "VaToastPlugin - provides methods for creating Toasts within vue context. [Read more](/ui-elements/toast)",
```
asvae, line 3048: VaToastPlugin?

asvae (approved):
I like new interface a lot! Thanks.

m0ksem:
For tree-shaking we want to have following structure

```ts
import { createVuesticEssential, VaToastPlugin } from 'vuestic-ui'
import 'vuestic/styles/essential.css'

...

createVuesticEssential({
   plugins: [VaToastPlugin(options)]
})

```

So, internal plugins must be a function fabric, as createVuestic. But, user should be able to pass it without call as rollup plugin. So we need to check if plugin is function or object.

PR #1541, opened 2022-03-04
Add option grouping to select component

### LINKED ISSUE #1373: optgroup in va-select
### Discussed in https://github.com/epicmaxco/vuestic-ui/discussions/1240

<div type='discussions-op-text'>

<sup>Originally posted by **gitguidev** November 16, 2021</sup>
How can I use optgroup in va-select?</div>

Present in html select, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup.
We might want to have something similar for vuestic select.
Implementation could be different though, as long as it solves the same problem.

LighthouseKeeperYN:
close #1373

## Description
VaSelect options can now be grouped by providing "group" value. 

Also refactored VaSelectOptionsList to composition api.
```
options: [
  { id: '0', text: 'one', value: 'one', group: 'foo' },
  { id: '1', text: 'two', value: 'two' },
  { id: '2', text: 'three', value: 'three', group: 'bar' },
  { id: '3', text: 'four', value: 'four', group: 'bar' },
  { id: '4', text: 'five', value: 'five', group: 'foo' },
],
```
Untitled

## Types of changes


packages/ui/src/composables/useSelectableList.ts
```diff
@@ -3,17 +3,27 @@
...
   trackBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'value' },
   disabledBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'disabled' },
+  groupBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'group' },
 }
 
-export function useSelectableList (props: any) {
+export function useSelectableList (props: SelectableListProps) {
   const isSelectableListComponent = true
 
   const isStringOrNumber = (option: any): boolean => {
```
m0ksem, line 29:
```suggestion
  const isStringOrNumber = (option: any): option is (string | number)=> {
```

packages/ui/src/components/va-select/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -95,143 +70,180 @@ const SelectOptionListPropsMixin = Vue.with(SelectOptionListProps)
...
-) {
-  itemRefs: Record<number, any> = {}
-
-  created () {
-    watch(() => this.$props.hoveredOption, (newOption: any) => {
-      if (newOption) {
-        this.scrollToOption(newOption)
+  props: {
+    ...useColorProps,
+    options: { type: Array as PropType<any[]>, default: () => [] },
```
m0ksem, line 75: Let's have a SelectOption type.

```suggestion
    options: { type: Array as PropType<SelectOption[]>, default: () => [] },
```

m0ksem (approved):

packages/ui/src/components/va-select/VaSelectOptionList/VaSelectOptionList.vue
LighthouseKeeperYN: Aren't options can be literally of any type?

m0ksem: It's actually `stringOrNumber | Record<string, string>`. Anyway typing SelectOption would make code easier to understand, for example you can use this type in hoveredOption instead of repeating `null | string | number | Record<string, any>` `selectOption | null`. This will not impact performance so it;s just nice to have.

asvae: To add up, select generally functions as `Select<T>`, where `T` is option type. So having same type for all options is pretty important.

asvae (approved):
Cool. Thanks!

m0ksem (approved):

packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -192,7 +193,7 @@ export default defineComponent({
     ...useFormProps,
 
     modelValue: {
-      type: [String, Number, Object, Array] as PropType<string | number | Record<string, any> | any[]>,
+      type: [String, Number, Object] as PropType<SelectableOption>,
```
Derranion, line 196: Hey @LighthouseKeeperYN 

Could you tell why `Array` type was removed from `VaSelect` `modelValue`?

Because it has Array as its value most of the time.

Issue #825, opened 2021-06-01
useToast() for Composition API


m0ksem:
> I have a question! Still new to Vue 3, how do I get $vaToast.init({ message: "danger color", color: "danger" }); working in Typescript in the setup() function?

$vaToast can be used only in Options API for now.

smellyshovel:
I did some research and here's what I've found out.

First of all, the `$vaToast` is actually just an app's global property. That means that you can still use `$vaToast.init(...)` in your templates (even if you use the Composition API) and everything works jsut fine.

If instead you want a wrapper (i.e. you want to use smtn like `showRedToast('Message')` instead of constantly using `$vaToast.init({ message: 'Message', color: '#ff0000' })` throughout the template (which, I believe, is the case of the issue)) *and* you want to return the wrapper from the `setup` function then you can simply do the following thing:

```typescript
import { getCurrentInstance } from 'vue';

export default defineComponent(function VaToastCompositionDemo(props, context) {
  const app = getCurrentInstance()
  const vaToast = app.appContext.config.globalProperties.$vaToast

  return {
    showToast: vaToast.init
  }
})
```

and then use the returned method in the template:

```
<button @click="showToast('test')">notification</button>
```

@m0ksem Do you think we steel should implement a `useToast` composable? It might be better to rethink the whole VaToast's concept instead (e.g. use inject/provide or idk refuse from using `globalProperties` at all and provide methods for working with toasts as a straightforward `export`s).

m0ksem:
@smellyshovel, I don't think we need to use `provide/inject` for toasts. In my opinion, `provide/inject` should be used only for data flow.

For now, we have $vaToast because we simply took it from Options API.

I think we should have a simple fabric for toasts. It looks like it's Vue utility. I believe for such utilities we should use a composable approach.

I think @asvae might help us.

Baloneo:
How to use in axios?
```
service.interceptors.response.use(
    (response) => {
        const res = response.data
        $vaToast.init('Hello')
        return res
    }
```

sergey-kravcov:
> How to use in axios?
> 
> ```
> service.interceptors.response.use(
>     (response) => {
>         const res = response.data
>         $vaToast.init('Hello')
>         return res
>     }
> )
> ```

Hello, could you please describe your problem in more detail so that we can help you

m0ksem:
> > How to use in axios?
> > ```
> > service.interceptors.response.use(
> >     (response) => {
[...]

PR #1543, opened 2022-03-07
Feat/toast composition hook

m0ksem:
Just a little fix. Do not use window in global context. This broke SSR.

closes https://github.com/epicmaxco/vuestic-ui/issues/825


asvae (approved):
Looks great!

PR #1559, opened 2022-03-14
Adding version badges to several Readme.md files

### LINKED ISSUE #1361: Improve versions handling solution
[See Quasar implementation](https://github.com/quasarframework/quasar)

Basically the idea is:

* In root readme - have latest npm versions for all our packages.
* In package readme - have same badge, but only for specific package.
* If package doesn't have npm badge - use package.json version.

Also see https://github.com/epicmaxco/vuestic-ui/pull/1234

xiongmao86:
## Description

Adding badges to root readme, and to package readmes. Closes #1361

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,32 +1,41 @@
 <p align="center">
-    <a href="https://vuestic.dev" target="_blank">
-        <img alt="Vuestic UI Logo" width="400" src="packages/ui/src/assets/vuestic-ui-logo.png">
-    </a>
-        <br>
-    Vue.js 3.0 UI Library
-    <br>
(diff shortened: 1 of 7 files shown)


xiongmao86:
Thanks for adding the links. @daurenSab. I missed that.

daurenSab:
You have done most of the work. Versions of packages are now immediately visible in Readme.md files
Thanks for your effort

m0ksem:
@daurenSab, I think we don't need version tags for `bundlers-test` and `vue-cli-example` packages in Readme.

Derranion:
@m0ksem,
I removed them, maybe we don't need docs badge as well?

docsBadge

PR #1575, opened 2022-03-17
Set undefined as default value for VaParallax target

### LINKED ISSUE #1572: VaParallax target prop must be undefined
By default, its emtpy string, is used in querySelector. So querySelector by '' throws error. If there is no string, let's use document.body. 

Think about ssr, if there is no document component must work.

Sync271:
## Description
Fixes #1572 
When it's an empty string, querySelector throws an error as it's only checking if it `string` type or not. Setting the default to `undefined` will let it access the parent element.

## Markup:

```diff
- target = prop<Element | string>({ type: [Object, String], default: '' })
+ target = prop<Element | string | undefined>({ type: [Object, String], default: undefined })
```
<details>

```vue
target = prop<Element | string | undefined>({ type: [Object, String], default: undefined })
```
</details>

in:
https://github.com/epicmaxco/vuestic-ui/blob/8a33fbde75b6092b9842268791267b7b741a1323/packages/ui/src/components/va-parallax/VaParallax.vue#L62-L66
not sure if it has to be changed to the following, when "this.$props.target" is `undefined`, there's no need for `this.$props.target || this.$el.parentElement`:
```vue
  get targetElement () {
    return typeof this.$props.target === 'string'
      ? document.querySelector(this.$props.target)
      : this.$el.parentElement
  }
```

## Types of changes


m0ksem (approved):
Thanks!

> not sure if it has to be changed to the following, when "this.$props.target" is undefined, there's no need for this.$props.target || this.$el.parentElement:

User can actually pass HTMLElement instead of string.

---

I'm just curious about this.$el.parentElement... Shouldn't it be a window as default scroll container? 

We can make it even better, if we will try to find the first parent of VaParallax that has overflow scroll property.

Sync271:
> We can make it even better if we will try to find the first parent of VaParallax that has overflow scroll property.

Wouldn't that require the target to accept the `null` type too, since `.parentNode` would return `ParentNode | null` or `Element | null` ?
Wouldn't that end up changing types of a lot of things that depend on the target element?
PS: Sorry, I am new to vue

m0ksem:
> > We can make it even better if we will try to find the first parent of VaParallax that has overflow scroll property.
> 
> Wouldn't that require the target to accept the `null` type too, since `.parentNode` would return `ParentNode | null` or `Element | null` ? Wouldn't that end up changing types of a lot of things that depend on the target element? PS: Sorry, I am new to vue

Yes. Also, targetElement can be null on SSR, because nodejs server doesn't have document or window globals, like browser.

Sync271:
> Yes. Also, targetElement can be null on SSR, because nodejs server doesn't have document or window globals, like browser.
Alright, I think I've made the necessary changes, some images for the parallax demo are broken so I used the same image for all cases to test. Let me know if there are any other changes required.

m0ksem:
> > Yes. Also, targetElement can be null on SSR, because nodejs server doesn't have document or window globals, like browser.
> 
> Alright, I think I've made the necessary changes, some images for the parallax demo are broken so I used the same image for all cases to test. Let me know if there are any other changes required.

I updated your PR a bit. Removed type casting (`as Element`), added fallback to `document.body` if there is no parent with scroll, added error message if target selector returns null. I also updated our docs demos, because target can be found automatically now 😊. 

Appreciate your help!

m0ksem (approved):

packages/ui/src/components/va-parallax/VaParallax.vue
```diff
@@ -59,10 +59,27 @@ export default class VaParallax extends mixins(
...
+  get targetElement (): Element {
+    if (typeof this.$props.target === 'string') {
+      const element = document.querySelector(this.$props.target)
+      if (element) { return element }
+      throw new Error('VaParallax target prop got wrong selector. Target is null')
+    }
+
+    if (this.$props.target) { return this.$props.target }
+
+    return this.getScrollableParent(this.$el.parentElement) || document.body
```
LighthouseKeeperYN, line 82:
```suggestion
    return this.$props.target || this.getScrollableParent(this.$el.parentElement) || document.body
```

LighthouseKeeperYN (approved):

packages/ui/src/components/va-parallax/VaParallax.vue
```diff
@@ -59,10 +59,27 @@ export default class VaParallax extends mixins(
     }
   }
 
-  get targetElement () {
-    return typeof this.$props.target === 'string'
-      ? document.querySelector(this.$props.target)
-      : this.$props.target || this.$el.parentElement
+  getScrollableParent (element: Element | null): Element | null {
+    if (element === null) {
+      return null
```
LighthouseKeeperYN, line 64:
```suggestion
      return window
```

asvae:
Thanks.
Doesn't work for me for some reason :<.

image

Sync271:
> Thanks. Doesn't work for me for some reason :<.

Should be fixed now. I implemented the suggested changes without testing, so that's on me. I reverted the changes and fixed the broken `serve:book` links for images.

RVitaly1978 (approved):

asvae:
@LighthouseKeeperYN can you take a look if we can merge it manually?

LighthouseKeeperYN:
Resolved conflicts and merged in #1621

Thanks for your contribution!

PR #1578, opened 2022-03-18
Refactor/va-progress-bar

m0ksem:
Changes:
- Split components into two different directories. 
- Rewrite to Composition API
- Removed junk (we had two same files for each -circle and -bar components)
- Made progress-bar rounded instead of rounded-square.


packages/ui/src/components/va-progress-bar/VaProgressBar.vue
```diff
@@ -1,112 +1,96 @@
...
+    buffer: { type: Number, default: 100 },
+    rounded: { type: Boolean, default: true },
+    reverse: { type: Boolean, default: false },
+  },
+
+  setup (props) {
+    const { getColor } = useColors()
+
+    const isLarge = computed(() => props.size === 'large')
+    const isTextSize = computed(() => typeof props.size === 'string' && ['small', 'medium', 'large'].includes(props.size))
```
LighthouseKeeperYN, line 53:
```suggestion
    const isTextSize = computed(() => typeof props.size === 'string' && ['small', 'medium', 'large'].includes(props.size))
```
 const isTextSize = computed(() => props.size instanceof VaProgressSize

LighthouseKeeperYN (approved):

rustem-nasyrov (approved):

packages/ui/src/components/va-progress-bar/VaProgressBar.vue
m0ksem: Instanceof checks prototype constructors and can not be used on abstract typescript type.

LighthouseKeeperYN: Oh, my bad

PR #1579, opened 2022-03-20
Add translations for code blocks and examples.

### LINKED ISSUE #1370: Add translations for docs code blocks
We want to translate the text inside [code blocks](https://vuestic.dev/en/contribution/documentation-page#code) and code previews for examples.

[Page on docs](https://vuestic.dev/en/contribution/translation#translations-inside-code-blocks) describes how to do it.

xiongmao86:
## Description
Add translations for code blocks in code-examples and examples.

fixes #1370 

## Types of changes

### DIFF
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -1278,7 +1278,30 @@
     "api": "API",
     "faq": "FAQ",
     "code": {
-      "or": "or"
+      "or": "or",
+      "folderStructure": {
+        "componentDirectory": "component directory",
+        "componentItself": "component itself",
+        "componentDemo": "component demo",
+        "componentTests": "compoennt tests"
+      },
+      "npmScripts": {
+        "runVueBookDevServer": "Run vue-book dev server",
+        "generateFilesForComponent": "Generate all needed files for component (component itself, docs, tests)",
+        "buildVueBook": "Build vue-book bundle",
+        "lint": "Lint everything",
+        "runTests": "Run tests",
+        "runVuepressDevServer": "Run vuepress dev server",
+        "buildVuepress": "Build vuepress bundle"
+      },
+      "aliasAndFontExample": {
+        "twitterColor": "Twitter blue brand color",
+        "VuesticColor": "Vuestic green brand color",
+        "from": "from",
+        "alias": "alias",
+        "font": "font",
+        "to": "to"
+      }
     },
     "examplesTexts": {
       "button": "Your button"
--- a/packages/docs/src/page-configs/contribution/guide/code-examples/folder-structure.ts
+++ b/packages/docs/src/page-configs/contribution/guide/code-examples/folder-structure.ts
@@ -1,6 +1,6 @@
 export const componentFolderStructure = `
-va-checkbox // component directory
-|- VaCheckbox.vue // component itself
-|- VaCheckbox.demo.vue // component demo
-|- VaCheckbox.spec.js // component tests
+va-checkbox // {{ $t('all.code.folderStructure.componentDirectory') }}
+|- VaCheckbox.vue // {{ $t('all.code.folderStructure.componentItself') }}
+|- VaCheckbox.demo.vue // {{ $t('all.code.folderStructure.componentDemo') }}
+|- VaCheckbox.spec.js // {{ $t('all.code.folderStructure.componentTests') }}
 `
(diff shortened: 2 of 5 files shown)


xiongmao86:
Hi there, I am currently adding translation for code blocks and code examples. When I add translations for button docs, there are translation in the examples, but it doesn't cover all the text. 

For example in `Default.vue` it doesn't include 'Button' in translation. I can add all text to translation, but I am wondering whether that is a good idea. What should I do?

image

asvae:
Hey and thanks for your effort 🤗.

> I can add all text to translation, but I am wondering whether that is a good idea. What should I do?

We try our hardest not to include any translations in examples - makes them harder to support. So for majority of cases - no translation is the way to go. As text either holds no meaning or is a keyword.

packages/docs/src/page-configs/getting-started/configuration-guide/code-examples/icons.ts
```diff
@@ -1,6 +1,6 @@
 export const iconsInstall = `
 yarn add material-design-icons-iconfont -D
-// {{ $t('all.code.or') }}
+// {{ $t('configurationGuide.icons.code.or') }}
```
asvae, line 3: Any reason not to use "all.code.or"?

packages/docs/src/page-configs/styles/typography/examples/blockquote.vue
```diff
@@ -1,12 +1,10 @@
 <template>
   <blockquote class="va-blockquote">
     <p>
-      BQ: Let’s talk about meat fondue recipes and what you need to know
```
asvae, line 4: Filler text we generally don't translate. :)

Doesn't give any info to end user.

packages/docs/src/page-configs/styles/typography/examples/blockquote.vue
```diff
@@ -1,12 +1,10 @@
...
-      first. Meat fondue also known as oil fondue is a method of cooking
-      all kinds.
+      {{ $t('typography.examples.blockQuote.content') }}
     </p>
     <p>
-      <i>— Mister Lebowski</i>
```
asvae, line 9: Though we actually translate stuff like this in [vuestic-admin](https://github.com/epicmaxco/vuestic-admin), there it's used as an example/boilerplate of proper application.

packages/docs/src/locales/en/en.json
```diff
@@ -3513,7 +3522,76 @@
...
+      "headings": {
+        "display1": "Display 1 Heading",
+        "display2": "Display 2 Heading",
+        "display3": "Display 3 Heading",
+        "display4": "Display 4 Heading",
+        "display5": "Display 5 Heading",
+        "display6": "Display 6 Heading"
+      },
+      "links": {
+        "defaultLink": "Default Link",
```
asvae, line 3544: 🥲 
In these cases text is either keyword (from "Secondary Link" user can somewhat derive that class is `link-secondary`, which is not the case for translated text) or filler text.

So not applicable for translation.

I [updated docs](https://github.com/epicmaxco/vuestic-ui/pull/1580) to better represent this idea.

packages/docs/src/page-configs/getting-started/configuration-guide/code-examples/icons.ts
xiongmao86: Thanks @asvae, I try organize translation around the `page-config` which the majority of doc info put together, I think it would be easier to find and organize. And put them under `all.code` is harder to seen the context, and eventually harder to find.

packages/docs/src/page-configs/styles/typography/examples/blockquote.vue
xiongmao86: Thanks, I'll revert all filler text.

packages/docs/src/page-configs/getting-started/configuration-guide/code-examples/icons.ts
xiongmao86: I have reread the posts, is it because there are few changes to make, so I can put them all inside `all.code` category?

xiongmao86:
I have skim through the examples of ui-element and found `mode.vue` example of `date-input`, and `Advanced.vue` of `color-input` is not used in related `page-config`.

packages/docs/src/page-configs/getting-started/configuration-guide/code-examples/icons.ts
asvae: Checked code and things look good :).

Using all is not mission critical.

xiongmao86:
Thanks, @asvae, I have difficult deciding which text should in translation, there are some text that lay between the two, some text that put in translation is ok, and don't is also seem good. I think a third perspective from some one that is not familiar with English would be very useful.  So I think most of the work that can do is done here.

xiongmao86:
Thanks for the issue, @Derranion. Not very confident opening one.

PR #1585, opened 2022-03-23
Refactor/va rating

m0ksem:
image

refactoring to composition api
fixed numbers color


packages/ui/src/components/va-rating/VaRating.vue
```diff
@@ -1,208 +1,101 @@
...
+            :item-number="itemNumber"
+            @click="props.onClick"
+          />
+        </template>
+      </va-rating-item>
     </div>
-    <span class="va-rating__text-wrapper" v-if="$props.texts.length === $props.max" :style="{ color: computeColor($props.textColor) }">
-      {{ $props.texts[Math.round(valueProxy) - 1] }}
+    <span
+      v-if="$props.texts && $props.texts.length === $props.max"
```
LighthouseKeeperYN, line 32: $props.texts && 
```suggestion
      v-if="$props.texts.length === $props.max"
```
props.texts has array as default value so it's redundant

packages/ui/src/components/va-rating/components/VaRatingItem/VaRatingItem.vue
```diff
@@ -0,0 +1,124 @@
...
+  },
+
+  emits: ['update:modelValue', 'click', 'hover'],
+
+  setup (props, { emit }) {
+    const rootEl = ref<HTMLElement>()
+    const [modelValue] = useSyncProp('modelValue', props, emit, RatingValue.EMPTY)
+    const hoveredValue = ref<number | null>(null)
+
+    const visibleValue = computed(() => hoveredValue.value ? hoveredValue.value : modelValue.value)
```
LighthouseKeeperYN, line 58:
```suggestion
    const visibleValue = computed(() => hoveredValue.value || modelValue.value)
```

packages/ui/src/components/va-rating/hooks/useVaRatingColors.ts
```diff
@@ -0,0 +1,47 @@
...
+  const backgroundComputed = computed(() => {
+    if (props.modelValue === RatingValue.HALF) {
+      return `linear-gradient(90deg, ${computedColor.value} 50%, ${backgroundColor.value} 50%`
+    }
+
+    if (props.modelValue === RatingValue.EMPTY) {
+      return backgroundColor.value
+    }
+
+    return computedColor.value
```
LighthouseKeeperYN, line 31: A switch would be more appropriate

packages/ui/src/composables/useSize.ts
```diff
@@ -59,6 +59,10 @@ export const useSize = (
   props: SizeProps,
   componentName?: string | undefined,
 ) => {
+  if (!componentName) {
+    componentName = getCurrentInstance()?.type.name
+  }
+
```
LighthouseKeeperYN, line 65:
```suggestion
  componentName?:  string = getCurrentInstance()?.type.name,
) => {
```

packages/ui/src/components/va-rating/components/VaRatingItem/VaRatingItem.vue
```diff
@@ -0,0 +1,124 @@
...
+
+    const onMouseMove = (ev: MouseEvent) => {
+      if (!props.hover) { return }
+      const { offsetX } = ev
+      const iconWidth = rootEl.value!.clientWidth
```
LighthouseKeeperYN, line 63: Linter forbids non-null assertions

packages/ui/src/components/va-rating/hooks/useVaRatingColors.ts
m0ksem: I think "if" makes this much easier :) Usually switch is harder to read on two cases...

packages/ui/src/components/va-rating/VaRating.vue
m0ksem: I have ts error here. Vue think even if there is default in prop definition, it can be undefined. This behavior caused because of some technically issue inside vue.

LighthouseKeeperYN (approved):

PR #1589, opened 2022-03-24
Check if color name is correct

rustem-nasyrov:
## Types of changes


m0ksem:
@asvae, @rustem-nasyrov 

100% sure this wouldn't work on SSR because it's a DOM method.

We want getColor to work on SSR, since it used in lots of components and on first request all colors must be correct. Better write somewhere in docs that we do not support named colors.

PR #1595, opened 2022-03-28
Fix/select value by with multiple

### LINKED ISSUE #1587: Doesn`t work value-by prop with multiple va-select element
**Vuestic-ui version:** 1.3.4

**Steps to reproduce**
Create va-select element with array of objects for options then make it multiple with value-by prop and specify what value from options object return
https://codesandbox.io/s/amazing-goldwasser-557ndl?file=/src/App.vue
Return array of selected specified object property values in value-by
Return array of selected options objects
**Other information**
For single select work as intendent

LighthouseKeeperYN:
close #1587

Multiple select with value-by prop now returns value-by value instead of entire object

## Types of changes


packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -309,8 +309,12 @@ export default defineComponent({
         return value
       },
 
-      set (value: any) {
-        emit('update:modelValue', getValue(value))
+      set (value: SelectableOption | SelectableOption[]) {
+        if (Array.isArray(value)) {
+          emit('update:modelValue', value.map(v => getValue(v)))
```
rustem-nasyrov, line 314: If `getValue` doesn't accept any other arguments, you can just provide it as callback.

RVitaly1978 (approved):

m0ksem (approved):
You can ignore my suggestions 👍. Would be nice to have refactoring and bug fix in separated PRs.

rustem-nasyrov (approved):

Issue #1479, opened 2022-02-03
Vuestic Docs website tweaks


xx13:
- [x] increase the default `p` font size to 1.2rem (not 1.1 rem)
- [x] `va-content h5` the first only of the h5 sequence: margin-top: 1.25 rem (not  …) 
- [x] `va-content h5` the first only of the h5 sequence:  margin-bottom: .75 rem (not …)
- [x] `va-content h5` other than the first one: margin-top: 4rem (not …)
- [x] `va-content h5` line-height: 1.25
- headings: for some reason we use h1, h3, h5, h6. can we use h1, h2, h3, h4 instead?

PR #1614, opened 2022-04-01
[#1479] vuestic docs website tweaks

aluarius:
## Description
close: #1479 

Seems, that these changes are needed only for docs website, not for va-content component, so, linked styles were just overwritten.

> for some reasons we use h1, h3, h5, h6. can we use h1, h2, h3, h4 instead.

It is possible, but more styles should be overwritten. I guest, this is not a todo item, but just a suggestion for this moment.

## Types of changes

### DIFF
--- a/packages/docs/src/components/DocsContent.vue
+++ b/packages/docs/src/components/DocsContent.vue
@@ -119,3 +119,19 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss">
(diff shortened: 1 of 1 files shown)


m0ksem (commented):
Let's have it on hold for nuxt docs.

rustem-nasyrov (approved):

RVitaly1978 (approved):

Issue #1596, opened 2022-03-28
Change text "Drag’n’drop files or" in va-file-upload


ghsgabriel:
**Is your feature request related to a problem? Please describe.**
no

**Describe the solution you'd like**
Add possibility to Change text "Drag'n'drop files or" in va-file-upload 

Thanks =D

PR #1615, opened 2022-04-01
 [#1596] custom drag n drop label text

aluarius:
close: #1596 

## Description
Added optional props for custom drag'n'drop label text.

## Types of changes


m0ksem (commented):
I think we also want to care about "Upload file" text here. Also, would be nice to have slots here, maybe. 

https://github.com/epicmaxco/vuestic-ui/blob/ec2c65ce510f285606f381573bbb63e4baa1c33e/packages/ui/src/components/va-file-upload/VaFileUpload.vue#L14-L25

aluarius:
> I think we also want to care about "Upload file" text here. Also, would be nice to have slots here, maybe.

Agree, slots looks fine here.

Derranion (approved):
Thanks,

Could you please add this to the docs description.

Because otherwise users will not know how to use it without checking source code.

[Here are some slots examples in our docs](https://vuestic.dev/en/ui-elements/select#slots).

packages/ui/src/components/va-file-upload/VaFileUpload.vue
```diff
@@ -9,7 +9,7 @@
         class="va-file-upload__field__text"
         v-if="dropzone"
       >
-        Drag’n’drop files or
+        <slot name="dropZoneLabel">Drag’n’drop files or</slot>
```
m0ksem, line 12: Having prop for label is more appropriate. We can use [Components Config]( https://vuestic.dev/en/services/components-config) for i18n for all components :) 

When I spoke about slot, I meant a slot that can be used to change overall file-upload visual (to replace default button and label text). Sorry for that. Let's better have props that can be changed in global config.

m0ksem (changes requested):

aluarius:
Added props for custom upload button & drop zone. Added slot for the whole upload area (no v-bind for events needed, cause there is hidden input, covering the whole area). Added examples and docs description.

m0ksem (approved):
Pretty nice now 😊

Derranion (approved):

PR #1616, opened 2022-04-02
Bring back modal exit animation

### LINKED ISSUE #1570: Modal exit animation is gone :<
**Vuestic-ui version:** 1.3.4

**Steps to reproduce**

Go to https://vuestic.dev/en/ui-elements/modal#base-layout

Seeing closing animation

Not seeing closing animation

**Other information**

Closing animation was there before, I'm pretty sure.

xiongmao86:
Animation disappear with v-if on teleport component at root that contains transition components inside. 

fixes #1570.

## Description
Because turning v-if to false the whole tree disappear without triggering subtree components' transitions. Thanks for the suggestion @m0ksem  give:
```
Teleport
  Transition
    .va-modal v-if="valueComputed"
      .va-modal-overlay
      .va-modal-container
```
I have try several approaches.

- I changes the structure first. But it didn't work, the enter and leave transitions simply disappear. 
- I have begin with vue example of modal, and try to turn it into the va-modal.vue. But with weird animation in the middle I can proceed.
- I have don't several other tweaks but none of them work out.

So, I decide to put the first approach to ask for review. This is the most straight forward approach, and I think is better for code review.

## Types of changes


xiongmao86:
It's strange with this approach that if I put `&-enter-from &__overlay, &-leave-to &__overlay: { opacity: 0; }` and change the transition duration to 13 second so that I can observe class assigning to transit element.

 I didn't see `va-modal-enter-from` in the same element with `va-modal` class and `v-if=valueComputed` element. But if I put `&-enter-from, &-leave-from { opacity: 0 }` then `va-modal-enter-from` `va-modal-enter-active` show up in the `.va-modal` element. And so is the case with leave classes.

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -295,9 +285,20 @@ export default class VaModal extends mixins(
...
+
+  &-enter-from &__container,
+  &-leave-to &__container {
+    opacity: 0;
+    transform: translateY(-30%);
+  }
+
+  &-enter-active &__container,
+  &-leave-active &__container {
+    transition: $modal-transform-transition, $modal-transform-transition;
```
m0ksem, line 301:
```suggestion
    transition: $modal-opacity-transition, $modal-transform-transition;
```

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -1,30 +1,20 @@
...
-        name="va-modal__container--transition"
-        :isTransition="!$props.withoutTransitions"
-        appear
-        @beforeEnter="onBeforeEnterTransition"
-        @afterEnter="onAfterEnterTransition"
-        @beforeLeave="onBeforeLeaveTransition"
-        @afterLeave="onAfterLeaveTransition"
-      >
         <div
           v-if="valueComputed"
```
m0ksem, line 11: Minor: Extra v-if, same as on `.va-modal`.

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -1,30 +1,20 @@
...
-  <teleport v-if='valueComputed' :to="attachElement" :disabled="disableAttachment">
-    <div class="va-modal">
-      <modal-element
-        name="va-modal__overlay--transition"
-        :isTransition="!$props.withoutTransitions"
-        @click="onOutsideClick"
-        appear
-      >
+  <teleport :to="attachElement" :disabled="disableAttachment">
+    <modal-element name="va-modal" :isTransition="!$props.withoutTransitions" appear>
```
m0ksem, line 3: As I can see here used nested transition. Vue listens for transition events (transitioned or animationend). Since there is no transition property on `.va-modal` vue can not determinate transition duration. 

So, it's required `Transition` child to have transition property set, or you can manually provide duration to transition component.

```suggestion
    <modal-element name="va-modal" :isTransition="!$props.withoutTransitions" appear :duration="300">
```

Then vue will set animation classes for specific duration and nested animations (`.va-modal__container`) will not ended in 0.00001 sec :)

m0ksem: Previously, animation duration was taken from `va-modal-transition` scss mixin. I think it's better to have animation duration prop 😁.

m0ksem (changes requested):
Good work overall here. Thanks for your effort. 

As I can see, VaModal component has been refactored to composition api, and we have a lot of conflicts here. 

I believe having animation in resources is some old approach. Thanks for removing @va-modal-transition scss mixin 👍 .

If you are interested, why do we need providing duration manually: https://vuejs.org/guide/built-ins/transition.html#css-based-transitions

xiongmao86:
> As I can see, VaModal component has been refactored to composition api, and we have a lot of conflicts here.

Sorry for the delay here, as I can't `yarn serve` in the latest `develop` commit, something wrong with the `VaDateInput.vue`, I have chose the working commit before I last update `develop` branch. I hope to deal with that issue maybe later. 

I'll carefully research the information you given later. It will take some time. Before that I think it's best to give a quick feedback here. Thank you very much for your help, @m0ksem.

xiongmao86:
I see how it works. Thanks for your help, @m0ksem. I think duration should work as  internal component `modal-element` so I guess we don't need to expose it outside. Are there anything left to do before getting this pr merged?

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -308,10 +299,16 @@ export default defineComponent({
     z-index: var(--va-modal-overlay-z-index);
     width: var(--va-modal-overlay-width);
     height: var(--va-modal-overlay-height);
+  }
 
-    &--transition {
-      @include va-modal-transition(true);
```
m0ksem, line 313: We can remove this old mixin from resources now 👍

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -282,10 +266,17 @@ export default defineComponent({
...
-    }
+  &-enter-from &__container,
+  &-leave-to &__container {
+    opacity: 0;
+    transform: translateY(-30%);
+  }
+
+  &-enter-active &__container,
+  &-leave-active &__container {
+    transition: $modal-opacity-transition, $modal-transform-transition;
```
m0ksem, line 279: SCSS variables are not included in final build. We use CSS variables, so user can easily redefine them. Please, store this values in _variables.scss. 

(You can see there are a lot of them used in components)

m0ksem (approved):
Pretty good 🤗. I found a few things to improve.

packages/ui/src/components/va-modal/VaModal.vue
xiongmao86: I think I removed that in the subsequent commit. I think I found it right after the request for review. I am sorry that I ask hastily before recheck anything left to do. I am prone to making that kind of mistake, just like washing disk and think I am finish and clean the dishcloth only to find there is one left undone in the hind of eye. I think I have to make it a routine check here.

xiongmao86:
@m0ksem, I have change scss variables into css variable. But the animation failed to work again. :<.

LighthouseKeeperYN (approved):

xiongmao86:
@m0ksem, are there anything that I missed to do?

xiongmao86:
I think you self-request a review because there are something I have to do. Never mind. Sorry for bordering.

m0ksem (approved):
@xiongmao86 Just wanted to check all one last time before merge. Thanks! 😋

Issue #1597, opened 2022-03-28
Check for not used files in the `page-configs`


Derranion:
Some of the currently existed example files (`src\page-configs\componentName\examples`) are not included in our `page-configs`.

For the instance:
[Mode.vue example of Date-input, and Advanced.vue of Color-input are not used in the related page-configs.](https://github.com/epicmaxco/vuestic-ui/pull/1579#issuecomment-1076309484)

Look for other similar cases and add them.

PR #1619, opened 2022-04-04
[#1597] outdated docs examples

aluarius:
close: #1597

## Description
All docs examples were checked. Mentioned files are not forgotten - they're outdated, linked to the components or props, that are not existing. For examle, va-color-input-advanced component, selected props for va-color-input or mode props for va-date-input.

## Types of changes


m0ksem:
Let's have this on hold. We will need to copy-paste these changes to nuxt docs later...

Derranion (changes requested):
These docs examples are not outdated.

For instance we have `mode` property in the `va-date-input` but this example is not used in the `page-config` of our docs page.

It will look this way, when we add them:

addDocsExamples


This means that we need to add these files to the related pages and add some description:
- [ ] Check if examples are working correctly.
- [ ] Add these examples to the page configs.

aluarius:
@Derranion thank you, i've understood about mode.vue. But what about other files? Not existing components va-color-input-advanced, va-color-palette-advanced, va-color-picker, va-color-slider.

For example, i see va-color-input-advanced call in VaColorInput.vdemo.vue, but that part of code commented since 2020.

Or select props, which is not existing both component code and docs (ve checked one more time).

rustem-nasyrov (approved):

RVitaly1978 (approved):

PR #1628, opened 2022-04-05
Added global config to create helper classes to style background and …

daurenSab:
…text color

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/services/color-config/color-css-helpers-updater.ts
+++ b/packages/ui/src/services/color-config/color-css-helpers-updater.ts
@@ -1,0 +1,54 @@
+import { watch } from 'vue'
+import { useGlobalConfig } from '../global-config/global-config'
+import { isServer } from '../../utils/ssr-utils'
+import { addOrUpdateStyleElement } from '../dom-functions'
+import type { ColorConfig } from './color-config'
+
+export type HelperConfig = {
+  stylePrefix: string;
+  styleProperty: string;
+}
+
+export const helperConfigValues: HelperConfig[] = [
+  {
+    stylePrefix: 'bg',
+    styleProperty: 'background-color',
+  },
+  {
+    stylePrefix: 'text',
+    styleProperty: 'color',
+  },
+]
+
+export const renderCSSHelpers = (helpers: ColorConfig) => {
+  const helperNames = Object.keys(helpers)
+  let resultHelperClasses = ''
+
+  helperNames.forEach((key) => {
+    helperConfigValues.forEach((helper: ColorConfig) => {
+      resultHelperClasses += `.${helper.stylePrefix}--${key} {${helper.styleProperty}: ${helpers[key]};}`
+    })
+  })
+
+  return resultHelperClasses
+}
+
+const ColorHelpersPlugin = {
+  install () {
+    if (isServer()) {
+      return
+    }
+
+    const { globalConfig } = useGlobalConfig()
+
+    watch(() => globalConfig.value.classHelpers, (newValue) => {
+      if (!newValue) {
+        return
+      }
+
+      addOrUpdateStyleElement('css-helpers', () => renderCSSHelpers(newValue))
+    }, { immediate: true, deep: true })
+  },
+}
+
+export default ColorHelpersPlugin
--- a/packages/ui/src/services/global-config/types.ts
+++ b/packages/ui/src/services/global-config/types.ts
@@ -5,6 +5,7 @@ import type { IconConfig } from '../icon-config/types'
 export type GlobalConfig = {
   colors?: ColorConfig,
   icons?: IconConfig,
+  classHelpers?: ColorConfig,
   components?: ComponentConfig
   componentsAll?: Props
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10184 @@
+{
+  "name": "vuestic-ui-wrapper",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@achrinza/node-ipc": {
+      "version": "9.2.2",
+      "resolved": "https://registry.npmjs.org/@achrinza/node-ipc/-/node-ipc-9.2.2.tgz",
+      "integrity": "sha512-b90U39dx0cU6emsOvy5hxU4ApNXnE3+Tuo8XQZfiKTGelDwpMwBVgBP7QX6dGTcJgu/miyJuNJ/2naFBliNWEw==",
+      "dev": true,
+      "requires": {
+        "@node-ipc/js-queue": "2.0.3",
+        "event-pubsub": "4.3.0",
+        "js-message": "1.0.7"
+      }
+    },
+    "@babel/code-frame": {
+      "version": "7.16.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
+      "integrity": "sha512-iAXqUn8IIeBTNd72xsFlgaXHkMBMt6y4HJp1tIaK465CWLT/fG1aqB7ykr95gHHmlBdGbFeWWfyB4NJJ0nmeIg==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.16.7"
+      }
+    },
+    "@babel/helper-validator-identifier": {
+      "version": "7.16.7",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.16.7.tgz",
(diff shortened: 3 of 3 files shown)


packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const ColorHelpersPlugin = {
+  install () {
+    if (isServer()) {
+      return
+    }
+
+    const { globalConfig } = useGlobalConfig()
+
+    watch(() => globalConfig.value.classHelpers, (newValue) => {
```
m0ksem, line 44: Watch for colors to change. 

For example, user can switch between dark and light theme, then only colors will be updated in globalConfig.

```suggestion
    watch(() => globalConfig.value.colors, (newValue) => {
```

no need for additional ColorConfig in GlobalConfig with different name :)

packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,54 @@
...
+import { isServer } from '../../utils/ssr-utils'
+import { addOrUpdateStyleElement } from '../dom-functions'
+import type { ColorConfig } from './color-config'
+
+export type HelperConfig = {
+  stylePrefix: string;
+  styleProperty: string;
+}
+
+export const helperConfigValues: HelperConfig[] = [
```
m0ksem, line 12: This thing looks pretty useful! 😊

This can be actually moved to GlobalConfig or different plugin as optional service later. 

Can we have some fabric of CSS Helpers? I mean, can user make my own helpers, like `.botder-bottom--primary`?

---


`text--${color}` I think also must care about `fill` CSS property - this will make SVG icons change their colors as well.

packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const { globalConfig } = useGlobalConfig()
+
+    watch(() => globalConfig.value.classHelpers, (newValue) => {
+      if (!newValue) {
+        return
+      }
+
+      addOrUpdateStyleElement('css-helpers', () => renderCSSHelpers(newValue))
```
m0ksem, line 49:
```suggestion
      addOrUpdateStyleElement('va-css-helpers', () => renderCSSHelpers(newValue))
```

m0ksem: Considering that starting from Vue3 user can have many vue apps on same page, maybe we need to add vue app uid as suffix. But, let's think about it later in a different task.

packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+]
+
+export const renderCSSHelpers = (helpers: ColorConfig) => {
+  const helperNames = Object.keys(helpers)
+  let resultHelperClasses = ''
+
+  helperNames.forEach((key) => {
+    helperConfigValues.forEach((helper: ColorConfig) => {
+      resultHelperClasses += `.${helper.stylePrefix}--${key} {${helper.styleProperty}: ${helpers[key]};}`
```
m0ksem, line 29:
```suggestion
      resultHelperClasses += `.va-${helper.stylePrefix}--${key} {${helper.styleProperty}: ${helpers[key]};}`
```

User can use Vuestic along with Tailwind or another CSS framework.

packages/ui/src/services/global-config/types.ts
```diff
@@ -5,6 +5,7 @@ import type { IconConfig } from '../icon-config/types'
 export type GlobalConfig = {
   colors?: ColorConfig,
   icons?: IconConfig,
+  classHelpers?: ColorConfig,
```
m0ksem, line 8: We already have ColorConfig here. 

Better have here config with class and properties - `helperConfigValues` and some settings, for example `classesPrefix`.

Make sure `helperConfigValues` will have a default value and simpler name 😁

m0ksem (changes requested):
Good work. Let's make this feature more customizable.

packages/ui/src/services/color-config/color-css-helpers-updater.ts
daurenSab: Yeah, the idea of this type was to provide some functionality to create own utlilty first classes.

packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export let helperConfigValues: HelperConfig[] = [
+  {
+    stylePrefix: 'bg',
+    styleProperty: 'background-color',
+  },
+  {
+    stylePrefix: 'text',
+    styleProperty: 'color',
```
m0ksem, line 20: I think color--primary should use both color and fill.
Maybe change it to styleProperty: ['color', 'fill']

So it should have MaybeArray<string> type.

packages/ui/src/services/color-config/color-css-helpers-updater.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+  {
+    stylePrefix: 'fill',
+    styleProperty: 'fill',
+  },
+]
+
+const renderHelperPresets = (newColors: any) => {
+  return Object.entries(newColors).map(color => {
```
m0ksem, line 29: Better destruct color into variables 

[coloName, colorValue] => {

packages/ui/src/vue-book/book-main.ts
```diff
@@ -34,6 +35,7 @@ const router = createRouter({
 })
 
 app.use(VueBookComponents)
+app.use(ColorHelpersPlugin)
```
m0ksem, line 38: This should be in vuestic-plugins.ts

m0ksem (changes requested):
Good work. Let's make it a bit better.

packages/ui/src/vuestic-plugin/vuestic-plugin-without-components.ts
```diff
@@ -1,6 +1,6 @@
 import type { GlobalConfig } from '../main'
 import type { App } from 'vue'
-import { DropdownPopperSubplugin, ToastInstall, GlobalConfigPlugin, ColorHelpersPlugin } from './vuestic-plugins'
+import { DropdownPopperSubplugin, ToastInstall, GlobalConfigPlugin, ColorHelpersPlugin, ClassHelpersPlugin } from './vuestic-plugins'
 
 export const VuesticPluginsWithoutComponents = {
   install (app: App, vuesticConfig: GlobalConfig): void {
```
m0ksem, line 6: If you made color as required prop, make sure we don't force user to provide whole config in plugin and `mergeGlobalConfig`.  (`setGlobalConfig` likely must request all required fields.)

```suggestion
  install (app: App, vuesticConfig: Partial<GlobalConfig>): void {
```

packages/ui/src/vue-book/book-main.ts
```diff
@@ -45,6 +45,12 @@ app.use(VuesticPluginsWithoutComponents, {
     ...colorsPresets.default,
     banana: '#d0f55d',
   },
+  classHelpers: [
```
m0ksem, line 48: You can create color-css-helpers-updater.demo.vue and have simple demo here. It would be helpful for manual testing (and cypress later).
```vue
<div class="demo--banana"> I'm yellow </div>
```

m0ksem (approved):
Good. Let's move to docs.

Feel free to make interesting demos like:
https://vuestic.dev/en/services/icons-config#font-name-pattern
https://vuestic.dev/en/services/colors-config

packages/ui/src/services/global-config/global-config.ts
```diff
@@ -8,6 +8,7 @@ import { colorsPresets } from '../color-config/color-theme-presets'
 
 const globalConfigRef = ref<GlobalConfig>({
   colors: colorsPresets.default,
+  classHelpers: [],
```
RVitaly1978, line 11: isn't `helperConfigValues` (from `packages/ui/src/services/color-config/color-css-helpers-updater.ts`) an initial value?

RVitaly1978 (changes requested):

packages/ui/src/services/color-config/ColorCSSHelpersUpdater.demo.vue
```diff
@@ -0,0 +1,99 @@
...
+        <p class="va-text--success">
+          This is success color
+        </p>
+        <p class="va-text--warning">
+          This is warning color
+        </p>
+        <p class="va-text--primary">
+          This is primary color
+        </p>
+        <p class="va-demo--gray">
```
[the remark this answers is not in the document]
m0ksem: Here we test if custom rules work.

https://github.com/epicmaxco/vuestic-ui/pull/1628/files#diff-38502ba5906cc044b1a35c5bb4473e58e7fcee3c3a7a4e541f0908a6615810bb

Issue #1554, opened 2022-03-11
VaInput add text-align prop


Derranion:
Having ability to align text out of the box is really useful.

Right now text inside of our `VaInput` is aligned by default.

The only way to change it for the user is to use deep selectors, this could be improved by adding new prop `text-align`.

asvae:
Maybe css variable? prop seems like an overkill.

Derranion:
Task has been updated.

m0ksem:
> @Derranion wanted to solve this issue with `input-class`, which i found in VaInput.demo.vue:
> 
> https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-input/VaInput.demo.vue#L568-L584
> 
> But it seems, that it's not working:
> 
> http://joxi.ru/krDavbahGJKdqm.jpg
> 
> Is it bug or just an old and removed functional?

It's a bug.

PR #1629, opened 2022-04-05
[#1554] input class fix & build-in input text aligment classes

aluarius:
## Description
- Fixed проблем with input-class prop - was't binded to the input.
- Added build-in classes for input text aligment.
- Docs updated.

In my opinion, considering input class ressurection, build-in classes don't looks necessary.

## Types of changes


packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -244,3 +246,15 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss" scoped>
```
RVitaly1978, line 250:
```suggestion
<style lang="scss">
```
we don't want to use `scoped` styles in components

aluarius: Changed to global classes.

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -60,12 +60,14 @@
...
       v-bind="{ ...computedChildAttributes, ...textareaProps, ...inputEvents }"
       class="va-input__content__input"
+      :class="inputClass"
     />
 
     <input
       v-else-if="!$slots.content"
       ref="input"
       class="va-input__content__input"
+      :class="inputClass"
```
m0ksem, line 70: This can be in `computedChildAttributes`

m0ksem (approved):

LighthouseKeeperYN (approved):

Issue #1411, opened 2022-01-13
Input wrapper resize if it has more than 2 icons.


m0ksem:
**Vuestic-ui version:** 1.3.3

**Steps to reproduce**
![cUCDxRd9iG](https://user-images.githubusercontent.com/23530004/149334506-b55bef52-72f0-477d-ab23-691ffdd08478.gif)

**What is the expected behavior?**
Input size mustn't change.

PR #1630, opened 2022-04-05
[#1411] input wrapper resize fix

aluarius:
close: #1411 

## Description
CSS fill available fixes this problem. Input will grow-shrink depending on free space and component's width would not be affected by amount of active icons.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/components/VaInputWrapper.vue
+++ b/packages/ui/src/components/va-input/components/VaInputWrapper.vue
@@ -307,8 +307,11 @@ export default defineComponent({
 
   &--labeled {
     .va-input__content-wrapper {
-      padding-top: 12px;
       height: 100%;
+      width: -moz-available;
+      width: -webkit-fill-available;
+      width: fill-available;
+      padding-top: 12px;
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -307,8 +307,11 @@ export default defineComponent({
 
   &--labeled {
```
RVitaly1978, line 308: not only for `--labeled`

---

But it seems that now the default `va-input` field `width` is `100%` and there is no more this problem

aluarius: > not only for `--labeled`
> 
> But it seems that now the default `va-input` field `width` is `100%` and there is no more this problem

Yeah, didn't mention. Moved to common input content wrapper class.

Problem was reproduced by me yesterday (honestly, can't remember how exactly), moreover, width 100% was put there in 2021, but this is 2022 issue.

RVitaly1978 (approved):

PR #1634, opened 2022-04-06
Remove withTransportConfig wrapper

### LINKED ISSUE #1297: Fix withTransportConfig setup expose
Expose function expose component properties only to instance, so there is no way to check exposed keys before component is rendered. 

Because of that, `withTransportConfig` can not determinate which methods are should be proxied from setup function.

There are few options that I can see:
- ✔ We should remove this proxy component and create some `transportConfig` hook. (This will require us to refactor all the components.)
- ✔ We should add `expose` component as option (like `emits`), so in `withTransportConfig` we can see exactly which methods should be proxied.  
- 🤔 Maybe we can see exposed methods after component is rendered and update methods on runtime.
- 🤔 Maybe we can modify props or propsData in withTransportConfig and use mixins here.

### LINKED ISSUE #1581: With config transport remake
That's pretty important.

Currently we have a wrapper for component, that connects it to config. It creates a number of problems. We need a simpler method for that.

Needs an in-depth investigation as impact area is pretty large.

In practical terms right now it
* complicates event handling (as events have to pass through wrapper)
* and vue dev-tools show 2 components instead of one.

Pretty sure we had some other complications as well.

### LINKED ISSUE #1631: `va-checkbox` doesn't work after build
**Vuestic-ui version:** 1.3.5

**Steps to reproduce**
Create `data-table` with `selectable` `prop` set to `true`

Selection will stop working after the build:

dataTableSelectionBuildIssue

**Other information**
[Initial community report from discord](https://discord.com/channels/457593300799717377/771405157527191585/960237163532914759).

m0ksem:
closes #1297
closes #1631
closes #1581

### General 
Patch instance props with vue reactive proxy. I looked in vue source code, Setup called, where we can access instance and patch it. Props in render function used from instance, so render function uses patched props. Same for DevTools. 
I believe using proxy is a nice approach, because we don't modify any object. 

### List of changes
- Patch vue instance props with Proxy that can return props from Vuestic Config,
- Patch class-component props
- Patch Options API component (added setup for patching props).

### Test
- No unit tests until review and manual tests.
- Tested in build (can't see any difference). 
- Demo link: http://localhost:8080/#/demo/va-config/VaConfig.demo.vue

### Check this PR
- Vite build/serve 🤔
- Webpack serve/serve:productin/build ✔
- SSR Vite 🤔 - the only one difference I found from vue source code is that props are not reactive in SSR build. Shouldn't be a big difference with SPA builds.
- SSR Nuxt 😋 (later)


packages/ui/src/services/config-transport/createProxyComponent.ts
```diff
@@ -0,0 +1,66 @@
+import { getCurrentInstance, ComponentInternalInstance, DefineComponent, SetupContext, Ref } from 'vue'
+import { useComponentConfigProps } from '../component-config/component-config'
+import { computedShallowReactive } from './computedShallowReactive'
+
+/** Compiled and reactive props. By default they passed to setup fn */
+type Props = Record<string, unknown>;
```
RVitaly1978, line 6: Is this type somehow related to the type `Props` from the file `../component-config/component-config.ts`?

m0ksem: This type actually must be:
```suggestion
type Props = Parameters<NonNullable<DefineComponent<Record<string, unknown>>['setup']>>[0];
```

But it's a very redundant 😅.

Derranion:
VaAffix doesn't work.
VaAffixError

Derranion:
`VaConfig` -> `Global config` works while using dev variables (`yarn serve`)
But does not work in the production mode (`yarn serve:production`) same behavior we can expect after the build.

Steps to reproduce:

1. Press "Should set dark red color on click" button in the `VaConfig`
2. "Must change" should be red after this in production

Production mode:
VaConfigProductionMode

Expected:
VaConfigExpected

Derranion:
`VaButton` styles do not match in `dev` and `production` modes.

Not so sure about this one - it should work as expected after the build but still looks a bit weird from the developers side that they are different.

Dev mode:
ButtonStylesInDevMode

Production mode:
ButtonStylesDoNotMatchInProductionMode

Derranion:
`VaToast` is broken - does not show any toasts.

Derranion:
`VaDateInput` text displays function body instead of expected visual representation of the date.
Displayed function seems to not be executed.

DateInputTextIsBroken

Derranion:
`VaConfig` `local config` does not update after change and uses `default` values instead.

Steps to repdoduce:

1. Change `config color` to any value.

localConfigUpdateIssue

Derranion:
Please tell if you need some help or busy to update this.

m0ksem:
@Derranion, pretty funny thing! 😁Nice catch.

withTransportConfigWrapper has resolved Function type as Object, But in vue docs I can see that Function type default value is resolved as Function, not like object fabric.

https://vuejs.org/guide/components/props.html#prop-validation

```ts
  // Function with a default value
  propG: {
    type: Function,
    // Unlike object or array default, this is not a factory function - this is a function to serve as a default value
    default() {
      return 'Default function'
    }
  }
```

I will change all components with `() => () => result` default value for Function props.

m0ksem:
> `VaDateInput` text displays function body instead of expected visual representation of the date. Displayed function seems to not be executed.
> 
> DateInputTextIsBroken

Marked all other comments with 🚀 reaction as resolved.

Derranion (commented):
Usage of incorrect prop value inside of the `Local config` breaks component completely and makes it disappear.

Steps to reproduce:

1. Open `VaConfig` demo
2. In the `local config` color input field remove one symbol
3. Hover `VaButton` component.

localConfigInvalidValueBreaskComponent

Derranion (commented):
> @Derranion, pretty funny thing! 😁Nice catch.
> 
> withTransportConfigWrapper has resolved Function type as Object, But in vue docs I can see that Function type default value is resolved as Function, not like object fabric.
> 
> https://vuejs.org/guide/components/props.html#prop-validation
> 
> ```ts
>   // Function with a default value
>   propG: {
>     type: Function,
>     // Unlike object or array default, this is not a factory function - this is a function to serve as a default value
>     default() {
>       return 'Default function'
>     }
>   }
> ```
> 
> I will change all components with `() => () => result` default value for Function props to `() => result`.


Thanks, I've found only one such case in the `VaDateInput` and updated it.

Default values for other `props` with `Function` `type` are either not set or set as `undefined`.

m0ksem:
> Usage of incorrect prop value inside of the `Local config` breaks component completely and makes it disappear.
> 
> Steps to reproduce:
> 
>     1. Open `VaConfig` demo
> 
>     2. In the `local config` color input field remove one symbol
> 
>     3. Hover `VaButton` component.
> 
> 
> localConfigInvalidValueBreaskComponent

It's correct behavior 😋, 
First, it wasn't disappearing on hover - #ff00 is cast to white color for some reason.  
Having error during rendering process will stop component rendering. This issue is present even in develop.

packages/ui/src/services/component-config/component-config.ts
```diff
@@ -1,3 +1,26 @@
+import { useLocalConfig } from '../../components/va-config/VaConfig'
+import { useGlobalConfig } from '../global-config/global-config'
+import { computed, DefineComponent, ref } from 'vue'
+
 // TODO ideally we want to type component props instead  of any
```
m0ksem, line 5: We don't need this comment anymore, because now withTransportConfig return component type instead of any.

Derranion:
> > Usage of incorrect prop value inside of the `Local config` breaks component completely and makes it disappear.
> > Steps to reproduce:
> > ```
> > 1. Open `VaConfig` demo
> > 
> > 2. In the `local config` color input field remove one symbol
> > 
> > 3. Hover `VaButton` component.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > localConfigInvalidValueBreaskComponent
> 
> It's correct behavior 😋, First, it wasn't disappearing on hover - #ff00 is cast to white color for some reason. Having error during rendering process will stop component rendering. This issue is present even in develop.

Fallback to white instead of the default value doesn't seem like correct behavior.

packages/ui/src/components/va-affix/VaAffix.vue
```diff
@@ -18,6 +18,7 @@
 <script lang="ts">
```
Derranion, line 18: https://github.com/epicmaxco/vuestic-ui/pull/1634#issuecomment-1096474275

Fallback to white instead of `default` value doesn't seem like correct behavior.

m0ksem: 4 digit CSS color is RGBA, so in `#ff00` the last 0 is alpha.

`#ff00` is the same as `#ffff0000`.

I was wrong about white color, it is just transparent color.

Try `#ff09`

Derranion: It was transparent because "0" is getting parsed as alpha in this color `#ff00`.

Should be fine. Though we could throw a warning instead of the error, but this is not that important.

m0ksem: We throw it only for 5 digit values.

packages/ui/src/components/va-affix/VaAffix.vue
```diff
@@ -27,16 +28,18 @@ import {
...
   emits: ['change'],
   props: {
     offsetTop: { type: Number as PropType<number>, default: undefined },
     offsetBottom: { type: Number as PropType<number>, default: undefined },
-    target: { type: Function as PropType<() => HTMLElement | Window>, default: () => () => window },
+    target: { type: [Object, Function] as PropType<Target | (() => Target)>, default: getWindow },
   },
   setup (props, { emit }) {
-    const getTargetElement = () => props.target()
+    const getTargetElement = () => (typeof props.target === 'function' ? props.target() : props.target)
```
LighthouseKeeperYN, line 42: Can this be computed?

packages/ui/src/utils/ssr-utils.ts
```diff
@@ -1,2 +1,4 @@
 export const isServer = () => typeof window === 'undefined'
 export const isClient = () => !isServer()
+export const getWindow = (): Window | undefined => typeof window === 'undefined' ? undefined : window
+export const getDocument = (): Document | undefined => typeof document === 'undefined' ? undefined : document
```
LighthouseKeeperYN, line 4: What does it do? Isn't it the same as calling window and document directly and either getting them or undefined?

m0ksem: In SSR you can not access window. It is part of browser's DOM. So you need first check if it defined, then access. We care about it in codes that run before components mounted.

packages/ui/src/components/va-affix/VaAffix.vue
m0ksem: Ye. But I think it should be another PR. I just fixed it so it work with new withTransportConfig.

packages/ui/src/components/va-affix/VaAffix.vue
Derranion, line 18: `VaButtonToggle` `color` and `border-color` falls back to white color after selection.

ButtonToggleWhiteDefault

LighthouseKeeperYN (approved):

packages/ui/src/components/va-affix/VaAffix.vue
```diff
@@ -18,6 +18,7 @@
 <script lang="ts">
 import { defineComponent, computed, PropType, ref, Ref, nextTick, onMounted, onBeforeUnmount } from 'vue'
```
Derranion, line 19: This issue is resolved in the `dev` mode but still exists in the `production` mode.

Steps to reproduce:

1. Run yarn serve:production
2. Press "Should set dark red color on click" button in the `VaConfig`



> `VaConfig` -> `Global config` works while using dev variables (`yarn serve`) But does not work in the production mode (`yarn serve:production`) same behavior we can expect after the build.
> 
> Steps to reproduce:
> 
> 1. Press "Should set dark red color on click" button in the `VaConfig`
> 2. "Must change" should be red after this in production
> 
> Production mode: VaConfigProductionMode
> 
> Expected: VaConfigExpected

Derranion: Update:
I've noticed that "Must change color" is our `va-chip` `component`, for `va-button` it works correctly.

Tested the same thing on the build version of the project and it works the same.

Derranion: I think this happens because `va-chip` uses `VueClassComponent` inside unlike `va-button` and this case is not handled in the new logic - we have fixed similar cases in the previous version of `WithConfigTransport`.

m0ksem: So we wait until VaChip will be refactored. I don't want to patch vue-class-component instance... It's made of getters 😭

Derranion: I can do this but looks like Oleg started it already.

asvae (approved):
Didn't test fully, but conceptual side looks excellent 👍 .

packages/ui/src/components/va-affix/VaAffix.vue
m0ksem: Fixed in [eacc354](https://github.com/epicmaxco/vuestic-ui/pull/1634/commits/eacc354057588314777b4ed8fb17fdbebe52796b)

packages/ui/src/components/va-affix/VaAffix.vue
m0ksem: Fixed in https://github.com/epicmaxco/vuestic-ui/pull/1668

I rebased this branch with current develop. Works well in `yarn serve:production` and `yarn build && serve ./dist`.

packages/ui/src/components/va-affix/VaAffix.vue
Derranion: I guess link was removed after force push.

Derranion (approved):
Amazing,

Just a remainder:
Some components that use `VueClassComponent` will not get values from configs but this is a minority and we will fix it soon.

Well done.

Issue #1290, opened 2021-12-01
Optimize VaList, VaSelect long list etc.


m0ksem:
![xMCqPDxciw](https://user-images.githubusercontent.com/23530004/144199878-c1b0bfb2-7291-46e7-9637-ed3b139da73b.gif)

I have ~400 options in this select, and it takes a long time to render.

It can be also solved if we add slot for `va-select-options-list`, so user can change this list and optimize it by himself.

----

Changing `v-if` to `v-show`  in va-dropdown here can be helpful...

m0ksem:
Also, additional slots here can be helpful. In my example, I would like to add additional “Search Filed” with custom search logic.

asvae:
Discussed:
* We can document existing approach (see http://vuestic-ui-demo.sub.asva.by/#/demo/va-select/VaSelect.demo.vue scroll-bottom event).

Potentially we might want to implement virtual scroll, but not right now.

aluarius:
Moved to https://github.com/epicmaxco/vuestic-ui/issues/2290

PR #1635, opened 2022-04-06
[#1290] va select optimization and options slot

aluarius:
close: #1290

## Description
1. Added new bool prop - `longList`. It changes options list rendering from v-if to v-show (no more friezes). Why prop? Because if page will contain lots of select components, it will affect whole page performance.
2. Added new slot for custom options list and related ability to use custom search filtering logic.

Tell your thoughts and then i'll update examples and docs.

## Types of changes

### DIFF
--- a/packages/docs/src/components/DocsContent.vue
+++ b/packages/docs/src/components/DocsContent.vue
@@ -119,3 +119,19 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss">
+.va-content p {
+  font-size: 1.2rem;
+}
+
+.va-content h5 {
+  margin-top: 4rem;
+  line-height: 1.25;
+
+  &:first-of-type {
+    margin-top: 1.25rem;
+    margin-bottom: 0.75rem;
+  }
+}
+</style>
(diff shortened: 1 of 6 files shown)


m0ksem:
How about computed based on options length instead of prop?

aluarius:
> How about computed based on options length instead of prop?

Not sure. We may have friezes because of 1000 textual options or of 100 complex (title+icon+description, for example) as well. And if user will have lags, but have no way to solve them by build-in functional, thats seems not good.

Derranion:
Yeah, I also believe that we want to optimize `options` out of the box.

We could apply `v-show` optimization at a lower `options.length` or just use `v-show` all the time.
To not go deeper into all of this.

We can discuss this tomorrow.

Derranion (commented):
Yeah, I also believe that we want to optimize `options` out of the box.

We could apply `v-show` optimization at a lower `options.length` or just use `v-show` all the time.
To not go deeper into all of this.

We can discuss this tomorrow.

aluarius:
I guess this pr make no sense since we've decided to implement virtual scroll. We'll return to this issue after it.

PR #1659, opened 2022-04-14
Feat/improve validators

### LINKED ISSUE #1400: Form input control can close  autaomatic validation, so that we can validate form when required

I checked vuestic doc [https://vuestic.dev/](url),  and didn't found way to disable automatcic validation on form control,  but under some requirement, we don't want to trigger validation automactilly,  and we cannot specify control to trigger validation with specific event.

By currently desgin,   I only found  way to specify  validation state [https://vuestic.dev/en/ui-elements/input#validation-and-error-success-state](url)  manualy,  but with this way we must put validation logic on seperate part which out input control self, not very well.

So sugguest have following ways to control validation with accurately:
- Have way to specify when to trigger validation, like `click` `change`
- Have a property like `automaticValidate`  wiht defualt value `true`

### LINKED ISSUE #1491: Form controls should support server validation
**Vuestic-ui version:** 1.3.4

In the current version, the form control only has `rules` for verification,  but in the most real scenario, we need a request to the server to check whether already exists, or check it whether is valid. 

and under the current version, it's hard to integrate the server validation.  we can use `error-messages` and `change` to achieve the server validation goal but not smoothly and prop `rules` will be out of work.

**Other information**

rustem-nasyrov:
## Description
- [x] Added support for the async or promise rules validations.
- [x] Supported inner loading state for the form controls/input components. 

## Also need to add validation to:
- [x] ✅ File upload component.
- [x] ✅ Radio component (need to discuss).
- [x] ✅ Color input component ~~(need to discuss)~~.

## Types of changes

Closes #1400
Closes #1491.


packages/ui/src/composables/useValidation.ts
```diff
@@ -32,6 +32,8 @@ export const useValidationProps = {
 
 export const useValidationEmits = ['update:error', 'update:errorMessages']
 
+export type validationResult = boolean | string | Promise<any>
```
m0ksem, line 35: Replace `any` with `string | boolean`

packages/ui/src/composables/useValidation.ts
```diff
@@ -66,12 +69,28 @@ export const useValidation = (
...
         if (isString(validationResult)) {
           errorMessages = [...errorMessages, validationResult]
           error = true
         } else if (validationResult === false) {
           error = true
+        } else if (validationResult instanceof Promise) {
+          innerLoadingState.value = true
+
+          validationResult
+            .then(promisedValidationResult => {
```
m0ksem, line 82: You forget to handle boolean result here. Let's have some recursive handle function here.

packages/ui/src/composables/useValidation.ts
```diff
@@ -66,12 +69,28 @@ export const useValidation = (
...
+              }
+            })
+            .catch(e => { throw new Error(e) })
+            .finally(() => {
+              computedErrorMessages.value = [...errorMessages]
+              computedError.value = error
+              innerLoadingState.value = false
+            })
         }
       })
```
m0ksem, line 96: Need to deal with `Promise.all`, looks like !error will be returned before promises resolved.

m0ksem: Also, would be nice to have dirty state, to prevent extra validation requests. Remember that we validate a whole form with a lot of fields.

m0ksem (changes requested):

packages/ui/src/composables/useValidation.ts
```diff
@@ -60,20 +64,37 @@ export const useValidation = (
...
+        promises.push(validationResult)
+      }
+    }
+
+    normalizeValidationRules(rules, props.modelValue).forEach(validationResultHandler)
+
+    if (promises.length) {
+      Promise.all(promises)
+        .then(promisedRules => promisedRules.forEach(validationResultHandler))
+        .catch(e => { throw new Error(e) })
```
LighthouseKeeperYN, line 91: This catch looks redundant

LighthouseKeeperYN (approved):

packages/ui/src/composables/useValidation.ts
rustem-nasyrov: I wrote it here if the back-end validation returns error: network error, server error, etc.

LighthouseKeeperYN: Yeah but then you throw that same error anyway. Essentially this catch replicates default behavior.

rustem-nasyrov:
@asvae, @Derranion, @m0ksem can you review please? :)

packages/ui/src/composables/useValidation.ts
```diff
@@ -60,20 +63,36 @@ export const useValidation = (
...
+        innerLoadingState.value = true
+
+        promises.push(validationResult)
+      }
+    }
+
+    normalizeValidationRules(rules, props.modelValue).forEach(validationResultHandler)
+
+    if (promises.length) {
+      Promise.all(promises)
```
m0ksem, line 88: How about returning this Promise?

m0ksem: Or maybe even make separate method validateAsync.

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -46,7 +46,7 @@
         @click.stop="reset()"
       />
       <va-icon
-        v-if="loading"
+        v-if="loading || isInnerLoading"
```
m0ksem, line 49: This can be `loadingComputed` from `useValidation` hook and used across all components.

m0ksem (changes requested):
It looks well in scope of components. 
But in scope of VaForm and manual validation this will not work.

First, I, as user, unable to manually trigger validation and get result if my VaInput is valid. We return validation result before promise is fulfilled. 

What I can suggest: make two different function validate and validateAsync. Then user and VaForm can deal with async validation. 

We also need some demo here. And some docs later.

packages/ui/src/components/va-input/VaInput.vue
rustem-nasyrov: @m0ksem got it, I'll make separated validation functions.

rustem-nasyrov:
I close the PR and will start doing the issue from scratch again later.

Issue #1639, opened 2022-04-07
Problems with components in Nuxt 3


aluarius:
Problems with components in Nuxt 3.

- [ ] icons are not loading (everywhere, for example - `VaCheckBox`, `VaCollapse`)
- [ ] accordion in not working properly (nested content for active element has height: 0px)

Example: `VaAccordion` + `VaCollapse`.

![](http://joxi.ru/n2YpwlpckZlDEm.jpg)

- [ ] document is not defined in `VaButtonDropdown`
- [ ] document is not defined in `VaBacktop`

![](http://joxi.ru/Vm6jNGjH3jJPWm.jpg)

- [ ] hover is't working in `VaButtonGroup` (works for `VaButton`)
- [ ] VaColorIndicator after page reload "loosing" props state - square, for example (may be not only this component)

...to be continued

m0ksem:
@aluarius, check out this issue. https://github.com/epicmaxco/vuestic-ui/issues/1577

aluarius:
Related with https://github.com/epicmaxco/vuestic-ui/issues/1641

aluarius:
The last item (docs update) is waiting for nuxt rework.

aluarius:
Closed, cause docs update aren't needed anymore.

PR #1661, opened 2022-04-14
[#1639] va backtop, va switch, va color indicator nuxt fix

aluarius:
Related to #1639 (not closing, cause of item about docs, which is waiting for nuxt).

## Description
- Fixed error in Nuxt 3 with `VaColorIndicator` (`square` style disappears after page reload). It seems, that 0 as Number somehow rerenders as if it was Boolean. Changed to the String.
- Fixed hydration node mismatch with `VaSwitch` in Nuxt 3. Problem was solved by conditional rendering of labels, if there're no passed content slots or props for them.
- `VaBackTop`. Fixed Nuxt 3 error with undefined document (target is now binding after component was mounted). Default target was changed to `window` cause `document.body`, that was used before, always has `scrollTop` === 0. Window doesn't have scrollTop at all, so, in case of default target, will be used `scrollTo` method (and `scrollY` attribute for visibility behaviour as well) with smooth behivor (can't control speed, should mention this in docs).

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-backtop/VaBacktop.vue
+++ b/packages/ui/src/components/va-backtop/VaBacktop.vue
@@ -3,7 +3,7 @@
     v-if="visible"
     class="va-backtop"
     :style="computedStyle"
-    @click="scrollToTop()"
+    @click="scrollToTop"
   >
     <slot>
       <va-button
@@ -15,8 +15,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, computed, PropType, ref, onMounted, onBeforeMount, Ref } from 'vue'
-
+import { defineComponent, PropType, ref, computed, onMounted, onBeforeUnmount } from 'vue'
 import VaButton from '../va-button'
 
 export default defineComponent({
@@ -24,8 +23,8 @@ export default defineComponent({
   components: { VaButton },
   props: {
     target: {
-      type: [Object, String] as PropType<Element | string>,
-      default: () => window as any as Element,
+      type: [Object, String] as PropType<Element | string | null>,
+      default: null,
     },
 
     visibilityHeight: { type: Number as PropType<number>, default: 300 },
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/va-backtop/VaBacktop.vue
```diff
@@ -15,17 +15,16 @@
...
 
 export default defineComponent({
   name: 'VaBacktop',
   components: { VaButton },
   props: {
     target: {
-      type: [Object, String] as PropType<Element | string>,
-      default: () => window as any as Element,
+      type: [Object, String] as PropType<Element | string | null>,
+      default: null,
```
m0ksem, line 27:
```suggestion
      default: undefined,
```

packages/ui/src/components/va-backtop/VaBacktop.vue
```diff
@@ -65,27 +71,41 @@ export default defineComponent({
...
+      const targetScrollValue = targetElement instanceof Window
+        ? targetElement.scrollY
+        : targetElement.scrollTop
+      visible.value = targetScrollValue > props.visibilityHeight
     }
 
     onMounted(() => {
       targetElement = getTargetElement()
-      targetElement?.addEventListener('scroll', handleScroll)
+      targetElement.addEventListener('scroll', handleScroll, true)
```
m0ksem, line 105: Does it prevent to capture scroll event from children?

packages/ui/src/components/va-backtop/VaBacktop.vue
```diff
@@ -53,10 +52,17 @@ export default defineComponent({
...
+    let targetElement: Element | Window
+
+    const getTargetElement = () => {
+      if (props.target) {
+        return (typeof props.target === 'string'
+          ? document.querySelector(props.target)
+          : props.target) as Element
+      } else {
+        return window as Window
+      }
```
m0ksem, line 64:
```suggestion
    const getTargetElement = () => {
      if (!props.target) { return window }
      if (typeof props.target === 'string') { return document.querySelector(props.target) }
      
      return props.target
    }
```

m0ksem (approved):

packages/ui/src/components/va-backtop/VaBacktop.vue
aluarius: One of the reasons, the second is that without capture scroll event listener not working properly for window.

Issue #1641, opened 2022-04-08
Nuxt alpha release


m0ksem:
Need to make a complete release, so we can test it with nuxt. (Maybe from release/v1.4.0 branch).
This task shouldn't be connected with nuxt-docs features.

- [x] Deal with withTransportConfig https://github.com/epicmaxco/vuestic-ui/pull/1634
- [x] Update colorTranslator to min 1.8 - allows esm build.
- [x] Compile tslib with rollup (or update it someway). Check if we even need to prebuild it with nuxt.
- [x] Check if all components are moved to composition api.
- [x] Check if all fixes from nuxt-docs branch are available in release/v1.4.0. 

@aluarius, feel free to grow this task list.

aluarius:
- VaChip was moved to composition api (https://github.com/epicmaxco/vuestic-ui/pull/1568).
- VaTreeRoot the same (https://github.com/epicmaxco/vuestic-ui/pull/1568).
- VaPopup & VaScrollbar are not in composition api, but they are deprececated.
- There are some doubts about VaTimeLine.
- Others are in composition api.

PR #1668, opened 2022-04-18
[#1641] VaChip moved on composition api

aluarius:
## Description
VaChip moved to composition api.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-chip/VaChip.vue
+++ b/packages/ui/src/components/va-chip/VaChip.vue
@@ -42,7 +42,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType, ref, reactive, computed } from 'vue'
+import { defineComponent, PropType, ref, computed } from 'vue'
 import {
   getBoxShadowColor,
   getHoverColor,
@@ -90,7 +90,7 @@ export default defineComponent({
 
     const { getColor } = useColors()
     const colorComputed = computed(() => getColor(props.color))
-    const borderColor = computed(() => props.outline ? colorComputed : '')
+    const borderColor = computed(() => props.outline ? colorComputed.value : '')
 
     const { tagComputed, hrefComputed } = useRouterLink(props)
 
@@ -123,12 +123,12 @@ export default defineComponent({
     })
 
     const computedStyle = computed(() => {
-      const result = reactive({
+      const result = {
         color: colorComputed.value,
         borderColor: borderColor.value,
         background: '',
         boxShadow: shadowStyle.value,
-      })
+      }
 
       if (props.outline || props.flat) {
         if (hasKeyboardFocus.value) {
--- a/packages/ui/src/components/va-chip/VaChip.vue
+++ b/packages/ui/src/components/va-chip/VaChip.vue
@@ -17,8 +17,8 @@
       class="va-chip__inner"
       v-on="keyboardFocusListeners"
       @focus="$emit('focus')"
-      @mouseenter="updateHoverState(true)"
-      @mouseleave="updateHoverState(false)"
+      @mouseenter="onMouseEnter"
+      @mouseleave="onMouseLeave"
       :tabindex="indexComputed"
     >
       <va-icon
@@ -42,7 +42,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType, ref, computed } from 'vue'
+import { defineComponent, PropType, computed } from 'vue'
 import {
   getBoxShadowColor,
   getHoverColor,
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
-  get computedStyle () {
-    const borderColor = this.outline ? this.colorComputed : ''
-    const computedStyle = {
-      color: this.colorComputed,
-      borderColor,
-      background: '',
-      boxShadow: this.shadowStyle,
-    }
+    const computedStyle = computed(() => {
+      const result = reactive({
```
m0ksem, line 126: There is no reason to create reactive variable inside computed.

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
-    const size: any = {
+    const size = {
       small: '0.875rem',
       medium: '1rem',
       large: '1.25rem',
-    }
-    return size[this.size]
-  }
+    } as Record<string, string>
+    const iconSize = computed(() => size[props.size])
```
m0ksem, line 107: Sad to see size value doesn't connected with SizeConfig. Let's have some TODO here on new issue.

packages/ui/src/components/va-chip/VaChip.vue
aluarius: Without reactive `borderColor` in not working in proper way. Think just can move it in `computedStyle`.

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
     const { hasKeyboardFocus, keyboardFocusListeners } = useKeyboardOnlyFocus()
 
-    return {
-      hasKeyboardFocus,
-      keyboardFocusListeners,
-    }
-  })
+    const { getColor } = useColors()
+    const colorComputed = computed(() => getColor(props.color))
+    const borderColor = computed(() => props.outline ? colorComputed : '')
```
m0ksem, line 93:
```suggestion
    const borderColor = computed(() => props.outline ? colorComputed.value : '')
```

m0ksem (changes requested):
Here is found one minor issue.

packages/ui/src/components/va-chip/VaChip.vue
aluarius: Thanks, overlooked this. Related to the question with reactive.

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
-) {
-  SetupContext = setup(() => {
+
+  emits: [...useStatefulEmits, 'focus'],
+
+  props: {
+    ...useRouterLinkProps,
+    ...useColorProps,
+    ...useStatefulProps,
+    modelValue: { type: Boolean as PropType<boolean>, default: true },
```
m0ksem, line 69:
```suggestion
    modelValue: { type: Boolean, default: true },
```

Vue will cast type from `Boolean` constructor :)

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
-  hoverState = false
-  focusState = false
+    const { tagComputed, hrefComputed } = useRouterLink(props)
 
-  created () {
-    watch(() => this.hoverState, (value) => {
-      this.updateHoverState(value)
-    })
-  }
+    const hoverState = ref(false)
```
m0ksem, line 97: Take a look at `useHover` composable.

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,136 @@
...
     }
-  }
-}
+
+    return {
+      keyboardFocusListeners,
+      updateHoverState,
+      indexComputed,
+      computedStyle,
+      computedClass,
```
m0ksem, line 159: What I'm thinking about. Can't we declare computedClass here in return statement? And another things like close, computedStyle and other properties.

This can be better code structure, because you know exactly which properties are required for template and which properties are required for another properties (computes).

m0ksem (approved):

packages/ui/src/components/va-chip/VaChip.vue
aluarius: https://github.com/epicmaxco/vuestic-ui/issues/1671

packages/ui/src/components/va-chip/VaChip.vue
```diff
@@ -35,139 +35,128 @@
...
-  get iconSize () {
-    const size: any = {
+    const size = {
       small: '0.875rem',
       medium: '1rem',
       large: '1.25rem',
-    }
-    return size[this.size]
-  }
+    } as Record<string, string>
```
LighthouseKeeperYN, line 96:
```suggestion
    }
```

LighthouseKeeperYN (approved):

packages/ui/src/components/va-chip/VaChip.vue
aluarius: Without `Record<string, string>` there is a typescript error for `size[props.size]`. Anyway, this part will be refactored (https://github.com/epicmaxco/vuestic-ui/issues/1671).

Issue #1669, opened 2022-04-18
Slider with "min", "max" and "pins" is displayed incorrectly 


4000degrees:
**Vuestic-ui version:** 1.3.5

**Steps to reproduce**
Create slider with the following settings:
```vue
    <va-slider 
      v-model="value"
      pins
      :min="0"
      :max="23"
     />
```

**What is the expected behavior?**
The component is displayed with pins spanning the whole width of the slider

**What is the current behavior?**
The pins are all stacked on the left side of the slider
![image](https://user-images.githubusercontent.com/42625050/163799446-03c20e90-2297-4e3c-8a5b-9312550ca779.png)

**Other information**
https://codesandbox.io/s/vigilant-wind-z8f5yg

PR #1672, opened 2022-04-18
[#1669] fixed pins position style

aluarius:
close: #1669 

## Description
Fixed pins position style. The initial problem is caused by some props combinations that  weren't processed.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-slider/VaSlider.vue
+++ b/packages/ui/src/components/va-slider/VaSlider.vue
@@ -519,11 +519,8 @@ export default defineComponent({
       if (props.min === 0 && props.max === 100 && props.step !== 1) {
         // only step prop passed
         return props.step
-      } else if ((props.min !== 0 || props.max !== 100) && props.step === 1) {
-        // only min or/and max prop passed
-        return (100 / (props.max - props.min))
       } else {
-        // step and (min or max) prop passed
+        // min or max prop passed (with or without step)
         return ((props.step) * 100) / (props.max - props.min)
       }
     })
--- a/packages/ui/src/components/va-slider/VaSlider.vue
+++ b/packages/ui/src/components/va-slider/VaSlider.vue
@@ -515,11 +515,11 @@ export default defineComponent({
       }
     }
 
-    const pinPositionStep = computed(() => ((props.step) * 100) / (props.max - props.min))
+    const pinPositionStep = computed(() => props.step / (props.max - props.min) * 100)
     const getPinStyles = (pin: number) => ({
       backgroundColor: checkActivePin(pin) ? getColor(props.color) : getHoverColor(getColor(props.color)),
       [pinPositionStyle.value]: `${pin * pinPositionStep.value}%`,
-      transition: hasMouseDown.value ? 'none' : 'background-color .3s ease-out .1s',
+      transition: hasMouseDown.value ? 'none' : 'var(--va-slider-pin-transition)',
     })
 
     const getPos = (e: MouseEvent | Touch) => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-slider/VaSlider.vue
```diff
@@ -515,9 +515,21 @@ export default defineComponent({
...
+    const pinPositionStep = computed(() => {
+      if (props.min === 0 && props.max === 100 && props.step !== 1) {
+        // only step prop passed
+        return props.step
+      } else if ((props.min !== 0 || props.max !== 100) && props.step === 1) {
+        // only min or/and max prop passed
+        return (100 / (props.max - props.min))
+      } else {
+        // step and (min or max) prop passed
+        return ((props.step) * 100) / (props.max - props.min)
```
m0ksem, line 527: Can't we just use this last formula?

aluarius: Yes, you're right. There was another formula before, but now (taking into account, that step can be === 1 in the second case), they're the same.

packages/ui/src/components/va-slider/VaSlider.vue
```diff
@@ -515,9 +515,10 @@ export default defineComponent({
       }
     }
 
+    const pinPositionStep = computed(() => ((props.step) * 100) / (props.max - props.min))
     const getPinStyles = (pin: number) => ({
       backgroundColor: checkActivePin(pin) ? getColor(props.color) : getHoverColor(getColor(props.color)),
-      [pinPositionStyle.value]: `${pin * props.step}%`,
+      [pinPositionStyle.value]: `${pin * pinPositionStep.value}%`,
       transition: hasMouseDown.value ? 'none' : 'background-color .3s ease-out .1s',
```
rustem-nasyrov, line 522: `'background-color .3s ease-out .1s'` better move it to CSS variable.

rustem-nasyrov (changes requested):

packages/ui/src/components/va-slider/VaSlider.vue
```diff
@@ -515,9 +515,10 @@ export default defineComponent({
       }
     }
 
+    const pinPositionStep = computed(() => ((props.step) * 100) / (props.max - props.min))
```
RVitaly1978, line 518:
```suggestion
    const pinPositionStep = computed(() => (props.step * 100) / (props.max - props.min))
```

RVitaly1978 (changes requested):
Pay attention to code line #326. I guess there is a mistake.

packages/ui/src/components/va-slider/VaSlider.vue
m0ksem: IMHO, the percent formula (from school) looks like `part / whole * 100`, so my suggestion here ✌:

```suggestion
    const pinPositionStep = computed(() => (props.step / (props.max - props.min) * 100) 
```

packages/ui/src/components/va-slider/_variables.scss
```diff
@@ -12,6 +12,9 @@
   --va-slider-track-transition: 0.5s ease-out;
   --va-slider-track-opacity: 1;
 
+  /* Pin */
+  --va-slider-pin-transition: 0.3s ease-out 0.1s;
```
RVitaly1978, line 16:
```suggestion
  --va-slider-pin-transition: background-color 0.3s ease-out 0.1s;
```

RVitaly1978 (changes requested):

aluarius:
> Pay attention to code line 326. I guess there is a mistake.

Yes, it may be related to https://github.com/epicmaxco/vuestic-ui/issues/1673, but it will be solved separately from this pr.

LighthouseKeeperYN (approved):

rustem-nasyrov (approved):

RVitaly1978 (approved):

Issue #1657, opened 2022-04-13
`VaRadio` and `VaSwitch` have not `readonly` state


RVitaly1978:
**Vuestic-ui version:** 1.3.5

PR #1680, opened 2022-04-19
[#1657] VaRadio readonly state, VaRadio & VaSwitch examples updated

aluarius:
close: #1657 

## Description
- `VaSwitch` has readonly state. Example added.
- Added readonly state for `VaRadio` (and example). I'd thought about implementing this via `useSelectable`, but it is not suitable in its current form, and its adaptation looks like an unnecessary complication of the common composable cause of a single case.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-switch/VaSwitch.vue
+++ b/packages/ui/src/components/va-switch/VaSwitch.vue
@@ -147,6 +147,7 @@ export default defineComponent({
       'va-switch--small': props.size === 'small',
       'va-switch--large': props.size === 'large',
       'va-switch--disabled': props.disabled,
+      'va-switch--readonly': props.readonly,
       'va-switch--left-label': props.leftLabel,
       'va-switch--error': computedError.value,
       'va-switch--on-keyboard-focus': hasKeyboardFocus.value,
@@ -249,6 +250,15 @@ export default defineComponent({
 
   &--disabled {
     @include va-disabled;
+  }
+
+  &--readonly {
+    @include va-readonly;
+
+    .va-switch__label {
+      cursor: initial;
+      pointer-events: auto;
+    }
   }
 
   &--left-label {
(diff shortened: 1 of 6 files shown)


rustem-nasyrov:
What's the difference here between `readonly` and `disabled` states for the `VaRadio`?

aluarius:
> What's the difference here between `readonly` and `disabled` states for the `VaRadio`?

Difference between `readonly` and `disabled`  everywhere is only in appearance - readonly has a usual components view, disabled - the same, but with opacity.

rustem-nasyrov:
> > What's the difference here between `readonly` and `disabled` states for the `VaRadio`?
> 
> Difference between `readonly` and `disabled` everywhere is only in appearance - readonly has a usual components view (without ability to interact), disabled - the same, but with opacity.
> 
>  

Thank you, now I got it. :)

LighthouseKeeperYN (approved):

rustem-nasyrov (changes requested):
For `readonly` state you need to add a couple of CSS code:
```
cursor: default;
pointer-events: none;
```

RVitaly1978 (changes requested):
I think it would be better if we add this CSS for the `readonly` state of the components:
```
cursor: default;

.component-label-class-selector {
  cursor: initial;
}
```
in this case, it will be clearer for the user that the text of the label can be selected.

RVitaly1978 (approved):

packages/ui/src/components/va-switch/VaSwitch.vue
```diff
@@ -251,6 +252,15 @@ export default defineComponent({
     @include va-disabled;
   }
 
+  &--readonly {
+    @include va-readonly;
+
+    .va-switch__label {
+      cursor: initial;
+      pointer-events: auto;
+    }
```
rustem-nasyrov, line 261: Doesn't it inherits from parent `--readonly` class?

aluarius: No. Everything was tested and works as Vitaly suggested - "arrow" cursor for interactive area and "text" cursor for label.

PR #1683, opened 2022-04-19
[#1667] Blurry background for VaModal 

### LINKED ISSUE #1667: Blurry background for VaModal
Looks hot 🔥

image

How to do it. 

On show, add `filter: blur(0.75rem)` to #app, on hide remove it. It would blur app content. Because modal appends to body it wouldn't be buried. (Remember about nested modals)

zvenigorodskaia:
close #1667

## Description

Add blured overlay for VaModal 

## Types of changes


packages/ui/src/components/va-modal/VaModal.demo.vue
```diff
@@ -123,6 +123,17 @@
         :message="message"
       />
     </VbCard>
+    <VbCard title="blur: true">
+      <button @click="showModalBlur = !showModalBlur">
+        Show modal
+      </button>
+      <va-modal
+        v-model="showModalBlur"
+        :blur="true"
```
RVitaly1978, line 132:
```suggestion
        blur
```

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -4,9 +4,18 @@
...
           :style="$props.overlay && computedOverlayStyles"
           @click="onOutsideClick"
         />
+
+        <div
+          class="va-modal__overlay-blur"
+          v-if="$props.blur"
+          :style="$props.blur && computedBlurStyles"
+          @click="onOutsideClick"
+        />
```
RVitaly1978, line 17: Can we remove that and only use `class="va-modal__overlay"` (div above) with blurry styling if the `blur` property is on?

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -140,6 +149,8 @@ export default defineComponent({
     withoutTransitions: { type: Boolean as PropType<boolean>, default: false },
     overlay: { type: Boolean as PropType<boolean>, default: true },
     overlayOpacity: { type: [Number, String] as PropType<number | string>, default: 0.6 },
+    blur: { type: Boolean as PropType<boolean>, default: false },
+    blurOpacity: { type: [Number, String] as PropType<number | string>, default: 0.2 },
```
RVitaly1978, line 153: Can we remove that and use `overlayOpacity` for the blur mode as well?

RVitaly1978 (changes requested):

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -167,6 +178,17 @@ export default defineComponent({
...
     })
+    const computedBlurStyles = computed(() => {
+      // NOTE Not sure exactly what that does.
+      // Supposedly solves some case when background wasn't shown.
+      // As a side effect removes background from nested modals.
+
+      if (!props.blur) { return }
+
+      return {
+        'backdrop-filter': `blur(${props.blurOpacity}rem)`,
```
m0ksem, line 189: Backdrop-filter doesn't work in Firefox. The original idea was to use filter on modal's target.

This way blur will not multiply on nested modals.

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -140,6 +149,8 @@ export default defineComponent({
     withoutTransitions: { type: Boolean as PropType<boolean>, default: false },
     overlay: { type: Boolean as PropType<boolean>, default: true },
     overlayOpacity: { type: [Number, String] as PropType<number | string>, default: 0.6 },
+    blur: { type: Boolean as PropType<boolean>, default: false },
```
m0ksem, line 152: I think having one prop for blur would be enough.
```
<va-modal blur />

<va-modal blur="1.5rem" />
```

m0ksem (changes requested):

RVitaly1978 (approved):

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -213,8 +214,12 @@ export default defineComponent({
     watch(valueComputed, (value: boolean) => {
       if (value) {
         window.addEventListener('keyup', listenKeyUp)
+        if (props.blur) {
+          document.body.classList.add('overlay--blured')
```
asvae, line 218: I would gently suggest moving this feature into separate block. So that features are not mixed up.

I.e.

```js
if (props.blur) {
  if (value) {
    document.body.classList.add('overlay--blured')
  } else {
    document.body.classList.remove('overlay--blured')
  }
}
```

packages/ui/src/components/va-modal/_variables.scss
```diff
@@ -15,6 +15,7 @@
   --va-modal-basic-duration: 0.3s;
   --va-modal-opacity-transition: opacity calc(var(--va-modal-basic-duration) * 0.5) cubic-bezier(1, 0.5, 0.8, 1);
   --va-modal-transform-transition: transform var(--va-modal-basic-duration) ease;
+  --va-modal-blur: 4px;
```
asvae, line 18: Very small thing:
blur-radius?

asvae (approved):
Couple of small things, but overall looks good! 😁

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -264,6 +269,13 @@ export default defineComponent({
 @import "../../styles/resources";
 @import "variables";
 
+.overlay--blured > :not(div[class*="va-"]) {
```
m0ksem, line 272:
```suggestion
.va-modal-overlay-background--blured > :not(div[class*="va-"]) {
```

packages/ui/src/components/va-modal/VaModal.vue
m0ksem: Let's have a guide in docs, where users can find how to change `-va-blur-radius` CSS variable, instead of making this prop as string.

m0ksem (approved):

Derranion (approved):

m0ksem (approved):
Hey. Will you mind updating docs as well? We need text about blur prop in API block down below the page and maybe some example.

It also will be great to mention  `--va-modal-blur-radius` css variable and how used can change it.

PR #1686, opened 2022-04-20
added opportunity to add string and number dates

legit1One:
## Description

Now dateInput may accept string and number as modelValue and convert it to Date object

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-date-input/hooks/range-model-value-guard.ts
+++ b/packages/ui/src/components/va-date-input/hooks/range-model-value-guard.ts
@@ -13,8 +13,15 @@ export type useRangeModelValueGuardProps = {
 export const useRangeModelValueGuard = (
   modelValue: Ref<VaDatePickerModelValue | undefined>,
   disabled: Ref<boolean>,
+  inputDateParse?: ((input: string | undefined | VaDatePickerModelValue) => Ref<VaDatePickerModelValue>) | undefined,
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-date-input/hooks/range-model-value-guard.ts
```diff
@@ -13,8 +13,15 @@ export type useRangeModelValueGuardProps = {
...
+  inputDateParse?: ((input: string | undefined | VaDatePickerModelValue) => Ref<VaDatePickerModelValue>) | undefined,
 ) => {
-  const bufferValue = ref<VaDatePickerModelValue | undefined>(modelValue.value)
+  let bufferValue: Ref<VaDatePickerModelValue | undefined>
+
+  if (inputDateParse !== undefined) {
+    bufferValue = inputDateParse(modelValue.value)
+  } else {
+    bufferValue = ref<VaDatePickerModelValue | undefined>(modelValue.value)
+  }
```
RVitaly1978, line 24: Probably we need to use the `computed` (instead of `ref`) here with getters and seters

RVitaly1978 (changes requested):

LighthouseKeeperYN:
Closing it for now. I'll integrate those changes as part of the date rework.

PR #1689, opened 2022-04-20
Unnecessary dependencies removing

aluarius:
## Description
Deleted unnecessary dependencies, used in the only one outdated component, which is already seems to be non-working. But just in case - functionality was partially compensated.

P.S. Instead of `erd` i wanted to use `useResizeObserver` - in this case component should be partially moved to composition api, but was told not to waste time on this.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-scrollbar/VaScrollbar.vue
+++ b/packages/ui/src/components/va-scrollbar/VaScrollbar.vue
@@ -32,11 +32,6 @@
 </template>
 
 <script>
-import { detect } from 'detect-browser'
-import erd from 'element-resize-detector'
-
-const browser = detect()
-
 export default {
   name: 'VaScrollbar',
   props: {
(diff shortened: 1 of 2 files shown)


asvae:
@aluarius could you create an issue for scrollbar component replacement/fix?

Apart from that looks good to me 👍. In a sense that it reduces dependencies, thus simplifies nuxt transition (was that an objective)?

asvae (approved):

packages/ui/src/components/va-scrollbar/VaScrollbar.vue
```diff
@@ -108,7 +107,7 @@ export default {
     },
     scroll (e) {
       let delta = (e.deltaY * 0.01 * this.speed)
-      if (browser.name === 'firefox') {
+      if (navigator?.userAgent?.toLowerCase().indexOf('firefox') > -1) {
```
LighthouseKeeperYN, line 110:
```suggestion
      if (navigator?.userAgent?.toLowerCase().includes('firefox')) {
```

LighthouseKeeperYN (approved):

aluarius:
> @aluarius could you create an issue for scrollbar component replacement/fix?

Sure.

> Apart from that looks good to me 👍. In a sense that it reduces dependencies, thus simplifies nuxt transition (was that an objective)?

Yes, it's a related task.

PR #1713, opened 2022-04-27
Requesting modal with method, without putting into <template>

### LINKED ISSUE #1216: Requesting modal without manual HTML tags, like toast
In the application I'm building I've had several situations where I've needed to use a modal but that it wasn't convenient to throw in a `<va-modal />` element. Ideally, I'd love a solution like the toast component which doesn't require adding tags. The main issue is that it's easy to lose the 'context' of the current interaction; when using `<va-modal />`, I need to store any context information as additional props in the component. Instead, by allowing an inline call, all the necessary context will be close by.

Following the toast pattern:

```js
this.$vaToast.init({
  message: err.message,
  position: 'bottom-right',
  color: 'warning'
});
```

I would hope for something maybe like this:

```js
const foo_id = 12;
this.$vaModal.init({
  message: `Are you sure you would like to delete foo ${foo_id}?`,
  title: `Delete Foo ${foo_id}`,
  mobileFullscreen: false,
  okText 'KILL IT!',
  ok: async () => { await del(foo_id); },
  cancel: myCancelHandler
});
```

In this regard it's almost like an asynchronous version of the `confirm()` function.

I understand that only a subset of functionality would work with this approach since additional DOM elements can't be added to the component. Choosing where to attach it to the DOM also loses granularity but at least in my case it should be fine attaching it to the root element.

### LINKED ISSUE #1849: OnBeforeLeave pointed on wrong element
Listeners moved from modal-element component to div.

Must be;
```
      <modal-element
        name="va-modal__container--transition"
        :isTransition="!$props.withoutTransitions"
        appear
        @beforeEnter="onBeforeEnterTransition"
        @afterEnter="onAfterEnterTransition"
        @beforeLeave="onBeforeLeaveTransition"
        @afterLeave="onAfterLeaveTransition"
      >
```
We have now:
image

xiongmao86:
## Description
Requesting va-modal with method, user can choose between to ways of using `va-modal`, for example you can bind building function as event listener, or adding inside <template>.

close #1216
close #1849 

```vue
const foo_id = 12;
this.$vaModal.init({
  message: `Are you sure you would like to delete foo ${foo_id}?`,
  title: `Delete Foo ${foo_id}`,
  mobileFullscreen: false,
  okText 'KILL IT!',
  ok: async () => { await del(foo_id); },
  cancel: myCancelHandler
});
```

- [x] add `VaModalPlugin`
- [x] add `useModal` hook
- [x] add docs with examples
- [x] add demo example

chrome-capture

chrome-capture (1)

## Types of changes


xiongmao86:
Hi, @asvae, `bootstrap-vue` is using plugin to implement and wire up modal message box. I try to wire up using plugin, and intent to implement vaModal building function using CreateVNode, Am I doing correctly?

m0ksem:
Hi, @xiongmao86! We have similar logic in [VaToast](https://vuestic.dev/en/ui-elements/toast#basic-usage). You can find an example [here](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-toast/index.ts#L121).

We have been [refactored ](https://github.com/epicmaxco/vuestic-ui/blob/release/v1.4.0/packages/ui/src/components/va-toast/toast.ts) plugins we use for better tree-shaking and SSR support. I think you can rebase to `release/v1.4.0` branch.

packages/ui/src/components/va-modal/modal.ts
```diff
@@ -0,0 +1,58 @@
...
+  children,
+  element,
+  appContext,
+}: { props?: { [key: string]: any }; children?: any; element?: HTMLElement, appContext?: AppContext } = {}): { vNode: VNode; el?: HTMLElement } => {
+  let el: HTMLElement | null | undefined = element
+
+  // eslint-disable-next-line prefer-const
+  let vNode: VNode | null
+
+  vNode = createVNode(component, { ...props, modelValue: value, onUpdateModelValue: ((event) => {value = event}), children)
```
xiongmao86, line 24: How do I wire up v-model, I mean to pass in `update:model-value` to vNode?

xiongmao86:
@m0ksem, it seems `modelValue` is turned into a Ref by setting `stateful` prop as true. But once I click the triggering button, and either click `OK` or `Cancel`, or just click outside the modal, the modal will disappear. And when I click the triggering button again, it failed to response. It's strange.

xiongmao86:
Sorry, @m0ksem, I don't know how to call show/hide method, and I wanted to put component under a parent component, but I don't know how to get a parent element of button triggering `$vaModal.init` either. Can you help me with that?

RVitaly1978 (commented):
@[xiongmao86](https://github.com/xiongmao86), FYI I will update this PR.

xiongmao86:
@RVitaly1978, Thanks for finishing the pr, I really don't know how to implement it.

packages/ui/src/components/va-modal/plugin/index.ts
```diff
@@ -0,0 +1,22 @@
+import { App } from 'vue'
+import { defineVuesticPlugin, defineGlobalProperty } from '../../../vuestic-plugin/utils'
+import { createModalInstance } from '../modal'
+import { ModalOptions } from '../types'
+
+const createVaModalPlugin = (app: App) => ({
+  init (options: string | ModalOptions) {
+    return createModalInstance(options, app?._context)
```
aluarius, line 8: Not app?.appContext? Think we shouldn't use private properties.

packages/ui/src/components/va-modal/modal.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (props?.withoutTransitions && !value) {
+      nextTick(() => { // this is required for all handlers to execute
+        destroy(el, vNode)
+      })
+    }
+  }
+
+  vNode = h(component, {
+    ...props,
+    ...(props?.stateful === undefined && { stateful: true }),
```
aluarius, line 49:
```suggestion
    stateful: props?.stateful ?? true,
```

aluarius (approved):

packages/ui/src/components/va-modal/plugin/index.ts
RVitaly1978: we don't have access to `appContext` here, only `_context`

packages/docs/src/page-configs/ui-elements/modal/modal-api.ts
```diff
@@ -0,0 +1,86 @@
...
+    { title: 'Method name', type: 'markdown' },
+    { title: 'Api Type', type: 'markdown' },
+    { title: 'Description', type: 'markdown' },
+  ],
+  [
+    ['init(options: string | ModalOptions)', 'Options | Composition', 'Creates new modal instance. Returns modal instance VNode'],
+  ],
+)
+
+const optionsTableData: [keyof ModalOptions, string][] = [
```
m0ksem, line 15: Better have object typeof ModalOptions here. We probably will forget to add some props here...

```ts
const o: Record<keyof ModalOptions, string>
```

Will it throw error if we forget to add `title` key here?

m0ksem (approved):

packages/ui/src/components/va-modal/plugin/index.ts
RVitaly1978: Opened new issue #1853

Issue #744, opened 2021-05-14
Create color themes as service


m0ksem:
closes #325
closes #1255

For now, we have colors and components configs as services. Also, we have a function for updating CSS variables on runtime.

I propose to combine these services in one big Theme service, which allow the user to define default colors and specific colors for each component.

For example:

```
const themeDark = {
  colors: {
     'primary': '#ff00ff',
     'background': '#000000',
     'text': '#ffffff',
  },
  componentsCSSVarialbes: { // or cssVariables
    VaCard: {
    	backgroundColor: '#1f1e1e',
    	textColor: '#eeee'
    },
    VaNavbar: {
       backgroundColor: 'primary' // Which should compile to `background-color: var(--va-primary);`
       textColor: '#fff'
    },
    VaInput: {
		padding: '0.75rem',
    },
    VaButton
    ....
  }
}
```

In the example below, I described high contrast theme with lighter cards.
Switching from to themeDark from the default theme will look like this:

![image](https://user-images.githubusercontent.com/23530004/118202025-9f28a300-b461-11eb-92fa-4b6f4abc3289.png)

m0ksem:
Vuestic admin themes test [implementation](https://github.com/epicmaxco/vuestic-admin/blob/master/src/services/vuestic-ui/themes.ts)

[Dark](https://vuestic.epicmax.co/admin/dashboard?theme=dark)
[Light](https://vuestic.epicmax.co/admin/dashboard?theme=light)
[Original](https://vuestic.epicmax.co/admin/dashboard?theme=original)
[Semi-dark](https://vuestic.epicmax.co/admin/dashboard?theme=semi-dark)

Take a look on this implementation in production (especially on dark theme [form components](https://vuestic.epicmax.co/admin/forms/form-elements))
Read how colors config [works](https://vuestic.dev/en/services/colors-config).

Investigate list of colors that we need. As I can see we need at least two colors for backgrounds, texts etc. Name them convenient and simple.

Make sure all components uses colors from color config. Use [color-function](https://github.com/epicmaxco/vuestic-ui/tree/develop/packages/ui/src/services/color-config). Feel free to extend colors-config if needed.

Maybe we need another implementation of getFocusColor etc.

m0ksem:
Make sure all components following auto text-color functional and/or can be configurable via props.
![image](https://user-images.githubusercontent.com/23530004/164259128-d5cd19b8-ec76-430f-bd2f-d1cd6f230499.png)
[...]

PR #1722, opened 2022-05-02
Fix#980. Fixed hardcode colors. 

### LINKED ISSUE #980: Fix hardcoded CSS variables in VaInput, VaSelect
https://localhost:3000/admin/dashboard?theme=dark

image
The dropdown must use some background CSS variables as input does.

image
Must use something like this `setHSLAColor(getColor('success'), { a: 0.3 })`

image
Checkbox and Radio should also be relative to `--va-background` color.

image

image
Must use auto text color

image
Must use auto text color like badge do

image

image
Modals must have a dark background

image
Divider color

----

There can be issues with theme color instead of unrelative CSS variables. 

You can implement a theme switcher (take it from Vuestic-Admin) for docs.

Sadnessa:
## Description
Closes #980.

## Types of changes


m0ksem (approved):
Looks like we need addtional colors in getTextColor: `text-dark, text-light` instead of `white, dark`.
We also need to replace `-va-white, -va-dark, -va-divider, -va-background` to `-va-background, -va-backgorund-mute, -va-background-soft`
I took this color names from vue sample project (made with vue init) here:
image



Here I just change -va-white, -va-dark and -va-background colors to get black theme. I also added text colors.
image

image


Colors should be reimagined in order of https://github.com/epicmaxco/vuestic-ui/issues/744

Need to deal with box-shadows. They are usually hardcoded. Must be another task.
image

m0ksem:
@aluarius, I suggest changing getTextColor to 
```ts
export const getTextColor = (color: ColorInput, darkColor = 'var(--va-text-dark)', lightColor = 'var(--va-text-light)') => {
  const { R, G, B } = new ColorTranslator(color)
  const isLightBackground = Math.sqrt(R * R * 0.241 + G * G * 0.691 + B * B * 0.068) > 130
  return isLightBackground ? darkColor : lightColor
}
```

Then you are able to check dark theme update packages/docs/src/config/theme-config.ts:

```ts
export const COLOR_THEMES: Record<ThemeName, Record<string, string>> = {
  [ThemeName.DEFAULT]: {
    primary: '#2C82E0',
    secondary: '#767C88',
    background: '#0d1117',
    success: '#40e583',
    info: '#2c82e0',
    danger: '#DE1041',
    warning: '#FFAC0A',
    gray: '#babfc2',
    white: '#1B1A1F',
    dark: '#ffffff',
    divided: '#000',
    'text-dark': '#000',
    'text-light': '#fff',
  },
```

packages/ui/src/services/color-config/color-config.ts
```diff
@@ -51,6 +51,10 @@ export const getColor = (prop?: string, defaultColor: string = DEFAULT_COLOR, pr
     return prop
   }
 
+  if (preferVariables && isCSSVariable(prop)) {
```
aluarius, line 54: @m0ksem won't be here again a SSR problem because of `isCSSVariable`?

aluarius (approved):

packages/ui/src/services/color-config/color-config.ts
m0ksem: `isCSSVariable` uses regex under the hood. CSS variable available in SSR with `renderCSSVariable`. Can't see any problems with SSR.

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -534,8 +534,11 @@ export default defineComponent({
     }
 
     .va-data-table__table-tr {
+      position: relative;
+      z-index: 0;
+
       &.selected {
-        background-color: var(--selected-color);
+        background-color: var(--va-selected-color);
```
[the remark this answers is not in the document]
Sadnessa: This variable used in useStyleable hook. There are a lot of variables without '--va' prefix. Should we add this prefix to all css variables in useStyleable hook?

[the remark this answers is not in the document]
RVitaly1978: Yes. That would be right.

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -533,25 +533,25 @@ export default defineComponent({
       }
     }
 
-    .va-data-table__table-tr {
-      position: relative;
-      z-index: 0;
-
-      &.selected {
-        background-color: var(--va-selected-color);
-      }
```
RVitaly1978, line 539: Table rows can be `selected` not only when table is `striped`. Please don't remove this from here.

RVitaly1978 (approved):

PR #1728, opened 2022-05-05
Feat/tree view rework

### LINKED ISSUE #1567: Missing docs for VaTreeView component
image
image

packages/ui/src/components/va-tree-view

Make sure this component satisfy common requirements. Also, would be nice to have these requirements documented.

### LINKED ISSUE #1693: VaTreeNode component
- [x] VaTreeNode should have some kind of modelValue or statefull
- [x] Should calculate text (icon) color relative to background color using getTextColor function
- [x] Need docs
- [x] Mustn't open all nested childs if parent open.

Feel free to make some research and add something to this list.

rustem-nasyrov:
## Related issues
Close #1693, close #1567, close #1693

## Description
demos

### Added:
* Demos
* Selection
* Expanding/collapsing tree nodes
* Filter
* Docs
* Model value
* Template
* Colors

## Types of changes


packages/ui/src/components/va-tree-view/hooks/useTreeBuilder.ts
```diff
@@ -0,0 +1,42 @@
...
+const useTreeBuilder: UseTreeBuilderFunc = ({ nodes, expandAll }) => {
+  const buildTree: TreeBuilderFunc = (nodes, level = 0) => {
+    return nodes.map((node) => {
+      if (node.children?.length) {
+        const children = buildTree(node.children, level + 1)
+
+        return createNode({
+          node,
+          children,
+          expandAll: expandAll,
```
aluarius, line 28:
```suggestion
          expandAll,
```

aluarius (approved):
Looks difficult :)

packages/ui/src/components/va-tree-view/hooks/useTreeBuilder.ts
```diff
@@ -0,0 +1,37 @@
...
+export const createNode: CreateNodeFunc = ({
+  node,
+  children = [],
+  expandAll = false,
+  level = 0,
+}) => ({
+  ...node,
+  children,
+  level,
+  hasChildren: !!children.length,
```
RVitaly1978, line 13:
```ts
key: getKey(node, props.keyBy) || fallbackKey,
label: getLabel(node, props.labelBy) || fallbackLabel,
```
where `fallbackKey` and `fallbackLabel` may include the key and label of parentNode, respectively, and the index of the current node in the node array.

RVitaly1978 (changes requested):

packages/ui/src/components/va-tree-view/VaTreeView.vue
```diff
@@ -1,62 +1,161 @@
...
     color: {
       type: String,
-      default: '',
+      default: () => 'primary',
+    },
+    nodes: {
+      type: Array as PropType<TreeNode[]>,
+      default: () => ([]),
+    },
+    nodeKey: {
```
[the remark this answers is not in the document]
rustem-nasyrov: I don't like it. It looks unclear, `keyBy` of what? I'll leave `nodeKey` name.

rustem-nasyrov:
@aluarius, can you later check how is implemented the accessibility in the component?

aluarius:
> @aluarius, can you later check how is implemented the accessibility in the component?

- every input element should have `aria-label` (for example, every checkbox for `selectable` state - `check *node label value*` for them or append `aria-hidden="true"` if you don't like this variant). See also VaCheckbox components for additional aria-attributes (for example, `aria-checked`);
- missing any interaction options with focused node (how can i expand it via keyboard?). Check this: https://miro.medium.com/max/1400/1*f3t8bHWDOktffVIchD_lTg.png
- according to https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-1/treeview-1a.html here should be also `aria-selected` attribute.

packages/ui/src/components/va-tree-view/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  level?: number
+  children?: TreeNode[]
+  filter?: Ref<string>
+  labelKey?: Ref<string>
+  expandAll?: Ref<boolean>
+}
+
+export type CreateNodeFunc = (props: CreateNodeProps) => TreeNode
+
+export type UseTreeBuilderProps = {
```
m0ksem, line 39: https://github.com/epicmaxco/vuestic-ui/issues/1907#issuecomment-1195712961

types.ts must include only types, that required for user. Store inner types like UseTreeBuilderProps in composables.

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
```diff
@@ -0,0 +1,229 @@
...
+    const {
+      nodeKey,
+      labelKey,
+      iconColor,
+      selectable,
+      colorComputed,
+      getKey,
+      toggleNode,
+      toggleCheckbox,
+    } = inject(TreeViewKey, {
```
m0ksem, line 101: Maybe throw error if VaTreeNode used without VaTreeView? Or warn? Or we want user to be able to use VaTreeNode without VaTreeView?

packages/ui/src/components/va-tree-view/hooks/useTreeBuilder.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    return node[labelKey.value].toLowerCase().includes(filter.value)
+  }).value,
+}))
+
+const useTreeBuilder: UseTreeBuilderFunc = (props) => {
+  const { nodes, expandAll, filter, labelKey } = props
+
+  const buildTree: TreeBuilderFunc = (nodes: TreeNode[], level = 0) => {
+    return nodes.reduce((acc: TreeNode[], node: TreeNode) => {
```
m0ksem, line 30: I'm a bit confused about nodes when somewhere it ref somewhere not 😅. 

Please, remove argument for this FN. And it actually can be moved to computed I see.
```suggestion
    return nodes.value.reduce((acc: TreeNode[], node: TreeNode) => {
```

m0ksem (approved):
Nice job on removing a lot of files from ui and auditing a lot of them in docs.

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
rustem-nasyrov: Recently I [implemented](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/composables/useStrictInject.ts) `useStrictInject` composable, I'll need to merge an actual develop branch into the mine.

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -342,6 +342,10 @@ export const navigationRoutes: NavigationRoute[] = [
         name: 'dropdown',
```
asvae, line 342: In tree-view demo we probably can remove `node-key` as it gets default anyway.

asvae, line 342: Tree view items shouldn't be focusable. That might be a part of bigger issue. If yes feel free to move it to another issue.

asvae, line 342: It's advantageous to use minimal amount of items to test your component.

I.e. if for all cases you only need 2 items (1 category and 1 item) - then we should have 2 items. But solution is yours and if you deem important to keep that number - so be it.

image

packages/docs/src/page-configs/ui-elements/tree-view/examples/CustomizableContent.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <va-tree-view
+    :nodes="nodes"
+    node-key="id"
```
asvae, line 4: image

Can't we use logic from select? The way I see it it's almost the same as for select.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
+          },
+        ],
+      },
+    ],
+  }),
 }
 </script>
+
+<style lang="scss" scoped>
+.custom-body-example {
```
asvae, line 341: Doesn't seem to be used anywhere?

packages/docs/src/components/sidebar/navigationRoutes.ts
asvae, line 342: I'm not sure if that's of any importance, but non selectable non category items are slightly smaller.

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -342,6 +342,10 @@ export const navigationRoutes: NavigationRoute[] = [
...
         displayName: 'menu.dropdown',
```
asvae, line 343: image

Filter looks broken. I would assume a couple of things:
* `filter-by` prop should define how filtering would work. (could be string or function)
* default value should be the same as for `text-by` (as user would search by text displayed normally)

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
+        :key="index"
+        v-model="selectionType"
+        :option="option"
+      />
+      <br />
+      <p>
+        {{ JSON.stringify(selectedNodes) }}
+      </p>
+      <br />
+      <va-tree-view
```
asvae, line 36: I'm able to select items even without v-model. That should not be the case unless component is `stateful`.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
+        v-model="selectedNodes"
```
asvae, line 37: image

That doesn't seem robust enough.

I.e., the way I see it, selecting all nodes of category should also select category, which it doesn't currently.
Yet, when I select category - items are getting selected.

Value also doesn't make sense from user perspective. I would say that categories should not be included.

Text for docs example could be improved a bit I think. You can use videogame genres as categories and titles as items for instance 😁.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
+        :nodes="nodesWithIcons"
+        node-key="id"
+        expand-all
+        selectable
```
asvae, line 41: image

When some items in category are selected - I think we should make it indeterminate.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
+        :selection-type="selectionType"
```
asvae, line 42: Both tree-views use the same v-model. I would expect structure to look the same, yet it doesn't.

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,352 @@
...
-          <va-tree-node>
-            Node text
-          </va-tree-node>
-          <va-tree-node>
-            Node text
-          </va-tree-node>
-        </va-tree-category>
+        <template #content="prop">
+          <div class="d-flex align--center">
+            <va-avatar
```
asvae, line 60: Not a big deal, but probably no much sense to do something so complicated for demo, just outlining the slot and scoped props would do.

asvae (changes requested):
Still needs some work, I believe.

packages/docs/src/components/sidebar/navigationRoutes.ts
rustem-nasyrov: @asvae, but it's accessibility requirements. I'll add a property to toggle tree items focusable state.

aluarius: Yeah, they should be focusable. The only problem we've met is that focus state because of selection via mouse or touchpad is annoying and should appear only for keyboard navigation. I think that property is redundant - i'll implement keyboard-only focus state for all component (after discussion via https://github.com/epicmaxco/vuestic-ui/issues/2200#issuecomment-1206972780).

rustem-nasyrov: I've used examples from the Figma design, where is given a lot of use cases. And I think better to have more examples with different variations to see if something was broken during the development.

asvae: All right, up to you. :)

asvae: @aluarius can't you explain to me why would items be focusable? (just checked a couple of examples and never that's the case)

Maybe as a prop at most?

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
rustem-nasyrov: It's a single tree view. :)

packages/docs/src/locales/en/en.json
```diff
@@ -1305,6 +1305,30 @@
       },
       "methods": {},
       "slots": {}
+    },
+    "VaTreeView": {
+      "props": {
+        "expandAll": "Expand or hide all nodes by default",
```
aluarius, line 1311: Dots here and below C:

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,252 @@
...
+            :name="node.icon"
+            :color="node.iconColor"
+            :title="node.label"
+            size="20px"
+          />
+        </template>
       </va-tree-view>
     </VbCard>
-    <VbCard title="Color">
+    <VbCard title="Selection and icons">
```
asvae, line 19: Selecting node shouldn't expand all its children. Same with deselection.

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,252 @@
...
-          <va-tree-category
-            label="Category text"
-            is-open
-          >
-            <va-tree-node>
-              Node text
-            </va-tree-node>
-          </va-tree-category>
-        </va-tree-category>
+    <VbCard title="Default">
```
asvae, line 3: image
  
Does it make sense that childe nodes have left padding without icon? Can we check other frameworks?

asvae:
Looks close to finish, but reactivity isn't there.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
rustem-nasyrov: I've kept it like in the figma. If it's not necessary, we can move it to new issue? As a little fix I can try to reduce the width of the highlighting background.
image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,266 @@
...
+        v-model="selectedColor"
+        :palette="colorsPalette"
+      />
+      <br />
+      <p>
+        {{ JSON.stringify(selectedNodes) }}
+      </p>
+      <br />
+      <va-tree-view
+        v-model:checked="selectedNodes"
```
asvae, line 37: Why can't we use just `v-model`?

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,266 @@
...
-            label="Category text"
-            is-open
-          >
-            <va-tree-node>
-              Node text
-            </va-tree-node>
-          </va-tree-category>
-        </va-tree-category>
+    <VbCard title="Default">
+      <va-tree-view :nodes="nodes" />
```
asvae, line 4: Please include icons here and anything else if it's supported in default tree view. It's not clear for me I can use icons without slot.

packages/ui/src/components/va-tree-view/hooks/useTreeHelpers.ts
```diff
@@ -0,0 +1,120 @@
...
+import type { TreeNode, TreeViewFilterMethod, TreeViewPropKey } from '../types'
+
+import { getValueByKey } from '../../../services/utils'
+
+export const useTreeViewProps = {
+  nodes: {
+    type: Array as PropType<TreeNode[]>,
+    default: [],
+  },
+  stateful: {
```
asvae, line 12: There is no demo for stateful.

packages/ui/src/components/va-tree-view/hooks/useTreeHelpers.ts
```diff
@@ -0,0 +1,120 @@
...
+  selectable: {
+    type: Boolean,
+    default: false,
+  },
+  selectionType: {
+    type: String,
+    default: 'leaf',
+    validator: (v: string) => ['leaf', 'independent'].includes(v),
+  },
+  valueBy: {
```
asvae, line 25: No demo for either of these. How would you know it works?

packages/ui/src/components/va-tree-view/hooks/useTreeHelpers.ts
```diff
@@ -0,0 +1,120 @@
...
+    type: [String, Function] as PropType<TreeViewPropKey>,
+    default: 'id',
+  },
+  textBy: {
+    type: [String, Function] as PropType<TreeViewPropKey>,
+    default: 'label',
+  },
+  trackBy: {
```
asvae, line 33: Each prop should have demo. I have no way to know if any of these work at all.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,266 @@
...
+      <p>
+        {{ JSON.stringify(selectedNodes) }}
+      </p>
+      <br />
+      <va-tree-view
+        v-model:checked="selectedNodes"
+        :nodes="nodesWithIcons"
+        :color="selectedColor"
+        :selection-type="selectionType"
+        expand-all
```
asvae, line 41: Let's split these into multiple demos, so that each demo covers one piece of functionality. This one does too much.

asvae:
We need to fix demos before continuing forward with that. Too much untested code.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
rustem-nasyrov: If we want to use the selection model as default behavior, I'll change it to the v-model.

packages/docs/src/page-configs/ui-elements/tree-view/examples/CustomizableContent.vue
```diff
@@ -0,0 +1,89 @@
...
+          { id: 9, label: 'Common node' },
+        ],
+      },
+    ],
+  }),
+}
+</script>
+
+<style scoped>
+.customizable-content b {
```
asvae, line 82: Maybe we can make do with helpers alone?

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
-        </va-tree-category>
-      </va-tree-view>
+    <VbCard title="Default">
+      <va-tree-view :nodes="nodes" />
+    </VbCard>
+    <VbCard title="Stateless">
+      <va-tree-view :nodes="nodes" :stateful="false" />
+    </VbCard>
+    <VbCard title="Selectable">
+      <p>Default</p>
```
asvae, line 10: There should be orange selection on mouse click (only on keyboard navigation).

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
-              Node text
-            </va-tree-node>
-          </va-tree-category>
-        </va-tree-category>
-      </va-tree-view>
+    <VbCard title="Default">
+      <va-tree-view :nodes="nodes" />
+    </VbCard>
+    <VbCard title="Stateless">
+      <va-tree-view :nodes="nodes" :stateful="false" />
```
asvae, line 7: This demo doesn't seem to be working.

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
asvae: There should be orange selection on mouse click (only on keyboard navigation).

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+        :selection-type="selectionType"
+        selectable
+      />
+    </VbCard>
+    <VbCard title="Colored checkboxes">
+      <va-color-palette
+        v-model="selectedColor"
+        :palette="colorsPalette"
+      />
+      <va-tree-view
```
asvae, line 30: image

This demo could have all checkboxes set, so that we see instantly it's about color.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
-      </va-tree-view>
+    <VbCard title="Default">
+      <va-tree-view :nodes="nodes" />
+    </VbCard>
+    <VbCard title="Stateless">
+      <va-tree-view :nodes="nodes" :stateful="false" />
+    </VbCard>
+    <VbCard title="Selectable">
+      <p>Default</p>
+      <va-tree-view :nodes="nodes" selectable />
```
asvae, line 11: This component doesn't seem useful. Why would you have selectable component without v-model?

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+        v-model:checked="selectedNodes"
+        :selection-type="selectionType"
+        :nodes="nodes"
+        selectable
+      />
+    </VbCard>
+    <VbCard title="valueBy">
+      <p>{{ JSON.stringify(selectedNodesValueBy) }}</p>
+      <br />
+      <va-tree-view
```
asvae, line 65: image

image

Notice I have different v-model depending on where I click. That should not be the case.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+        selectable
+      />
+      <va-tree-view
+        v-model:checked="selectedNodes"
+        :selection-type="selectionType"
+        :nodes="nodes"
+        selectable
+      />
+    </VbCard>
+    <VbCard title="valueBy">
```
asvae, line 62: Empty text breaks items

image

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
-          is-open
-        >
-          <va-tree-node>
-            Node text
-          </va-tree-node>
-          <va-tree-node>
-            Node text
-          </va-tree-node>
-        </va-tree-category>
+    <VbCard title="trackBy">
```
asvae, line 78: image

This demo doesn't seem to test functionality

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+        track-by="trackId"
+      />
+    </VbCard>
+    <VbCard title="iconBy">
+      <va-tree-view
+        :nodes="nodesIconBy"
+        icon-by="image"
+      />
+    </VbCard>
+    <VbCard title="disabledBy">
```
asvae, line 90: That behaviour doesn't seem useful.

Disabled should apply to selection instead. Expanding should be completely free.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+    </VbCard>
+    <VbCard title="Expandable">
+      <p>Expand all</p>
+      <va-tree-view :nodes="nodes" expand-all />
+      <p>Expanded by</p>
+      <va-tree-view :nodes="nodes" expand-all />
+      <p>Expanded v-model: {{ JSON.stringify(expandedNodes) }}</p>
+      <va-tree-view v-model:expanded="expandedNodes" :nodes="nodes" />
+    </VbCard>
+    <VbCard title="Filter">
```
asvae, line 104: Can we expand Filter demo so that it covers selection as well? Would be great to check interactions of these.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,340 @@
...
+      <va-input
+        v-model="filterValue"
+        placeholder="Filter the tree view"
+        style="margin-bottom: 1rem;"
+      />
+      <va-checkbox v-model="isFilterCaseSensitive" label="Case sensitive" />
+      <va-tree-view
+        :nodes="filterableNodes"
+        :filter="filterValue"
+        :filter-method="customFilterMethod"
```
asvae, line 114: Is there any point having both `filterValue` and `customFilterMethod`? Can't we have just one prop for both?

packages/ui/src/components/va-option-list/VaOptionList.vue
```diff
@@ -123,7 +123,7 @@ export default defineComponent({
 
     const isDisabled = (option: SelectableOption) => props.disabled || getDisabled(option)
 
-    const reset = () => { valueComputed.value = undefined }
+    const reset = () => { valueComputed.value = null }
```
asvae, line 126: probably should be `undefined`, `null` is more of domain logic thing.

packages/ui/src/components/va-tree-view/hooks/useTreeView.ts
```diff
@@ -90,8 +51,8 @@ const useTreeView: UseTreeViewFunc = (props, emit) => {
     getNodeProperty,
   } = useTreeHelpers(props)
   const { nodes, expandAll, filter, filterMethod, textBy } = toRefs(props)
-  const expandedList = useNamedStateful<TypeModelValue>('expanded', props, emit, [])
-  const checkedList = useNamedStateful<TypeModelValue>('checked', props, emit, [])
+  const { valueComputed: expandedList } = useStateful<TypeModelValue>(props, emit, undefined, 'expanded')
```
asvae, line 54: I sugges we remove `defaultValue` in context of this PR.

You can create a new method for tree view, then deprecate existing method and plan full transition in another issue if it feels too complicated.

packages/ui/src/composables/useStateful.ts
```diff
@@ -23,16 +23,17 @@ export const useStatefulEmits = ['update:modelValue']
...
-  emit: (event: 'update:modelValue', newValue: T) => void,
-  defaultValue?: D,
+export function useStateful<VALUE extends any> (
+  props: StatefulProps,
+  emit: (event: string, newValue: VALUE) => void,
+  defaultValue = undefined as VALUE,
+  propName = 'modelValue',
 ) {
-  const valueState = ref(defaultValue === undefined ? props.modelValue : defaultValue) as Ref<T>
+  const valueState = ref(defaultValue === undefined ? props[propName] : defaultValue) as Ref<VALUE>
```
asvae, line 32: You can extract `props[propName]` to variable.

packages/ui/src/components/va-option-list/VaOptionList.vue
rustem-nasyrov: In the component props definition added support only for the given types:
`defaultValue: ({ type: [String, Number, Object, Array] as PropType<OptionListValue | null> })`
I thought I need to use `null` here.

packages/ui/src/components/va-tree-view/VaTreeView.demo.vue
```diff
@@ -1,59 +1,391 @@
...
+      <p>
+        expansion/selection shouldn't work:
+      </p>
+      <va-tree-view
+        :stateful="false"
+        :nodes="nodes"
+        selectable
+      />
+    </VbCard>
+    <VbCard title="Selectable">
```
asvae, line 30: image

Selection model demo has more stuff. We don't need both.

Issue #1729, opened 2022-05-05
useTextColor hook


m0ksem:
Let's have hook, that requires one prop:

- textColor  - component's text color

Returns

textColorComputed - returns textColor prop (if passed) or automatically calculated color

Must be used across components like VaButton, VaChip, VaCheckbox, VaSwitch, VaCollapse etc.

```ts
const { textColorComputed }  = useTextColor(props.background)
// or
const { textColorComputed }  = useTextColor(props.color) // component color
```

Text Color must be taken from [getCurrentInstance](https://github.com/epicmaxco/vuestic-ui/issues/1576)
```ts
export const useTextColor = (textBackground: string | Ref<string>) => {
   const { props } = getCurrentInstance()!

  const textColor = computed(() => props.textColor || getTextColor(unref(textBackground)))
   
   return { textColor }
}
```

Sadnessa:
Hello. Can I try this task, please?

aluarius:
See also https://github.com/epicmaxco/vuestic-ui/issues/1295

m0ksem:
Take a notice that outlined VaAlert has color from `color` prop by default, but I must change to text-color if it provided.

https://vuestic.dev/en/ui-elements/alert#different-styles
In this example we should be able to pass textColor to outlined VaAlert, so it can have yellow outline, but black text.

Derranion:
Should set `text color` for `VaAlert` when `border` `prop` is active.

Derranion:
> Take a notice that outlined VaAlert has color from `color` prop by default, but it must change to text-color if it provided.
> 
> https://vuestic.dev/en/ui-elements/alert#different-styles In this example we should be able to pass textColor to outlined VaAlert, so it can have yellow outline, but black text.

Hey @m0ksem
This looks like another separate task though - to add `textColor` prop for `VaAlert` and related docs.
Because unlike components that were listed in the initial description it does not have this prop.

m0ksem:
I'm thinking how can we deal with components like date picker. It has a transparent background, so we are not able to calculate text color automatically. 

As a workaround, we can go through all component parent elements until we find element with background style set. Handle reactivity with MutationObserver. Not sure how it will impact performance.

Derranion:
We decided to create separate tasks to add `textColor` prop to `VaAlert` and to improve automatically selected color algorithm.

By the way @m0ksem here are some examples of parents traversing algorithms to get the real `background-color`, could be useful:
[...]

PR #1736, opened 2022-05-05
useElementBackground hook

m0ksem:
### Why do we need this PR?
What hook does is simply go thought all parents and returns first parent bg.

We can use it for useTextColor hook. We can not chose what text color to use if we don't know component background. We have a lot of transparent components as VaDatePicker, VaTimePicker, VaSidebarItem.

### Demo
I change vue-book theme to use dark colors. Vuestic don't even know about these colors.
I also update VaDatePicker to use this hook and getTextColor method. 
Don't look on buttons, I was lazy to pass color prop for them. 
image

### Perfomance
For me in docs it took 4ms to calculate color. While vue internal function can be called for 66ms. Tested on docs.
image
image

### TODO:
- Care about colors with opacity. We need to mix them until we find color with opacity 1.
- Make me be sure this thing doesn't break performance 🤣


m0ksem:
Added transparent colors handling and created new demo in docs sidebar.

You can play with sidebar background color (passed as HSLA) and active color in `packages/docs/src/components/sidebar/Sidebar.vue`
image
image

m0ksem:
For people who use position absolute and other CSS tricks, we have `textColor` prop 😊.
For example, it would be impossible to get background of Modal with transparent bg...

packages/ui/src/composables/useElementBackground/useElementBackground.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const rgba2hex = (rgba: number[]) => `#${rgba.map((n: number, i: number) => (i === 3 ? Math.round(n * 255) : n).toString(16).padStart(2, '0').replace('NaN', '')).join('')}`
+
+const isTransparent = (color: string) => color === 'rgba(0, 0, 0, 0)'
+
+const mixColors = (color1Array: ColorArray, color2Array: ColorArray, weight: number) => {
+  const newColor = []
+
+  for (let i = 0; i < 4; i++) {
+    newColor.push(Math.round((color1Array[i]) * (1 - weight) + (color2Array[i]) * weight))
```
RVitaly1978, line 18: The `color1Array ` or `color2Array` arrays for a `rgb` color can be `[number, number, number]`. In this case `newColor` array will by `[number, number, number, undefined]`

RVitaly1978 (approved):
Should be updated after merge PR #1729

aluarius (approved):

packages/ui/src/composables/useElementBackground/useElementBackground.ts
m0ksem: ColorArray can not be [number, number, number, undefined]

```ts
export type ColorArray = [number, number, number, number]
```

m0ksem:
List of components with transparent bg, where we want determinate text color.

- [x] VaSidebarItem
image
(Also react on transparent hover colors)

- [x] VaDatePicker
image
image

- [x] VaTimePicker
image
image


- [ ] VaBreadcrumbs
- [ ] VaFileUpload
- [ ] VaIcon
- [ ] VaInput (validation)
- [ ] VaRadio (labeles)
- [ ] VaSwitch (label)
- [ ] VaTimeline (unused and can be skipped)

m0ksem:
A pair of these two looks kinda cool.

image

m0ksem:
Moved to Dark themes PR.

Issue #1681, opened 2022-04-19
GUI color picker to VaColorInput


LighthouseKeeperYN:
VaColorInput allows input only as text value. Would be nice to invoke native color input when clicking on color circle.

(the video is just a demo, it's not currently implemented)
https://user-images.githubusercontent.com/47790197/164025877-2387b9bc-008d-404c-b42b-0777fa178f63.mp4

asvae:
Yeah, component looks super weird. As if something is broken.

As colors are clickable, but do nothing. Demos also look pretty broken.

PR #1738, opened 2022-05-06
[#1681] color picker added, refactoring

aluarius:
close: #1681 

## Description
Added HTML5 native color picker for color indicator. Updates to related components. Refactoring.

### DIFF
--- a/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
+++ b/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
@@ -20,7 +20,7 @@ import { useStateful, useStatefulProps, useStatefulEmits } from '../../composabl
 
 export default defineComponent({
   name: 'VaColorIndicator',
-  emits: [...useStatefulEmits, 'on:click'],
+  emits: useStatefulEmits,
   props: {
     ...useStatefulProps,
     modelValue: { type: Boolean, default: null },
@@ -43,10 +43,7 @@ export default defineComponent({
       'va-color-indicator--hoverable': valueComputed.value !== undefined,
     }))
 
-    const handleClick = () => {
-      valueComputed.value = !valueComputed.value
-      emit('on:click')
-    }
+    const handleClick = () => { valueComputed.value = !valueComputed.value }
 
     return {
       valueComputed,
--- a/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
+++ b/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     class="va-color-indicator"
-    @click="handleClick"
+    @click="toggleModelValue"
     :class="computedClass"
     :style="computedStyle"
   >
@@ -43,13 +43,13 @@ export default defineComponent({
       'va-color-indicator--hoverable': valueComputed.value !== undefined,
     }))
 
-    const handleClick = () => { valueComputed.value = !valueComputed.value }
+    const toggleModelValue = () => { valueComputed.value = !valueComputed.value }
 
     return {
       valueComputed,
       computedStyle,
       computedClass,
-      handleClick,
+      toggleModelValue,
     }
   },
 })
--- a/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
+++ b/packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
@@ -1,68 +1,58 @@
 <template>
-  <label :for="picker && inputId"
+  <div
     class="va-color-indicator"
-    @click="valueComputed = !valueComputed"
+    @click="handleClick"
     :class="computedClass"
     :style="computedStyle"
   >
-    <span
+    <div
       class="va-color-indicator__core"
       :style="computedStyle"
     />
-    <input
-      type="color"
-      class="visually-hidden"
-      :id="inputId"
-      v-model="colorPicker" />
-  </label>
+  </div>
 </template>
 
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -1,50 +1,68 @@
...
+    })
+
+    const inputId = computed(() => generateUniqueId())
+
     const colorComputed = computed(() => getColor(props.color))
 
-    const computedStyle = computed(() => ({ borderRadius: props.square ? '0px' : '50%' }))
+    const computedStyle = computed(() => ({
+      borderRadius: props.square ? '0px' : '50%',
+      backgroundColor: unref(colorComputed),
```
LighthouseKeeperYN, line 52: Why unref instead of computed.value?

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -1,50 +1,68 @@
...
+    }))
 
     const computedClass = computed(() => ({
-      'va-color-indicator--selected': valueComputed.value,
-      'va-color-indicator--hoverable': valueComputed.value !== undefined,
+      'va-color-indicator--selected': unref(valueComputed),
+      'va-color-indicator--hoverable': unref(valueComputed) !== undefined,
```
LighthouseKeeperYN, line 57: Same

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -1,50 +1,68 @@
 <template>
-  <div
+  <label :for="picker && inputId"
```
LighthouseKeeperYN, line 2:
```suggestion
  <label :for="inputId"
```

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -1,50 +1,68 @@
...
       class="va-color-indicator__core"
-      :style="{ ...computedStyle, backgroundColor: colorComputed }"
+      :style="computedStyle"
     />
-  </div>
+    <input
+      type="color"
+      class="visually-hidden"
+      :id="inputId"
+      v-model="colorPicker" />
```
LighthouseKeeperYN, line 16:
```suggestion
      :disabled="!picker"
      v-model="colorPicker" />
```

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -74,6 +92,7 @@ export default defineComponent({
   }
 
   &__core {
+    display: block;
     transition: transform 0.1s linear;
     vertical-align: baseline;
```
LighthouseKeeperYN, line 97: vertical-align is ignored when display: block. Remove it if it's not needed.

LighthouseKeeperYN (changes requested):

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
aluarius: Was set cause of div (with display block as default) -> span (with display inline as default). So, it's seems to be ignored before.

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
aluarius: Thought unref to be a pretty sugar for such situations. But after second look i'm agree, that `isRef(val) ? val.value : val` is excessive for such situations.

LighthouseKeeperYN (approved):

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -1,50 +1,58 @@
...
+    }))
 
     const computedClass = computed(() => ({
       'va-color-indicator--selected': valueComputed.value,
       'va-color-indicator--hoverable': valueComputed.value !== undefined,
     }))
 
+    const handleClick = () => {
+      valueComputed.value = !valueComputed.value
+      emit('on:click')
```
RVitaly1978, line 48: Why do we need this emit? After all, we are already throwing emit `update:modelValue`

packages/ui/src/components/va-color-input/VaColorInput.vue
```diff
@@ -4,18 +4,25 @@
       class="va-color-input__dot"
       :color="valueComputed"
       :indicator="indicator"
+      v-model:color="valueComputed"
+      v-on:click="callPickerDialog"
```
RVitaly1978, line 8:
```suggestion
      stateful
      @update:modelValue="callPickerDialog($event)"
```

packages/ui/src/components/va-color-input/VaColorInput.vue
```diff
@@ -41,7 +48,14 @@ export default defineComponent({
   setup: (props, { emit }) => {
     const { valueComputed } = useStateful(props, emit)
 
-    return { valueComputed }
+    const colorPicker = ref<HTMLInputElement>()
+    const callPickerDialog = () => {
+      if (props.disabled) { return }
```
RVitaly1978:
```suggestion
    const callPickerDialog = (isActive: boolean) => {
      if (props.disabled || !isActive) { return }
```

RVitaly1978 (changes requested):

RVitaly1978 (approved):

Issue #1710, opened 2022-04-26
Investigate best practices of composables


Derranion:
**Subject**:

- [ ] We want to find out if it's possible to make our `composables` more explicit and intuitive to use
- [ ] Add a conclusion, create a new task to implement this, if necessary

**Concern example**:

`useValidation` composable utilizes and exports `useFocus` composable inside of it.
In practice this means that our components will need to import `useFocus` logic from the `useValidation` in some cases.
This is a bit weird and not transparent, also may lead to unexpected behavior if component uses this composable already.
```
// related code example from VaSelect
const {
  isFocused, // this is from useFocus composable
  validate,
  computedError,
  computedErrorMessages,
} = useValidation(props, emit, () => reset(), () => focus())
```

**Additional info**:
You could find some good practices here:
https://vueuse.org/core/
As an easy solution we could just explicitly pass needed variables as an additional argument to the composable.
Find out what is the best approach to this issue.

PR #1740, opened 2022-05-06
[#1710] composables refactoring

aluarius:
close: #1710 

## Description
As discussed during the grooming, it's a bad practice to reexport composables or import them as args through components - their logic should be incapsulated as much as possible.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/VaInput.vue
+++ b/packages/ui/src/components/va-input/VaInput.vue
@@ -143,9 +143,9 @@ export default defineComponent({
   inheritAttrs: false,
 
   setup (props, { emit, attrs, slots }) {
-    const input = ref<HTMLInputElement | typeof VaTextarea | undefined>()
+    const input = ref<HTMLInputElement | HTMLTextAreaElement | undefined>()
 
-    const { isFocused } = useFocus()
+    const { isFocused } = useFocus(input, emit)
 
     const reset = () => {
       emit('update:modelValue', props.clearValue)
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -144,6 +145,8 @@ export default defineComponent({
   setup (props, { emit, attrs, slots }) {
     const input = ref<HTMLInputElement | typeof VaTextarea | undefined>()
 
+    const { isFocused } = useFocus()
```
RVitaly1978, line 148: In this case, `isFocused` will always be `false`.
Otherwise, you either need to pass an HTML element to `useFocus` or export `onFocus` and `onBlur` from `useFocus` to manually control `isFocused`

RVitaly1978: And the same in the cases below.

RVitaly1978 (changes requested):

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -142,7 +143,9 @@ export default defineComponent({
   inheritAttrs: false,
 
   setup (props, { emit, attrs, slots }) {
-    const input = ref<HTMLInputElement | typeof VaTextarea | undefined>()
+    const input = ref<HTMLInputElement | HTMLTextAreaElement | undefined>()
```
RVitaly1978, line 146: `VaTextarea` is not `HTMLTextAreaElement`

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -142,7 +143,9 @@ export default defineComponent({
...
+
+    const { isFocused } = useFocus(input, emit)
```
RVitaly1978, line 148: I think it would be better to export `onFocus` and `onBlur` from `useFocus` and use them in `inputEvents` (combined with `onFocus` and `onBlur` on lines 191 and 196 below)

RVitaly1978 (approved):

PR #1745, opened 2022-05-08
Remove unused emits.

### LINKED ISSUE #1733: Remove unused emits
Some of the components have legacy \ irrelevant `events` listed in our docs and not used anymore in the component itself.

For example `VaFileUpload` has `input` `Event` described in the documentation but we don't have this `emit` in the component:

unusedEmitsExample

- [x] Find similar cases and remove them

xiongmao86:
## Description
Some of the events are not emitted, and not used any more. This pr try to remove them from docs.
Fixes #1733

## Types of changes


packages/docs/src/page-configs/ui-elements/button/api-options.ts
```diff
@@ -4,9 +4,6 @@ export default defineManualApi({
   slots: {
     default: { },
   },
-  events: {
-    click: { types: '`() => Event`' },
```
xiongmao86, line 8: `VaButton` doesn't emit click, I think it is a common need, is this a bug of `VaButton`, or am I making mistakes with this sort of events?

packages/docs/src/page-configs/ui-elements/backtop/api-options.ts
```diff
@@ -1,11 +1,6 @@
 import { defineManualApi } from '../../../components/DocsApi/ManualApiOptions'
 
 export default defineManualApi({
-  events: {
-    click: {
-      types: 'Event',
```
xiongmao86, line 6: `VaBacktop` has use this event. So there are some event didn't emitted, but are used in demo. What kind of event have this characteristic, currently I am looking at relevant files such as `VaBacktop.vue` and `VaBacktop.demo.vue`, but what if event don't have example in `*.demo.vue`?

packages/docs/src/page-configs/ui-elements/button/api-options.ts
m0ksem: VaButton doesn't emit "click" as custom component event, its true. 
When user type:
```vue
<VaButton @click="doSmth" />
```
he actually listens to DOM event, which propagate from `button` tag.

So hypothetical user can make this code which work okay.
```vue
<VaButton @dblclick="doSmth" @contextmenu="doSmth2" />
```

So there is no bug, you can click on any vuestic component.

So, this is a lie. We do not emit any clicks 😁
image

I think we can safety remove it from docs. 
Or we can add DOM events section under spoiler, so no one will be confused with missing events. And add all DOM Mouse events: https://developer.mozilla.org/en-US/docs/Web/Events
What do you think?

packages/docs/src/page-configs/ui-elements/backtop/api-options.ts
m0ksem: Same. It is DOM event, not vue emit.

m0ksem (approved):

packages/docs/src/page-configs/ui-elements/button/api-options.ts
m0ksem: /cc @asvae

xiongmao86: Thanks, @m0ksem, I'll wait for your discussion.

Derranion: Yes, it will be very confusing for the users to remove this event from docs.

Derranion (approved):
Thanks a lot for your help.

packages/docs/src/page-configs/ui-elements/button/api-options.ts
Derranion: Hey @xiongmao86 

We will handle it in the other task to not bother you :)

Would be amazing if you could just add to such events description in the docs something like this in the end.

For example for this click `Event` description will be:

```
Emitted when user clicks on button
(native DOM event)
```

Thanks again 👍

xiongmao86: Thanks, @Derranion. I'll do what you said.

xiongmao86:
This pr is ready for review.

xiongmao86:
Thanks, @m0ksem and @Derranion.

PR #1746, opened 2022-05-09
fix: dismiss only last modal on esc

### LINKED ISSUE #1651: `Esc` closes all nested `VaModal`
**Vuestic-ui version:** 1.3.5

**Steps to reproduce**
1. go to https://vuestic.dev/en/ui-elements/modal#nested-modals
2. open first and second modals
3. press `esc`

should be closed only second (the last opened) modal

both modals close
chrome-capture

**Other information**

LighthouseKeeperYN:
close #1651

Feedback appreciated. This works but I wish I could figure out how to fix this in less costyl fashion.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/va-modal/VaModal.vue
@@ -11,6 +11,7 @@
     <teleport :to="attachElement" :disabled="$props.disableAttachment">
       <modal-element
         name="va-modal"
+        ref="modal"
         :isTransition="!$props.withoutTransitions"
         appear
         :duration="300"
@@ -34,7 +35,6 @@
               class="va-modal__dialog"
               :class="computedClass"
               :style="computedDialogStyle"
-              ref="modal"
             >
               <va-icon
                 v-if="$props.fullscreen"
@@ -164,6 +164,7 @@ export default defineComponent({
   setup (props, { emit }) {
     const { getTextColor, getColor } = useColors()
     const rootElement = ref<HTMLElement>()
+    const modal = ref<{ $el: HTMLElement }>()
     const { valueComputed } = useStateful(props, emit)
 
     const computedClass = computed(() => ({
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -213,7 +214,9 @@ export default defineComponent({
     const onAfterLeaveTransition = (el: HTMLElement) => emit('close', el)
 
     const listenKeyUp = (e: KeyboardEvent) => {
-      if (e.code === 'Escape' && !props.noEscDismiss && !props.noDismiss) {
+      const isLastNestedModal = !modal.value?.$el.nextElementSibling
```
RVitaly1978, line 217: in the `body` after the last modal there can be one or more dropdowns (as well as user elements)

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -11,6 +11,7 @@
     <teleport :to="attachElement" :disabled="$props.disableAttachment">
       <modal-element
         name="va-modal"
+        ref="modal"
```
RVitaly1978, line 14: Maybe
```suggestion
```
and below in 20th line: `<div class="va-modal" v-if="valueComputed" ref="modal">`

RVitaly1978 (changes requested):

packages/ui/src/components/va-modal/VaModal.vue
LighthouseKeeperYN: I wonder If this could actually be a feature? So dropdown get dismissed on esc before the popup.

RVitaly1978: Components like `VaToast` do not respond (are not removed from the `body`) when the `esc` key is pressed.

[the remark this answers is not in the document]
Derranion: Here is complete solution:

```
    const listenKeyUp = (e: any) => {
      e.modalsCounter = e.modalsCounter ? e.modalsCounter + 1 : 1
      const modalNumber = e.modalsCounter

      const hideModal = () => {
        if (e.code === 'Escape' && !props.noEscDismiss && !props.noDismiss && e.modalsCounter === modalNumber) {
          cancel()
        }
      }

      setTimeout(hideModal, 30)
    }
```

Derranion (approved):
Glad that it helped :)

RVitaly1978 (approved):

rustem-nasyrov (approved):

Issue #1763, opened 2022-05-12
TS register components globally.


m0ksem:
We can extend vue's GlobalComponents interface like this: 

vuestic-global-components.ts
```ts
declare module 'vue' {
  export interface GlobalComponents extends VuesticComponents {}
}
```

It will register component in IDE.

How to deal with treeshaking? 

For example, I want only VaButton, VaInput, then I don't need to see VaProgressCircle suggestions. 

We can move vuestic-global-components.ts out of default vuestic exports. We will need to ask user to `import 'vuestic-ui/types/components'`.

Dealing with tree-shaking will force user to manually implement this module  declaration.

We can make interface like this:
```ts
declare module 'vue' {
  export interface GlobalComponents extends VuesticComponents<'VaInput' | 'VaButton'> {}
}
```

So, do we even need any additional file for types, or we will ask to manually add three line snippet from Docs?

We will automatically register all components in nuxt.

m0ksem:
So we decided to register all components globally and not think about tree-shaking at all for better DX.

PR #1764, opened 2022-05-12
Feat/component global types

m0ksem:
Closes https://github.com/epicmaxco/vuestic-ui/issues/1763

Made a way to register components, do not register them or register only specific components globally.

Registered them for docs and nuxt module. Vue book uses tree-shaking, so we do not declare global components for vue-book package.

### DIFF
--- a/packages/ui/src/vuestic-plugin/global-components.ts
+++ b/packages/ui/src/vuestic-plugin/global-components.ts
@@ -13,13 +13,14 @@ type VuesticComponentNames = keyof VuesticComponentsMap
 /**
  * Names of Vuestic components that must be accessable globally.
  *
- * @exmaple
+ * @example
  * This will register all vuestic components globally:
  * ```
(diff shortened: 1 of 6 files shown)


aluarius (approved):

packages/ui/src/vuestic-plugin/global-components.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/** List of all exported components */
+type VuesticComponentsMap = Map<typeof vuesticComponents>
+/** Component names */
+type VuesticComponentNames = keyof VuesticComponentsMap
+
+/**
+ * Names of Vuestic components that must be accessable globally.
+ *
+ * @exmaple
```
RVitaly1978, line 16:
```suggestion
 * @example
```

RVitaly1978 (approved):

PR #1775, opened 2022-05-14
[Proposal] defineComposable

m0ksem:
This is a type safe solution, allow us and newbie easily read and understand composables.

Example:

```ts
const useTextColor = defineComposable(({ props, emit }) => (background: string) => {
  emit('initBg', background)

  return computed(() => props.textColor ? getColor(props.textColor) : getTextColor(background))
}, { props: { textColor: String }, emits: ['initBg'] })
```

```ts
export default defineComponent({
  props: {
    ...useTextColor.$props,
  },
  emits: [...useTextColor.$emits],

  setup() {
    return {
      ...useTextColor(),
    }
  }
})
```

## Plus
- First of all, this will not break emit or prop type (like useStateful makes):
VaAlert
image
- It type composable props and emit in composable. By default, getCurrentInstance() returns any...
- It easier to pass emits and props to component from composable from one thing, instead of importing useTextColorProps etc.

## Minus
- Something specific, not documented in vue docs. But, if we have agreement on `useTextColorProps`, let's better have it documented with Typescript. 
- Has additional wrapper (perfomance?), but who cares? getCurrentInstace just access to one stored let.

## Should be composables context free?
Yes.. But we don't want pass props every time to composable. Don't we? For example useTextColor works like a mixin, execpt we pass background color manually. Copying props.textColor every time will make code harder to read and refactor.

### DIFF
--- a/packages/ui/src/composables/useStateful.ts
+++ b/packages/ui/src/composables/useStateful.ts
@@ -1,4 +1,41 @@
 import { ref, computed, PropType, Ref } from 'vue'
+import { defineComposable } from '../types/defineComposable'
+
+// TODO: Remove this example
+const useStaateful2 = defineComposable(
+  ({ props, emit }) => <D>(initialValue: D) => {
+    const valueState = ref(initialValue === undefined ? props.modelValue : initialValue)
+
+    const valueComputed = computed({
+      get () {
+        if (props.stateful) {
+          return valueState.value
+        }
+        return props.modelValue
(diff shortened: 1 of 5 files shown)


m0ksem:
Looks like getCurrentInstance has been removed from docs. That mean we mustn't use it anymore...

m0ksem:
Gonna drop this, because it only adds overhead and no useful as much.

PR #1778, opened 2022-05-14
Fix #1758: `VaSelect` actions and navigation

### LINKED ISSUE #1758: `VaSelect` keyboard navigation is completely broken
**Vuestic-ui version:**
`1.3.5` same in `1.4.0`

- [x] Fix all navigation buttons

**Steps to reproduce**
Open any `VaSelect` component and try to navigate or close it using keyboard buttons.

It should be fully usable just by keyboard.

Nothing works - `esc`, `arrows`, `enter` etc.

**Other information**
Pretty sure that it was caused by the recent changes because it worked previously.

Derranion:
## Description
Close #1758

All `VaSelect` actions and keyboard navigation were completely broken, now it is fixed.

## Details:

- Keyboard navigation works
- UX improved - `VaSelect` will scroll to and focus last selected option
- All `actions` that were using `optionList` functions work now - `hintedSearch` etc.

</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.vue
+++ b/packages/ui/src/components/va-select/VaSelect.vue
@@ -16,6 +16,7 @@
     @keydown.up.stop.prevent="showDropdown()"
     @keydown.down.stop.prevent="showDropdown()"
     @keydown.space.stop.prevent="showDropdown()"
+    @keydown.enter.stop.prevent="showDropdown()"
     @click.prevent="onSelectClick()"
   >
     <template #anchor>
@@ -515,7 +516,7 @@ export default defineComponent({
 
     const focusOptionList = () => {
       optionList.value?.focus()
-      optionList.value?.hoverFirstOption()
+      !props.modelValue && optionList.value?.hoverFirstOption()
     }
 
     const focusSearchOrOptions = () => nextTick(() => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -16,6 +16,7 @@
     @keydown.up.stop.prevent="showDropdown()"
     @keydown.down.stop.prevent="showDropdown()"
     @keydown.space.stop.prevent="showDropdown()"
+    @keydown.enter.stop.prevent="showDropdown()"
```
RVitaly1978, line 19: The native html `select` element has no such option. Do we want to do it?

RVitaly1978 (approved):

packages/ui/src/components/va-select/VaSelect.vue
aluarius: Native select option selection also works for `enter`.

aluarius (approved):

packages/ui/src/components/va-select/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -221,9 +221,12 @@ export default defineComponent({
       hoverNextOption,
       hoverFirstOption,
       focus,
+      scrollToOption,
```
m0ksem, line 224: Expose must work now :) after we fixed withTransportConfig.

packages/ui/src/components/va-select/VaSelect.vue
m0ksem: Enter usually used for submit action in form. I would rather just use space to open select. Pressing enter on any component could submit my form.

Also, up and down buttons often used not for dropdown opening, but switching between options without opening dropdown.

m0ksem (approved):
So this PR fixes the bug. We can discuss correct component behavior later, in different task.

PR #1780, opened 2022-05-15
Fix/versions correct

m0ksem:
Sync package versions.

I also updated our yarn.lock file, because it had a lot of old packages for some reason. (For example @vue/sfc-compiller@3.0.0 instead of @vue/sfc-compiler@3.22.0)

Actually, you can see how junk we had in `yarn.lock`
image


asvae (approved):
🤗

PR #1815, opened 2022-05-19
Fix#1814/outline button text color

### LINKED ISSUE #1814: text-color not working on outline button
**Vuestic-ui version:** 1.4.0

**Steps to reproduce**
```
<va-button outline text-color="#ff1100" class="mr-4">Red text</va-button>
```

Text should turn red

Text color stays at it's default

**Other information**
I got the code directly from your documentation that even shows the bug and gives you a link to prove the bug in sandbox. If this is a known issue it should be stated in the docs.

m0ksem:
closes #1814

Before:
image
firefox_FfbUZY9wmb

After:
image
image

Here  is also some issue with sidebar, but I think we will fix it with https://github.com/epicmaxco/vuestic-ui/pull/1736 in different task.

### DIFF
--- a/packages/ui/src/composables/useTextColor.ts
+++ b/packages/ui/src/composables/useTextColor.ts
@@ -6,14 +6,12 @@ export const useTextColor = (componentColor: string | Ref<string>, isTransparent
   const { getColor, getTextColor } = useColors()
 
   const textColorComputed = computed(() => {
-    const hexColor = getColor(unref(componentColor))
     const textColorProp = props.textColor as string
 
-    if (unref(isTransparent)) {
-      return hexColor
-    }
+    if (textColorProp) { return getColor(textColorProp) }
+    const componentColorHex = getColor(unref(componentColor))
 
-    return props.textColor ? getColor(textColorProp) : getColor(getTextColor(hexColor))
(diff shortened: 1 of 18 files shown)


packages/ui/src/composables/useTextColor.ts
```diff
@@ -1,19 +1,25 @@
...
+  textColor: string,
+  color?: string
+}
+
+/**
+ * @param componentColor component color. By default `props.color`.
+ * @param isTransparent if transparent will return component color as text color.
+ * @returns Computed text color based on component's color if `props.textColor` if provided.
+ */
+export const useTextColor = (componentColor?: Ref<string> | undefined, isTransparent: boolean | Ref<boolean> = false) => {
```
RVitaly1978, line 14:
```suggestion
export const useTextColor = (componentColor?: Ref<string> | string | undefined, isTransparent: boolean | Ref<boolean> = false) => {
```

RVitaly1978 (approved):

packages/ui/src/composables/useTextColor.ts
m0ksem: I think, if we remove string, this will force use to use reactive variable. I don't believe we would have a component with static background.

aluarius (approved):

Issue #1812, opened 2022-05-19
mergeGlobalConfig doesn't work in setup function


m0ksem:
But other composables work fine.

m0ksem:
It because mergeGlobalConfig uses useGlobalConfig, and it can not be used in function and other callbacks.

I would suggest removing all this functions.

Derranion:
It doesn't work even if we use default `<script>`

Stopped working after `1.3.5` version

PR #1817, opened 2022-05-19
Fix#1812/remove confusing config aliases

### LINKED ISSUE #1573: Making global config independed per vue app.  Deprecating global-config aliases
In global config we have global ref which can cause memory leaks on SSR. Also, it makes impossible to use global config in different apps on page.

Let's have `createGlobalConfig` fabric, that provides `globalConfig` ref when plugin registered. 
`useGlobalConfig` then must be used only inside setup function
`useColors` must depend on `useGlobalConfig`

Aliases make user to believe that `getColor`, `setGlobalConfig` etc. can be used outside setup function. That's wrong, because provide or `getCurrentInstance` not working outside setup function.

`useGlobalConfig` methods should be added to app global context as well. Don't forget about Options API. 

Also, having only two hooks makes us easy to document them using JSDocs.

This can be uncomfortable, but it looks like a correct way.

### LINKED ISSUE #1816: Remove `vuestic-ui/src/main` from docs examples
image

Before we move to our docs to nuxt, let's just replace ugly alias to just `vuestic-ui` before it renders on page.

m0ksem:
closes [#1816](https://github.com/epicmaxco/vuestic-ui/issues/1812)
closes https://github.com/epicmaxco/vuestic-ui/issues/1809
close #1573

According to Composition API and [Multi Application Instances](https://vuejs.org/guide/essentials/application.html#multiple-application-instances) features we can not store globally ref with global config. For that, we use provide/inject. So, getGlobalConfig can not be used outside setup context anymore. I propose to remove this confusing functions and allow user to use the same methods from composables. 

LMK, if you think that we need to deprecate this hooks, instead of removing them. 

This PR also fixes some bugs in demo, for example.

### DIFF
--- a/packages/docs/src/components/DocsAnchor.vue
+++ b/packages/docs/src/components/DocsAnchor.vue
@@ -9,16 +9,20 @@
 import { defineComponent, PropType, computed } from 'vue'
 import { kebabCase } from 'lodash'
 import { TranslationString } from '../components/DocsApi/ManualApiOptions'
-import { getColors } from '../../../ui/src/main'
+import { useColors } from '../../../ui/src/main'
 
 export default defineComponent({
   name: 'DocsAnchor',
   props: {
     text: { type: String as PropType<TranslationString> },
   },
-  setup: (props) => ({
-    anchor: computed(() => kebabCase(props.text)),
-    colors: computed(() => getColors()),
-  }),
+  setup: (props) => {
+    const { getColors } = useColors()
+
+    return {
+      anchor: computed(() => kebabCase(props.text)),
+      colors: computed(() => getColors()),
+    }
+  },
 })
 </script>
--- a/packages/docs/src/components/DocsAnchor.vue
+++ b/packages/docs/src/components/DocsAnchor.vue
@@ -21,7 +21,7 @@ export default defineComponent({
 
     return {
       anchor: computed(() => kebabCase(props.text)),
-      colors: computed(() => getColors()),
+      colors: computed(getColors),
     }
   },
 })
--- a/packages/docs/src/components/header/components/HeaderSelector.vue
+++ b/packages/docs/src/components/header/components/HeaderSelector.vue
@@ -19,7 +19,8 @@
 
 import VaIconMenu from '../../iconset/VaIconMenu.vue'
 import VaIconMenuCollapsed from '../../iconset/VaIconMenuCollapsed.vue'
-import { getColors } from '../../../../../ui/src/main'
+import { useColors } from '../../../../../ui/src/main'
+import { computed } from 'vue'
 
 export default {
   name: 'DocsHeaderSelector',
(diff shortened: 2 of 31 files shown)


packages/docs/src/components/header/components/HeaderSelector.vue
```diff
@@ -34,10 +35,12 @@ export default {
...
 
-  computed: {
-    colors () {
-      return getColors()
-    },
+  setup () {
+    const { getColors } = useColors()
+
+    return {
+      colors: computed(() => getColors()),
```
RVitaly1978, line 42:
```suggestion
      colors: computed(getColors),
```

packages/docs/src/components/DocsAnchor.vue
```diff
@@ -9,16 +9,20 @@
...
-  setup: (props) => ({
-    anchor: computed(() => kebabCase(props.text)),
-    colors: computed(() => getColors()),
-  }),
+  setup: (props) => {
+    const { getColors } = useColors()
+
+    return {
+      anchor: computed(() => kebabCase(props.text)),
+      colors: computed(() => getColors()),
```
RVitaly1978, line 24:
```suggestion
      colors: computed(getColors),
```

packages/docs/src/page-configs/ui-elements/hover/examples/Slot.vue
```diff
@@ -8,14 +8,10 @@
   </va-hover>
 </template>
 
-<script>
-import { getColors } from 'vuestic-ui/src/services/color-config/color-config'
+<script setup>
+import { computed } from 'vue'
+import { useColors } from 'vuestic-ui/src/services/color-config/color-config'
```
RVitaly1978, line 13:
```suggestion
import { useColors } from 'vuestic-ui/src/main'
```

packages/ui/src/composables/useColor.ts
```diff
@@ -14,22 +14,23 @@ export const useColorProps = {
   },
 }
 
+/** @deprecated Use useColors instead */
 export function useColor (props: any) {
   const hasColorTheme = true
```
RVitaly1978, line 19:
```suggestion
```

packages/ui/src/services/global-config/global-config.ts
```diff
@@ -42,30 +42,19 @@ export const createGlobalConfig = () => {
...
+export function useGlobalConfigSafe () {
+  return inject<ProvidedGlobalConfig>(GLOBAL_CONFIG)
+}
+
 export function useGlobalConfig () {
-  const globalConfig = inject<ProvidedGlobalConfig>(
-    GLOBAL_CONFIG,
-    getCurrentInstance()?.appContext.config.globalProperties.$vaConfig,
-  )
+  const globalConfig = inject<ProvidedGlobalConfig>(GLOBAL_CONFIG)
```
RVitaly1978, line 51: Let's rename this variable (e.g. `globalConfigInstance`) because `ProvidedGlobalConfig` also contains `globalConfig`

RVitaly1978 (approved):

Issue #1275, opened 2021-11-26
Paginator jumping when size of page number changing


ishpartko:
* **What is the expected behavior?**
Paginator component not jumping when size of page number changing

* **What is the current behavior?**
![Анимация](https://user-images.githubusercontent.com/27618114/143616405-bdb9bdef-74c5-483f-a261-efae49987d83.gif)

asvae:
Would be great to investigate if fix is needed even.

asvae:
We can fix by hardcoding min width to 2 or 3 symbols

PR #1833, opened 2022-05-20
Prevent pagination numeric buttons changin its width if content is long.

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1275

firefox_cBAzqD40GP


m0ksem:
I also found the issue if you focus button, then alt+tab and focus another button - previous focus will stay.

Looks like issue in useKeyboardFocus...

RVitaly1978 (approved):

PR #1836, opened 2022-05-21
Fix docs colors.

m0ksem:
# Fixes
Before:
image
image
image

After:
image
image
image

# Question

@xx13, which color we want to use on links in general? info or primary?
Primary ():
image
Info (more logical to use):
image


xx13:
let's drop links on `#` only and have it in `primary`

PR #1840, opened 2022-05-22
Fix #1785: VaDateInput keyboard navigation, click and focus events

### LINKED ISSUE #1785: VaDateInput keyboard navigation
http://localhost:8080/en/ui-elements/date-input

Can't be open with space, unable to move in dropdown...

Must be simular to https://github.com/epicmaxco/vuestic-ui/issues/1783

Derranion:
## Description
Closes #1785

- [x] Keyboard navigation works now
- [X] `VaDateInput` focuses `VaDatePicker` by default
- [X] Manual input behavior corrected
- [x] Inner `VaDropdown` usage is fixed

</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-date-input/VaDateInput.vue
+++ b/packages/ui/src/components/va-date-input/VaDateInput.vue
@@ -3,7 +3,9 @@
     <va-dropdown
       v-model="isOpenSync"
       :offset="[2, 0]"
-      :close-on-content-click="false"
+      :close-on-content-click=false
+      :stateful=false
+      trigger="none"
       :disabled="disabled"
       anchorSelector=".va-input-wrapper__input"
     >
@@ -16,6 +18,10 @@
             v-on="inputListeners"
             :model-value="valueText"
             @change="onInputTextChanged"
+            @click="toggleDropdown()"
+            @keydown.enter.stop="showAndFocus"
+            @keydown.space.stop="showAndFocus"
+            @keydown.esc.stop.prevent="hideAndFocus()"
           >
             <template
               v-for="name in filterSlots"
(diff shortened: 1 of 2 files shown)


rustem-nasyrov (approved):

RVitaly1978 (approved):

packages/ui/src/components/va-date-input/VaDateInput.vue
```diff
@@ -3,7 +3,9 @@
     <va-dropdown
       v-model="isOpenSync"
       :offset="[2, 0]"
-      :close-on-content-click="false"
+      :close-on-content-click=false
```
aluarius, line 6: Why no quotes here and below?

m0ksem: > Because passing it with quotes will pass a `string` instead of expected `boolean`

Isn't `close-on-content-click="false"` will pass string, but `:close-on-content-click="false"` boolean? Looks at syntax highlight:

`:close-on-content-click=false` looks weird for me.

Derranion: > `:` forces Vue not to accept value as string but as js expression.
> 
> :attr="false" - boolean. attr="false" - string.
> 
> https://vuejs.org/guide/components/props.html#boolean

Yes, I thought that this is a more explicit way to assign `boolean`.
Because in pure JS expression like this - "false" will return a `string`.

Derranion: > > Because passing it with quotes will pass a `string` instead of expected `boolean`
> 
> Isn't `close-on-content-click="false"` will pass string, but `:close-on-content-click="false"` boolean? Looks at syntax highlight:
> 
> `:close-on-content-click=false` looks weird for me.

That's right.

By the way quotes here are completely optional - same goes to all `html5` `attributes` as well.

For me it looked more clear to assign `boolean` this way,
Though we could have quotes everywhere just for consistency because they will be removed while minified anyway.

Issue #1838, opened 2022-05-22
Safari desktop va-radio keyboard navigation doesn't work


asvae:
Click on radio, then press Tab,

expecting to select next radio

really selects Search

https://vuestic.dev/en/ui-elements/radio

PR #1843, opened 2022-05-23
[#1838] va-radio tab selection fix, accessibility updates

aluarius:
close: #1838 

## Description
`VaRadio` `Tab` navigation was broken - now it's fixed. Few minor related accessibility updates. 

P.S. Safari tab navigation - `Opt` + `Tab` (https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex). Can think about global service to replace this behaviour (not sure if we need to).

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-radio/VaRadio.vue
+++ b/packages/ui/src/components/va-radio/VaRadio.vue
@@ -8,15 +8,17 @@
       type="radio"
       :checked="isActive"
       :disabled="disabled"
-      :name="name"
+      :name="getName"
+      :value="computedLabel"
       @change="onClick"
       @focus="onFocus"
-      :tabindex="tabindex"
+      :tabindex="getTabIndex"
     >
 
     <span
       class="va-radio__icon"
       :style="iconComputedStyles"
+      aria-hidden="true"
     >
       <span
         class="va-radio__icon__background"
@@ -40,6 +42,7 @@
 import { defineComponent, PropType, computed } from 'vue'
 import { useColors } from '../../composables/useColor'
 import { useFormProps, useForm } from '../../composables/useForm'
+import { generateUniqueId } from '../../services/utils'
 
 export default defineComponent({
   name: 'VaRadio',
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-radio/VaRadio.vue
```diff
@@ -104,6 +107,8 @@ export default defineComponent({
       computedLabel,
       onClick,
       onFocus,
+      getName: computed(() => props.name || generateUniqueId()),
+      getTabIndex: computed(() => props.readonly || props.disabled ? -1 : props.tabindex),
```
RVitaly1978, line 111:
```suggestion
      computedName: computed(() => props.name || generateUniqueId()),
      computedTabIndex: computed(() => props.readonly || props.disabled ? -1 : props.tabindex),
```

RVitaly1978 (approved):

packages/ui/src/components/va-radio/VaRadio.vue
```diff
@@ -104,6 +109,8 @@ export default defineComponent({
       computedLabel,
       onClick,
       onFocus,
+      computedName: computed(() => props.name || generateUniqueId()),
+      computedTabIndex: computed(() => props.readonly || props.disabled ? -1 : props.tabindex),
```
Derranion, line 113: We can move away declaration from `return` to `setup` body.

Derranion (approved):

rustem-nasyrov (approved):

packages/ui/src/components/va-radio/VaRadio.vue
aluarius: Maxim suggested to move not-reusable final declarations to return.

Derranion: And what if it will become reusable in the future?
Not very convenient to move it back and forth.

aluarius: It makes sense. Think should discuss this pattern at next grooming.

Issue #1374, opened 2021-12-29
Accessibility improvements


asvae:
- [x] #1295
- [x] #1891

**Accessibility markup was checked for:**
- [x] VaAccordion
- [x] VaAffix
- [x] VaAlert
- [x] VaAppBar
- [x] VaAvatar
- [x] VaBackTop
- [x] VaBadge
- [x] VaBreadcrumbs
- [x] VaButton
- [x] VaButtonDropdown
- [x] VaButtonGroup
- [x] VaButtonToggle
- [x] VaCard
- [x] VaCarousel
- [x] VaCheckbox
- [x] VaChip
- [x] VaCollapse
- [x] VaColorIndicator
- [x] VaColorInput
- [x] VaColorPalette
- [x] VaContent
- [x] VaCounter
- [x] VaDataTable
- [x] VaDateInput
- [x] VaDatePicker
- [x] VaDivider
- [x] VaDropdown
- [x] VaFileUpload
- [x] VaForm
- [x] VaHover
- [x] VaIcon
- [x] VaImage
- [x] VaInfiniteScroll
- [x] VaInnerLoading
- [x] VaInput
- [x] VaList
- [x] VaModal
- [x] VaNavbar
- [x] VaOptionsList
- [x] VaPagination
- [x] VaParallax
- [x] VaPopover
- [x] VaProgressBar
- [x] VaProgressCircle
- [x] VaRadio
- [x] VaRating
- [x] VaSelect
- [x] VaSeparator
- [x] VaSidebar
- [x] VaSlider
- [x] VaSpacer
- [x] VaSwitch
- [x] VaTabs
- [x] VaTimeInput
- [x] VaTimePicker
- [x] VaToast

asvae:
Moved non finished requirements to separate issues.

PR #1848, opened 2022-05-24
`VaImage` alt prop, docs update

aluarius:
Related to: #1374 

## Description
Alt prop added (important for SEO & very important for accessibility). Docs updated.

## Types of changes


packages/ui/src/components/va-image/VaImage.vue
```diff
@@ -42,6 +43,7 @@ export default defineComponent({
     ratio: { type: Number as PropType<number>, default: 1 },
     contain: { type: Boolean as PropType<boolean>, default: false },
     src: { type: String as PropType<string>, required: true },
+    alt: { type: String, default: '' },
```
RVitaly1978, line 46: Maybe we should make it required?

RVitaly1978 (approved):

packages/ui/src/components/va-image/VaImage.vue
aluarius: `required` is a critically important for component functionality props, but its not that case. If user doesn't want to waste time on this, we shouldn't force him.

rustem-nasyrov (approved):

PR #1851, opened 2022-05-25
[#1850] Update `useForm` hook

### LINKED ISSUE #1850: Update `useForm` hook
Now we have:
```vue
const { createComputedClass } = useForm(props)
const classes = createComputedClass('va-some-component')
```

we want something like this
```vue
const { classes } = useForm(props, 'va-some-component')
```

RVitaly1978:
## Description
close #1850

changes:
- [x] update `useForm`: 
    - add `computedClasses` and `computedClassesArray`
    - remove `createComputedClass` and `createComputedArrayClass`
- [x] update components and tests that use this hook

Now we can do that:
```
const { computedClasses, computedClassesArray } = useForm(props, 'va-some-component')
```

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/radio/examples/CustomLabels.vue
+++ b/packages/docs/src/page-configs/ui-elements/radio/examples/CustomLabels.vue
@@ -20,7 +20,7 @@
   </div>
   <div class="mb-4">
     <va-radio
-      v-for="(option) in options"
+      v-for="(option, index) in options"
       :key="option"
       v-model="selectedOption"
       :option="option"


m0ksem:
Having two ways to make the same thing looks bad. `computedClasses` would be enough.

aluarius (approved):

rustem-nasyrov (approved):

m0ksem (approved):
Looks like `useBem` must be refactored the same way. It also can be used here.

```ts
const useForm = (props, prefix) => {
    return {
      computedClasses: useBem(prefix, () => pick(props, ['disabled', 'readonly']) 
   }
}
```

What do you think, @RVitaly1978?

RVitaly1978:
> What do you think, @RVitaly1978?

Agreed: `useBem` will be more convenient.
I will open a new issue for this

PR #1852, opened 2022-05-25
Fix #1846: useStateful works with modelValue

### LINKED ISSUE #1846: `Stateful` does not update from modelValue
**Vuestic-ui version:** 1.4.0

**Steps to reproduce**

1. Set `stateful` prop to `true`
2. Update `modelValue`
3. Stateful component will not get this update

`modelValue` should update component even if it `stateful`

Derranion:
## Description
Closes #1846

`useStateful` now works correctly with `modelValue` - previously `modelValue` updates were ignored completely.
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useStateful.ts
+++ b/packages/ui/src/composables/useStateful.ts
@@ -1,4 +1,4 @@
-import { ref, computed, PropType, Ref } from 'vue'
+import { ref, computed, PropType, Ref, watch, toRef } from 'vue'
 
 export type StatefulProps<T> = {
   stateful: boolean
@@ -29,6 +29,19 @@ export function useStateful<T, D extends T = T> (
   defaultValue?: D,
 ) {
   const valueState = ref(defaultValue === undefined ? props.modelValue : defaultValue) as Ref<T>
+  const modelValueRef = toRef(props, 'modelValue')
+  const statefulRef = toRef(props, 'stateful')
+  let unwatchModelValue: Function
+
+  const watchModelValue = () => {
+    unwatchModelValue = watch(modelValueRef, (modelValue) => {
+      valueState.value = modelValue
+    })
+  }
+
+  watch(statefulRef, (stateful: boolean) => {
+    stateful ? watchModelValue() : unwatchModelValue?.()
+  }, { immediate: true })
 
   const valueComputed = computed({
     get () {
--- a/packages/ui/src/composables/useStateful.ts
+++ b/packages/ui/src/composables/useStateful.ts
@@ -1,4 +1,4 @@
-import { ref, computed, PropType, Ref, watch, toRef } from 'vue'
+import { ref, computed, PropType, Ref, watch } from 'vue'
 
 export type StatefulProps<T> = {
   stateful: boolean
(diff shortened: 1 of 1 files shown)


aluarius (approved):

rustem-nasyrov (approved):

packages/ui/src/composables/useStateful.ts
```diff
@@ -29,6 +29,19 @@ export function useStateful<T, D extends T = T> (
...
+
+  const watchModelValue = () => {
+    unwatchModelValue = watch(modelValueRef, (modelValue) => {
+      valueState.value = modelValue
+    })
+  }
+
+  watch(statefulRef, (stateful: boolean) => {
+    stateful ? watchModelValue() : unwatchModelValue?.()
+  }, { immediate: true })
```
RVitaly1978, line 44: Maybe it will work the same?
```suggestion
  watch([props.stateful, props.modelValue], ([stateful: boolean, modelValue: T]) => {
    if (stateful && modelValue !== undefined) {
      valueState.value = modelValue
    }
  }, { immediate: true })
```

RVitaly1978 (changes requested):

packages/ui/src/composables/useStateful.ts
Derranion: This will never work because `props.stateful` and `props.modelValue` are not `refs`.

If we will make them `refs` to make it work there will be other issue - we will always watch for `modelValue` even when it is not needed.

Which is not optimal.

Derranion: This will never work because props.stateful and props.modelValue are not refs.

If we will make them refs, to make it work, there will be other issue - we will always watch for modelValue even when it is not needed.

Which is not optimal.

RVitaly1978: I updated my suggestion: replaced props with functions:
`watch([() => props.stateful, () => props.modelValue], ([stateful, modelValue]) => {...`

Derranion: Ok then

"there will be other issue - we will always have additional watcher for modelValue in every component and change valueState twice when it is not needed.

Which is not optimal."

Derranion: Ok then

"there will be other issue - we will always have additional watcher for modelValue in every component when it is not needed.

Which is not optimal."

RVitaly1978: If there are no cases where `stateful` can change dynamically, we can do this:
```
if (props.stateful) {
  watch(() => props.modelValue, (modelValue) => {
    if (modelValue !== undefined) {
      valueState.value = modelValue
    }
  }, { immediate: true })
}
```
otherwise we can update your solution a bit:
```
let unwatchModelValue: Function

const watchModelValue = () => {
  unwatchModelValue = watch(() => props.modelValue, (modelValue) => {
    if (modelValue !== undefined) {
      valueState.value = modelValue
    }
  }, { immediate: true }) // to handle if passed defaultValue
}

watch(() => props.stateful, (stateful) => {
  stateful ? watchModelValue() : unwatchModelValue?.()
}, { immediate: true })
```

Derranion: `Stateful` can be changed by the user any time.

`undefined` can be a valid `modelValue` in some cases.

Passed `defaultValue` is handled immediately here:
`  const valueState = ref(defaultValue === undefined ? props.modelValue : defaultValue) as Ref<T>`

I've moved `refs` to `functions` though, indeed slightly less verbose but less explicit as well.

RVitaly1978 (approved):

Issue #1810, opened 2022-05-19
Optimize DOM


Derranion:
**Vuestic-ui version:** 1.4.0

**Expected behavior:**
We want to have minimal amount of `DOM` elements in the components.
This means that every `DOM` element should be used or otherwise not present in the `DOM`.

**Current behavior:**
Some of the `html` elements in the components are needlessly present in the `DOM`.
Even if they are not usable in the current state of the component.

**How to do this:**
To achieve this we can think about overall structure optimization and add `conditional operators` \ `computed properties` to show only elements that are used.

Feel free to create a separate task i.e. "Optimize ComponentName DOM" and add this component that you have checked \ optimized to the list below, so that we will know what was done already:

**List of already optimized components:**
- [x] VaAccordion
- [x] VaAffix
- [x] VaAlert
- [x] VaAppBar
- [x] VaAvatar
- [x] VaBackTop
- [x] VaBadge
- [x] VaBreadcrumbs
- [x] VaButton
- [x] VaButtonDropdown
- [x] VaButtonGroup
- [x] VaButtonToggle
- [x] VaCard
- [x] VaCarousel
- [x] VaCheckbox
- [x] VaChip
- [x] VaCollapse
- [x] VaColorIndicator
- [x] VaColorInput
- [x] VaColorPalette
- [x] VaContent
- [x] VaCounter
- [x] VaDataTable
- [x] VaDateInput
- [x] VaDatePicker
- [x] VaDivider
- [x] VaDropdown
- [x] VaFileUpload
- [x] VaForm
- [x] VaHover
- [x] VaIcon
- [x] VaImage
- [x] VaInfiniteScroll
- [x] VaInnerLoading
- [x] VaInput
- [x] VaList
- [x] VaModal
- [x] VaNavbar
- [x] VaOptionsList
- [x] VaPagination
- [x] VaParallax
- [x] VaPopover
- [x] VaProgressBar
- [x] VaProgressCircle
- [x] VaRadio
- [x] VaRating
- [x] VaSelect
- [x] VaSeparator
- [x] VaSidebar
- [x] VaSlider
- [x] VaSpacer
- [x] VaSwitch
- [x] VaTabs
- [x] VaTimeInput
- [x] VaTimePicker
- [x] VaToast

PR #1854, opened 2022-05-25
Components accessibility updates & dom optimization checks

aluarius:
Related to:
#1374 
#1810 

## Description
`VaAlert`, `VaAppBar`, `VaAvatar`, `VaBackTop`, `VaBadge`, `VaBreadcrumbs`  accessibility updates & dom optimization checks.

## Types of changes


packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -17,6 +20,7 @@
         :style="contentStyle"
         class="va-alert__icon"
         v-if="hasIcon"
```
rustem-nasyrov, line 22: Better move `v-if` statement to the top of list attributes.

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -31,14 +35,18 @@
           :style="titleStyle"
           class="va-alert__title"
           v-if="hasTitle"
```
rustem-nasyrov, line 37: Same as above.

packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -13,6 +13,7 @@
       <img
         v-else-if="src"
         :src="src"
+        alt=""
```
rustem-nasyrov, line 16: Maybe we should add here some example text, like a `Avatar`?

aluarius: It won't be useful for screenreaders (no information about image - just its purpose). Empty `alt` here is because without it screenreaders will read `src` (just ignoring with empty `alt`).

packages/ui/src/components/va-badge/VaBadge.vue
```diff
@@ -63,7 +67,9 @@ export default defineComponent({
       opacity: props.transparent ? 0.5 : 1,
     }))
 
-    return { badgeClass, badgeStyle }
+    const labelIdComputed = computed(() => `label-${generateUniqueId()}`)
```
RVitaly1978, line 70:
```suggestion
    const labelIdComputed = computed(() => `aria-label-${generateUniqueId()}`)
```

RVitaly1978 (approved):

packages/ui/src/components/va-avatar/VaAvatar.vue
aluarius: Think i'll add `aria-hidden = true` to icon state and optional `alt` prop for image state (some others UI frameworks deal with it in such way).

RVitaly1978 (approved):

packages/ui/src/components/va-alert/VaAlert.vue
```diff
@@ -203,6 +225,10 @@ export default defineComponent({
       display: flex;
       align-items: center;
       cursor: pointer;
+
+      &:focus {
+        outline: 1px solid;
```
rustem-nasyrov, line 230: @xx13 what colors for the outlines we use?

rustem-nasyrov (approved):

Issue #1858, opened 2022-05-26
VaDivider doesn't use va-divider color


m0ksem:
😶

dnischeta:
Hello!
Do you mean, that content should has `va-divider-border-top-color` also? Because, as I understand, divider lines respect css-variable.
<img width="973" alt="image" src="https://user-images.githubusercontent.com/55282960/171444616-b56418e5-5245-46b9-8162-c62b9dba757e.png">

m0ksem:
> Hello!
> Do you mean, that content should has `va-divider-border-top-color` also? Because, as I understand, divider lines respect css-variable.
> <img width="973" alt="image" src="https://user-images.githubusercontent.com/55282960/171444616-b56418e5-5245-46b9-8162-c62b9dba757e.png">
> 

Hey. No it doesn't. It doesn't react on colors config changes. 

Forgot to mention, I'm working on it in scope of dark themes.

m0ksem:
Merged in 1.5.0 with #1865

Issue #1590, opened 2022-03-24
Calculate color for VaSidebarItem.


m0ksem:
We have getTextColor, but it's unused in VaSidebarItem

---

The problem here with alpha. We are not able to calculate text color if we have alpha background...  We want this feature as service later.

aluarius:
Already fixed.

m0ksem:
Fixed here: https://github.com/epicmaxco/vuestic-ui/issues/2028

Issue #1385, opened 2022-01-01
VaModal and VaCard background should depend on theme colors


m0ksem:
**Vuestic-ui version:** 1.3.3

![image](https://user-images.githubusercontent.com/23530004/147850594-06b8f33f-6f7f-47d3-b506-64cde96967e4.png)
![image](https://user-images.githubusercontent.com/23530004/147850625-5c15103b-4198-41da-b396-74feac1a827f.png)

I think there should be more default vuestic colors.

For example: 
background-1 for body
background-2 for cards
background-3 for form elements

m0ksem:
I would like to see a `useTheme` hook inside vuestic that will override colors config and CSS variables.

Can be done along with https://github.com/epicmaxco/vuestic-ui/issues/1357

m0ksem:
If `useTheme` hook will be able to change CSS variables, user can easy control things, like Select dropdown background.

![image](https://user-images.githubusercontent.com/23530004/147852001-89e54e9e-7a57-4fa2-a3bb-204b4d9019e2.png)

m0ksem:
Here is one more issue. There is not option to change va-switch false background:
![image](https://user-images.githubusercontent.com/23530004/147852129-34afdde1-e467-456b-a686-a679382cbfc4.png)
Global theme config could help here.

m0ksem:
Merged in 1.5.0 with #1865

Issue #1255, opened 2021-11-23
Create a dark theme


kushich:
Need to create a dark theme both for the UI framework and the docs page.
It's a pretty good thing that helps read pages in the evenings and at night.

- [ ] investigate how it works
- [ ] implement the dark mode for UI elements
- [ ] create a color palette for the dark mode
- [ ] create a design for docs
- [ ] implement the dark mode on the docs page

Design needed. (with pretty and high accessibility color palette) #1210

PR #1865, opened 2022-05-28
Feat/dark themes

m0ksem:
firefox_a82WxMzvMX
Better themes support and dark themes as well.

# Breaking changes.
- Colors changed according to https://github.com/epicmaxco/vuestic-ui/issues/744#issuecomment-1117201636 (removed white, dark, black and gray colors)
Added new colors to color-config: textLight, textDark, background, backgroundSoft, backgroundMute, shadow, focus.
- Added auto text color for VaTimePicker and VaDatePicker relative to their backgrounds. Here can be performance issues, but it is optimized by maximum.
- Removed useless CSS variables
- Some component colors maybe changed to fit new color pallet.

## TODO
- [ ] Better types for vuestic color config colors (we must require colors such as primary, background etc, but user must be able to add own.)
- [ ] Check if all colors used correctly. 

closes https://github.com/epicmaxco/vuestic-ui/issues/744
closes https://github.com/epicmaxco/vuestic-ui/issues/1590
closes https://github.com/epicmaxco/vuestic-ui/issues/1255
closes https://github.com/epicmaxco/vuestic-ui/issues/1385
closes https://github.com/epicmaxco/vuestic-ui/issues/757
closes https://github.com/epicmaxco/vuestic-ui/issues/1858
closes https://github.com/epicmaxco/vuestic-ui/issues/1867


m0ksem:
Theme demk from docs for now.

image
image

xx13:
@m0ksem thank you, here are my suggestions and questions:

- `primary` -> `accent` as this colour is mostly used for interactive elements that should stand out of plain text
- we can consider adding `accent-2` if the colour theme is more diverse, and an extra colour can be used to add style to static elements, like bullet points, icons, etc); a good example might be the first version of Vuestic where we had blue and green as accent colours
- `textDark` -> `primary` as it is the most commonly met style on the screen, and it can be applied not only to text, but also to other elements, like icons (if there is only dark theme, then textDark most likely will be white)
- `textLight` -> `inverted / tretiary` that is most likely to be #fff if `primary` is set to ~black, which can be used for elements over dark background to have accessible contrast

if we are to introduce a dark theme, it might have sense to introduce something like `primary const` and `inverted / tretiary const` that keeps the colour value even when switching from light to dark themes. a good example is when you have a white playback button over a video, and it should remain white for both light and dark themes.

I'd have background styles as following:
- `background` -> `background-primary` (the most commonly met style for widgets, cards, or the entire screen if there are no cards/widgets)
- `background-soft` -> `background-secondary`(background for inner groups, secondary elements, notes)
- `background-muted`->`background-form-element`(input bg, inline select, etc)
- if we are speaking of admin themes, I think having separate values like `background-sidebar`, `background-topbar`, `background-screen` will make customization more intuitive (these variables may have the same values as other `background-…` vars)

do you think it makes sense to consider adding `extra` colours, that would match the theme, and will be used for tags, graphs, etc? or will those colour be defined in a different way?

xx13:
`misc form` can be renamed to `divider` as it's used for dividers and for some variants of inputs, so `divider` is its primary function

xx13:
I'll post colour hex value updates to this thread

m0ksem:
@xx13, for text, we calculate color relative to background color. So on white background we use darkColor, on dark - whiteColor.
So user don't need to change darkColor to '#fff' in dark theme. The logic here that darkColor stay dark in any theme.

m0ksem:
> @m0ksem thank you, here are my suggestions and questions:
> 
>     * `primary` -> `accent` as this colour is mostly used for interactive elements that should stand out of plain text
> 
>     * we can consider adding `accent-2` if the colour theme is more diverse, and an extra colour can be used to add style to static elements, like bullet points, icons, etc); a good example might be the first version of Vuestic where we had blue and green as accent colours
> 
>     * `textDark` -> `primary` as it is the most commonly met style on the screen, and it can be applied not only to text, but also to other elements, like icons (if there is only dark theme, then textDark most likely will be white)
> 
>     * `textLight` -> `inverted / tretiary` that is most likely to be #fff if `primary` is set to ~black, which can be used for elements over dark background to have accessible contrast
> 
> 
> if we are to introduce a dark theme, it might have sense to introduce something like `primary const` and `inverted / tretiary const` that keeps the colour value even when switching from light to dark themes. a good example is when you have a white playback button over a video, and it should remain white for both light and dark themes.
> 
> I'd have background styles as following:
> 
>     * `background` -> `background-primary` (the most commonly met style for widgets, cards, or the entire screen if there are no cards/widgets)
> 
>     * `background-soft` -> `background-secondary`(background for inner groups, secondary elements, notes)
> 
>     * `background-muted`->`background-form-element`(input bg, inline select, etc)
> 
>     * if we are speaking of admin themes, I think having separate values like `background-sidebar`, `background-topbar`, `background-screen` will make customization more intuitive (these variables may have the same values as other `background-…` vars)
> 
> 
> do you think it makes sense to consider adding `extra` colours, that would match the theme, and will be used for tags, graphs, etc? or will those colour be defined in a different way?

- `accent` is a good replacement for `primary`, but having `accent-2` is kind of strange solution. Probably user will seek for `accent-1` or event `accent-0`. `primary` and `secondary` easy to understand, because it is two opposites.
- since we calculate text color relative to it background `text-dark` and `text-light` must stay constant.
- how about `backround-form`? Just simply to remember across a user's app.
- user can change any component color via props or global config, so we don't need extra color for one type of component. For example, carousel buttons use '#00000aa' color, but it is absent in config, because this color is used only in VaCarousel component.
For admin template, we can define any amount of colors we want. Here we just need a couple of them, so the framework can work.



In Android apps it is common to have the following colors: primary, secondary, triadic, complementary We can actually rename misc-form to triadic color. Same with backgorunds.

packages/ui/src/composables/useElementRef.ts
```diff
@@ -0,0 +1,22 @@
...
+type Maybe<T> = T | undefined
+export type MaybeElement<T extends Maybe<HTMLElement>> = T | ComponentPublicInstance
+
+const unrefElement = <T extends Maybe<HTMLElement>>(el: Ref<MaybeElement<T>>): T => {
+  const e = unref(el)
+  return (e as ComponentPublicInstance)?.$el ?? e
+}
+
+export const useElementRef = <T extends Maybe<HTMLElement>>() => {
+  const el = ref<MaybeElement<T>>()
```
RVitaly1978, line 12:
```suggestion
  const el = shallowRef<MaybeElement<T>>()
```

packages/ui/src/composables/useElementTextColor.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Returns text color based on element background.
+ *
+ * If HTMLElement provided will take element background counting opacity.
+ *
+ * If string (CSS color) provided will use it as background color.
+ *
+ * If element not provided current instance element will be used.
+ */
+export const useElementTextColor = (background?: Ref<HTMLElement | CssColor | undefined> | CssColor) => {
+  if (isRefString(background)) {
```
RVitaly1978, line 18: what if `background` type === `CssColor`?

m0ksem:
image
image

m0ksem:
Merging to prevent conflicts in feture PRs. Stuff to do:

- [ ] Optimize useElementBackgorund (with caching)
- [ ] Update colors demo.
- [ ] Setup theme service config.
- [ ] Maybe rename color names.

Issue #1879, opened 2022-05-30
App is not included in types build


m0ksem:
va-toast/plugin/index.d.ts
![image](https://user-images.githubusercontent.com/23530004/171059077-e00e7ff8-1a54-4a86-908e-99ad0fa8b226.png)

va-toast/plugin/index.ts
![image](https://user-images.githubusercontent.com/23530004/171059120-cbb8e039-c680-46a9-aa13-24f79d83f9d2.png)

PR #1881, opened 2022-05-30
Fix redeclaration of @vue/runtime-core.

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1879

For some reason, we need to declare 'vue' instead of '@vue/runtime-dom' if we want to add global properties.

# Before
image

Result (read before in issue)
# After
image
image

### DIFF
--- a/packages/ui/src/components/va-toast/plugin/index.ts
+++ b/packages/ui/src/components/va-toast/plugin/index.ts
@@ -1,10 +1,11 @@
-import { App } from 'vue'
+import { App, ref } from 'vue'
 import { defineVuesticPlugin, defineGlobalProperty } from '../../../vuestic-plugin/utils'
 import { createToastInstance, closeById, closeAllNotifications, NotificationOptions } from '../toast'
 
 const createVaToastPlugin = (app: App) => ({
   /** Returns toast instance id */
   init (options: string | NotificationOptions) {
+    console.log(ref)
     return createToastInstance(options, app?._context)
   },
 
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/va-toast/plugin/index.ts
```diff
@@ -1,10 +1,11 @@
-import { App } from 'vue'
+import { App, ref } from 'vue'
 import { defineVuesticPlugin, defineGlobalProperty } from '../../../vuestic-plugin/utils'
 import { createToastInstance, closeById, closeAllNotifications, NotificationOptions } from '../toast'
 
 const createVaToastPlugin = (app: App) => ({
   /** Returns toast instance id */
   init (options: string | NotificationOptions) {
+    console.log(ref)
```
rustem-nasyrov, line 8: Remove console.log. :)

packages/ui/src/components/va-toast/plugin/index.ts
```diff
@@ -19,11 +20,11 @@ const createVaToastPlugin = (app: App) => ({
 
 export const VaToastPlugin = defineVuesticPlugin(() => ({
   install (app) {
-    defineGlobalProperty(app, '$vaToast', createVaToastPlugin(app))
+    defineGlobalProperty(app, '$vaToast' as any, createVaToastPlugin(app))
```
rustem-nasyrov, line 23: Is it was on purpose?

rustem-nasyrov (changes requested):

PR #1918, opened 2022-06-07
[#1873] Add `VaDataTable` prop `sticky-footer`

### LINKED ISSUE #1873: va-data-table fixed header/footer
Right now table without pagination is extremely wonky as you can't interact with header once you scrolled down.

image

Fixed header is very important as it allows table to occupy limited space on page, be fully scrollable (pagination is not great on mobile) and filterable/sortable.
Splitting screen vertically with multiple tables also becomes a possibility.

Not sure how hard implementation side would be on that, but we have to take step in that direction. If it's too hard to do with current implementation - please mark that in notion and this issue, so that we're aware of tradeoffs.

- [ ] if header is fixed - headappend slot should also stay fixed
- [ ] if footer is fixed - footappend slot should also stay fixed

RVitaly1978:
## Description
close #1873 

changes:
- [x] add prop `sticky-footer`
- [x] fix tables border style
- [x] update docs, demos, locales

chrome-capture-2022-5-7

## Types of changes


aluarius (approved):
Mb we should delete VaDataTable.demo.vue & rename VaDataTable.new.demo.vue?

rustem-nasyrov (approved):

PR #1921, opened 2022-06-07
Feat #1806: Presets for all components

### LINKED ISSUE #1806: Component config variants
# Useage example
```vue
<va-button variant="buy" />

<va-button variant="cancel" />
```

We want to make user be able to create component from some kind of presets.

User should be able to define varians in global config

```ts
createVuestic({
   config: {
      components: {
         VaButton: {
             variants: {
                buy: {
                   color: '#0fe',
                   icon: 'money',
                   gradient: true,
                   
                   class: 'buy-button'
                },
                cancel: {
                   color: 'secondary',
                   flat: true,
                }
             }
         }
      }
   }
})
```

https://www.notion.so/epicmax/b1b5f6118bde486fb72b88fa932249ea

See also https://github.com/epicmaxco/vuestic-ui/issues/1527
and https://github.com/epicmaxco/vuestic-ui/issues/1856

Derranion:
## Description

- New feature - `Presets` has been added to all usable components.
- Configs demo has been updated as well to make it more clear.

Closes #1806

We can use `preset` `prop` everywhere to use as many `props` as we want at once in a single word.

Also it is possible to set this inside the `Va-config` (`local config`) and other `Global config` (`components` \ `componentsAll`).

Usage example is under the spoiler.
<details>

```js
<va-card preset="highlightTop">

const cardPreset = {
  highlightTop: {
    square: true,
    outlined: true,
    stripe: true,
  },
}

// Any set method
mergeGlobalConfig({ componentsPresets: { VaCard: cardPreset },
```
</details>

Demo:
presetsDemo

Documentation is in the other pull request because of the CI issues.
Just to make sure that this will be deployed.

## Types of changes


packages/ui/src/services/component-config/component-config.ts
```diff
@@ -18,7 +19,12 @@ export const useComponentConfigProps = <T extends DefineComponent>(component: T)
     const localConfigProps = localConfig.value
       .reduce((finalConfig, config) => config[component.name] ? { ...finalConfig, ...config[component.name] } : finalConfig, {})
 
-    const props = { ...globalConfigProps, ...localConfigProps }
+    const presetName = instance.props?.preset || localConfigProps.preset || globalConfigProps.preset
+    const getPresetProps = () => globalConfig.value.componentsPresets?.[component.name]?.[presetName]
```
m0ksem, line 23: Shouldn't we be able to set preset for all components?

For example global preset 
```
'green': { color: 'green', class: 'with-🐸' }
```

And use it both on VaInput and VaButton for example?

m0ksem (approved):

packages/ui/src/services/component-config/component-config.ts
asvae: BTW, as we do renaming and breaking, it might be cool to rename `all` to `global` as the latter sounds better. :)

asvae (approved):
Looks awesome. Thanks!

I believe this would be a huge asset for internal (picker, tables) and external component usage.

aluarius (approved):

packages/ui/src/services/component-config/component-config.ts
Derranion: @m0ksem @asvae 
`componentsAll` currently just applies `props` to all components, 
`preset` is also a `prop` and we can set it there - it will work fine.

Though creating `presets` in `componentsAll` tab will duplicate `componentsPresets` logic and seems a bit weird because its main purpose is to just apply a `prop` `value`, not to create a new entity.

packages/ui/src/services/component-config/component-config.ts
```diff
@@ -1,11 +1,12 @@
 import { useLocalConfig } from '../../components/va-config/VaConfig'
 import { useGlobalConfig } from '../global-config/global-config'
-import { computed, DefineComponent, ref } from 'vue'
+import { ComponentInternalInstance, computed, DefineComponent } from 'vue'
 
 export type Props = { [propName: string]: any }
 export type ComponentConfig = { [componentName: string]: Props }
+export type ComponentPreset = { [componentName: string]: { [presetName: string]: Props } }
```
RVitaly1978, line 7:
```suggestion
export type ComponentPreset = { [presetName: string]: { [componentName: string]: Props | Props } }
```
in this case we can remove `componentsAll` from `GlobalConfig` and add a key `default` (as example) to the `componentsPresets`. And `componentsPresets` could be like this:
```ts
componentsPresets = {
  default: {
    color: 'green' // for all components
    VaButton: {
      color: 'lightgreen',
      class: 'with-🐸',
    },
  },
  customPreset: {
    color: 'orange' // for all components
    VaButton: {
      class: 'with-custom-🐸'
    },
  },
}
```

Derranion: @RVitaly1978
Here is related discussion in Slack:
https://dream-support.slack.com/archives/CG75YRY1Y/p1654616615000269

In two words:
We didn't decide it yet because there are some pros and cons to this.

packages/ui/src/services/component-config/component-config.ts
Derranion: I can extend `Presets` logic to have `all` option next. Probably in a separate task after the core changes.

rustem-nasyrov (approved):

Issue #1923, opened 2022-06-08
New design of VaButton implementation


aluarius:
New buttons ([figma](https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/vuestic-ui?node-id=1538%3A7170)) have some new states, that will affect the list of component's props.

1. Add the `pressed` state. I can't see it anywhere at all. May be we should create new composable `usePressed`.
2. Delete the `outline` prop and state (no design for it, watch next option).
3. Add `borderColor` prop (undefined as default). In purpose to create old-like outline buttons based on `secondary` state of the new design (suggestion by Vasily).
4. Rename the `flat` prop to `plain`.
5. ~~Delete `type` prop. It's seems to be double of the `tag` prop. ~~
5.1. NEW: visual presets (`primary, secondary, plain, plain-opacity`) will be located in `Global Config` and connected with components via https://github.com/epicmaxco/vuestic-ui/pull/1921
5.2. NEW: opacity variables (e.g.: `opacity-hover, hover-opacity` etc.) will be located in `Color Config`.
7. ~~Add the `borderRadius` prop (`undefined` as default). To allow users to change default border radius (suggestion by Vasily).~~
8. CHANGED: `round` will serve as a helper for `border-radius: 50%`.
9. Delete the `rounded` prop (all new variants are square).
10. ~~Change default `color` prop value to `undefined` (by default appearance will depend on new `type`/`style` prop).~~
11. Add `iconColor` prop if we want different colors for text & icon (suggestion by Vasily).

aluarius:
@RVitaly1978 @asvae @Derranion @rustem-nasyrov take a look, please.

asvae:
@xx13 can you explain why we're getting rid of outline buttons?

asvae:
3 - I don't see a point. If outline buttons are solution we don't want our users to employ - why give them tools? I suggest supporting outline preset or removing altogether.
5 - probably we want to use new preset system for that.
6 - that's not great, definitely not prop, css variable - yes.
7-8 - I strongly feel we need to have a way for button to be rounded/not rounded. There are cases like fab button, so even in same design solution - you can have semantically different buttons of varying roundedness.
9. Probably should still have some kind of default on button. You can think of presets as external layer. It adds goodness, but button should be perfectly functional without it.

Other points look reasonable. 👯

asvae:
Let's discuss that in grooming :)

aluarius:
[...]

PR #1945, opened 2022-06-14
Feat/1923 new buttons implementation

aluarius:
Close: #1923 

## Description
Work in progress.

- [x] `VaButton` component was fully rewritten according to the new [design](https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/vuestic-ui?node-id=1538%3A7170).
- [ ] Integrate presets functionality after https://github.com/epicmaxco/vuestic-ui/pull/1921
- [ ] Make changes to other components (and their demos) that are using `VaButton`.
- [ ] Update the documentation.

Please, take a look to the new implementation and, if possible, compare it with figma.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-button/VaButton.demo.vue
+++ b/packages/ui/src/components/va-button/VaButton.demo.vue
@@ -1,247 +1,129 @@
 <template>
   <VbDemo>
-    <VbCard title="empty">
-      <va-button />
-    </VbCard>
-
-    <VbCard>
+    <VbCard class="va-button-demo">
       <table class="table table-bordered">
-        <tr>
-          <th>Description</th>
-          <th>Vuestic Button</th>
-        </tr>
         <tr>
           <td>Types</td>
           <td>
-            <va-button outline>
-              Outline Button
-            </va-button>
-            <va-button flat>
-              Flat Button
-            </va-button>
+            <va-config :components="{ VaButton: {} }">
+              <va-button>Accent (default)</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.primary } }">
+              <va-button>Primary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.secondary } }">
+              <va-button>Secondary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plain } }">
+              <va-button>Plain</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plainOpacity } }">
+              <va-button>Plain with opacity</va-button>
+            </va-config>
+          </td>
+        </tr>
+
+        <tr>
+          <td>Small size</td>
+          <td>
+            <va-config :components="{ VaButton: { size: 'small' } }">
+              <va-button>Accent (default)</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.primary, size: 'small' } }">
+              <va-button>Primary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.secondary, size: 'small' } }">
+              <va-button>Secondary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plain, size: 'small' } }">
+              <va-button>Plain</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plainOpacity, size: 'small' } }">
+              <va-button>Plain with opacity</va-button>
+            </va-config>
           </td>
         </tr>
         <tr>
-          <td>Sizes</td>
+          <td>With icons</td>
           <td>
-            <va-button>
-              Default Button
-            </va-button>
-            <va-button size="large">
-              Large Button
-            </va-button>
-            <va-button size="small">
-              Small Button
-            </va-button>
-            <va-button outline>
-              Default Button
-            </va-button>
-            <va-button
-              size="large"
-              outline
-            >
-              Large Button
-            </va-button>
-            <va-button
-              size="small"
-              outline
-            >
-              Small Button
-            </va-button>
-            <va-button flat>
-              Default Button
-            </va-button>
-            <va-button
-              size="large"
-              flat
-            >
-              Large Button
-            </va-button>
-            <va-button
-              size="small"
-              flat
-            >
-              Small Button
-            </va-button>
+            <va-config :components="{ VaButton: {} }">
+              <va-button icon="create">Edit Button</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.secondary } }">
+              <va-button icon="gavel" iconRight="clear">Mixed Button</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plainOpacity } }">
+              <va-button icon="gavel">Plain with icon</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { size: 'small' } }">
+              <va-button icon="create">Small Edit Button</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: {} }">
+              <va-button icon="gavel" icon-color="warning">Mixed Colors</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.primary } }">
+              <va-button icon-right="clear">Dropdown Button</va-button>
+            </va-config>
           </td>
         </tr>
         <tr>
-          <td>Primary Button Colors</td>
+          <td>Button Colors</td>
           <td>
-            <div>
-              <va-button>
-                Success Button
-              </va-button>
-              <va-button color="info">
-                Info Button
-              </va-button>
-              <va-button color="danger">
-                Danger Button
-              </va-button>
-              <va-button color="warning">
-                Warning Button
-              </va-button>
-              <va-button color="gray">
-                Gray Button
-              </va-button>
-              <va-button color="dark">
-                Dark Button
-              </va-button>
-            </div>
-          </td>
-        </tr>
-        <tr>
-          <td>Outline Button Colors</td>
-          <td>
-            <div>
-              <va-button outline>
-                Success Button
-              </va-button>
-              <va-button
-                outline
-                color="info"
-              >
-                Info Button
-              </va-button>
-              <va-button
-                outline
-                color="danger"
-              >
-                Danger Button
-              </va-button>
-              <va-button
-                outline
-                color="warning"
-              >
-                Warning Button
-              </va-button>
-              <va-button
-                outline
-                color="gray"
-              >
-                Gray Button
-              </va-button>
-              <va-button
-                outline
-                color="dark"
-              >
-                Dark Button
-              </va-button>
-            </div>
-          </td>
-        </tr>
-        <tr>
-          <td>Flat Button Colors</td>
-          <td>
-            <div>
-              <va-button flat>
-                Success Button
-              </va-button>
-              <va-button
-                flat
-                color="info"
-              >
-                Info Button
-              </va-button>
-              <va-button
-                flat
-                color="danger"
-              >
-                Danger Button
-              </va-button>
-              <va-button
-                flat
-                color="warning"
-              >
-                Warning Button
-              </va-button>
-              <va-button
-                flat
-                color="gray"
-              >
-                Gray Button
-              </va-button>
-              <va-button
-                flat
-                color="dark"
-              >
-                Dark Button
-              </va-button>
-            </div>
+            <va-config :components="{ VaButton: {} }">
+              <va-button color="info">Info</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.primary } }">
+              <va-button color="danger">Danger</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.secondary } }">
+              <va-button color="warning">Warning</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plain } }">
+              <va-button color="gray">Gray</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plainOpacity } }">
+              <va-button color="dark">Dark</va-button>
+            </va-config>
           </td>
         </tr>
         <tr>
           <td>Disabled state</td>
           <td>
-            <va-button disabled>
-              Default Button
-            </va-button>
-            <va-button
-              disabled
-              outline
-            >
-              Outline Button
-            </va-button>
-            <va-button
-              disabled
-              flat
-            >
-              Flat Button
-            </va-button>
+            <va-config :components="{ VaButton: {} }">
+              <va-button disabled>Accent (default)</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.primary } }">
+              <va-button disabled>Primary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.secondary } }">
+              <va-button disabled>Secondary</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plain } }">
+              <va-button disabled>Plain</va-button>
+            </va-config>
+            <va-config :components="{ VaButton: { ...vaButtonPresets.plainOpacity } }">
+              <va-button disabled>Plain with opacity</va-button>
+            </va-config>
           </td>
         </tr>
         <tr>
           <td>Button tags</td>
           <td>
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -68,173 +59,258 @@ export default defineComponent({
...
-    const hoverState = ref(false)
-    const focusState = ref(false)
+    // attributes
+    const { tagComputed, hrefComputed, isLinkTag } = useRouterLink(props)
+    const linkAttributesComputed = computed(() => {
+      if (!isLinkTag.value) { return {} }
 
-    const colorComputed = computed(() => computeColor(props.color, 'primary'))
-    const isTransparentBackground = computed(() => Boolean(props.outline || props.flat))
+      return tagComputed.value === 'a'
```
m0ksem, line 111: Isn't it must be handled in useRouterLink?

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -68,173 +59,258 @@ export default defineComponent({
...
-      'va-button--hover': hoverState.value,
-      'va-button--focus': focusState.value,
-      'va-button--large': props.size === 'large',
       'va-button--small': props.size === 'small',
       'va-button--normal': !props.size || props.size === 'medium',
-      'va-button--loading': props.loading,
-      'va-button--block': props.block,
-      'va-button--square': !props.rounded,
-      'va-button--round': props.round || (!slots.default && hasOneIcon.value),
+      'va-button--plain': props.plain,
```
m0ksem, line 166: May be refactored to useBem leter.

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -1,65 +1,56 @@
...
-    @mouseenter="hoverState = true"
-    @mouseleave="hoverState = false"
-    v-on="$attrs"
-    ref="button"
-    :aria-disabled="$props.disabled"
-    aria-live="polite"
+    v-bind="attributesComputed"
   >
-    <div class="va-button__content" :class="{ 'va-button__content--loading': loading }">
+    <div class="va-button__content" :class="wrapperClassComputed">
```
m0ksem, line 10: Would be great to remove this div!

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -1,65 +1,56 @@
...
         v-if="icon"
-        :name="icon"
-        :size="size"
-        :color="textColorComputed"
         class="va-button__left-icon"
+        :name="icon"
+        v-bind="iconAttributesComputed"
       />
       <slot />
       <va-icon
```
m0ksem, line 18: Icons can be wrapper with prepend and append slots.

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -250,82 +326,6 @@ export default defineComponent({
...
-      margin-right: calc(var(--va-button-lg-space-between-items) / 2);
-    }
-
-    .va-button__right-icon {
-      margin-left: calc(var(--va-button-lg-space-between-items) / 2);
-      margin-right: calc(var(--va-button-lg-content-px) / -2);
-    }
-  }
-
   &--small {
```
m0ksem, line 329: How about big one? :P 

I think having solution based on sizes-config is more appropriate here if it is not actually a "dense" modifier.

m0ksem (changes requested):
I dislike new disabled states - the're completely invisible. (Especially last one)
image

packages/ui/src/components/va-button/_variables.scss
```diff
@@ -1,54 +1,40 @@
...
-  --va-button-sm-border-radius: 1rem;
   --va-button-sm-letter-spacing: 0;
-  --va-button-sm-square-border-radius: 0.2rem;
-  --va-button-sm-space-between-items: 0.25rem;
-
-  /* Outlined */
-  --va-button-outline-border: 0.125rem;
-  --va-button-outline-icon-color: #ffffff;
+  --va-button-sm-line-height: 1rem;
+  --va-button-sm-border-radius: 0.125rem;
```
RVitaly1978, line 34:
```suggestion
  --va-button-sm-border-radius: calc(0.5 * var(--va-button-border-radius));
```
We could use the same approach in similar cases

packages/ui/src/composables/useHover.ts
```diff
@@ -1,4 +1,10 @@
-import { ref, watch, onMounted, onBeforeUnmount, Ref } from 'vue'
+import { PropType, ref, Ref, watch } from 'vue'
+
+export const useHoverProps = {
+  hoverBehaviour: { type: String as PropType<'opacity' | 'gradient'>, default: 'gradient' },
```
RVitaly1978, line 4: Should we add validator here? (and the same in `usePressed`)

packages/ui/src/composables/usePressed.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function usePressed (el?: Ref<HTMLElement | undefined>) {
+  const isPressed = ref(false)
+
+  const onMouseDown = () => { isPressed.value = true }
+
+  const onMouseUp = () => { isPressed.value = false }
+
+  onMounted(() => {
+    if (el?.value instanceof HTMLElement) {
```
RVitaly1978, line 17: This will not work for `nuxt-link` and `router-link`. Do we want it?

RVitaly1978 (changes requested):

packages/ui/src/components/va-button/_variables.scss
aluarius: Think this is redudant, because there is no rule like 'sm value is half normal value'. In the example above just a coincidence.

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -68,173 +59,258 @@ export default defineComponent({
...
+    })
 
-    const { textColorComputed } = useTextColor(colorComputed, isTransparentBackground)
-    const isSlotContentPassed = computed(() => !!slots.default?.()?.[0]?.children)
+    const typeComputed = computed(() => isLinkTag.value ? undefined : props.type)
+    const buttonAttributesComputed = computed(() => {
+      const disabledAttributes = {
+        ariaDisabled: props.disabled,
+        disabled: props.disabled,
+      }
```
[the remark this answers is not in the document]
aluarius: Non-link button can also be `disabled`. So, i think that moving it to `useRouterLink` is mistakenly.

Derranion (commented):
If I get this point correctly:
- [ ] Integrate presets functionality after [Feat #1806: Presets for all components](https://github.com/epicmaxco/vuestic-ui/pull/1921)

We will just have predefined list of `Presets` for the `VaButton`.

If you want you could just send me these presets object I'll add it in my current PR.

packages/ui/src/composables/usePressed.ts
aluarius: It prevents possible errors. I'll think, how to handle nuxt/router links here.

packages/ui/src/components/va-button-toggle/VaButtonToggle.vue
```diff
@@ -44,80 +35,68 @@ export default defineComponent({
   },
   emits: ['update:modelValue'],
   props: {
+    modelValue: { type: [String, Number] as PropType<string | number>, default: '' },
```
rustem-nasyrov, line 38: Isn't here enough to use just a type without a PropType?

packages/ui/src/components/va-toast/VaToast.vue
```diff
@@ -243,6 +243,8 @@ export default defineComponent({
 
     &:focus {
       @include focus-outline;
+
+      border-radius: 2px;
```
rustem-nasyrov, line 247: Let's replace it with a mixin (here and above).

packages/ui/src/composables/useHover.ts
```diff
@@ -1,6 +1,16 @@
...
+  hoverBehaviour: {
+    type: String as PropType<'opacity' | 'mask'>,
+    default: 'mask',
+    validator: (value: string) => ['opacity', 'mask'].includes(value),
+  },
+  hoverOpacity: { type: Number, default: 0.15 },
+  hoverMaskColor: { type: String, default: '#ffffff' },
+}
+
+const getEl = (el: any) => el.$el ?? el
```
rustem-nasyrov, line 13: Can be here `undefined` or `null`?

packages/ui/src/composables/useHover.ts
```diff
@@ -1,6 +1,16 @@
...
 
-const getEl = (el: any) => el.$el !== undefined ? el.$el : el
+export const useHoverProps = {
+  hoverBehaviour: {
+    type: String as PropType<'opacity' | 'mask'>,
+    default: 'mask',
+    validator: (value: string) => ['opacity', 'mask'].includes(value),
+  },
+  hoverOpacity: { type: Number, default: 0.15 },
+  hoverMaskColor: { type: String, default: '#ffffff' },
```
rustem-nasyrov, line 10: We need to use here color from color config (presets) against hard-coded variant.

packages/ui/src/composables/usePressed.ts
```diff
@@ -0,0 +1,37 @@
+import { PropType, ref, Ref, onMounted, onBeforeUnmount } from 'vue'
+
+export const usePressedProps = {
+  pressedBehaviour: {
+    type: String as PropType<'opacity' | 'mask'>,
+    default: 'mask',
+    validator: (value: string) => ['opacity', 'mask'].includes(value),
+  },
+  pressedOpacity: { type: Number, default: 0.13 },
+  pressedMaskColor: { type: String, default: '#000000' },
```
rustem-nasyrov, line 10: Same as above.

packages/ui/src/composables/useHover.ts
aluarius: Are you about 'el'? No, it isn't. But it can be proxy (thats why here is `any`).

packages/ui/src/components/va-button-group/VaButtonGroup.vue
```diff
@@ -1,74 +1,43 @@
...
     return {
-      buttonConfig,
-      computedStyle,
-      computedClass,
+      buttonConfig: computed(() => ({ VaButton: { ...props } })),
+      computedClass: useBem('va-button-group', () => ({
+        square: !props.round,
+        small: props.size === 'small',
+        large: props.size === 'large',
+      })),
```
Derranion, line 40: Don't see why we need to pollute `return` with logic because it makes code harder to reuse.

Also we will have to move this away in case of future `<script setup>` usage.

Derranion (approved):

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -1,241 +1,211 @@
...
-          ? shiftHSLAColor(colorComputed.value, { a: alpha })
-          : shiftHSLAColor(colorComputed.value, { l: lightness })
+    // classes
+    const wrapperClassComputed = computed(() => ({ 'va-button__content--loading': props.loading }))
 
-        background = props.gradient ? getGradientBackground(color) : color
-      }
+    const isSlotContentPassed = computed(() => !!slots.default?.()?.[0]?.children)
+    const isOneIcon = computed(() => !!((props.iconRight && !props.icon) || (!props.iconRight && props.icon)))
+    const computedClass = useBem('va-button', () => ({
```
asvae, line 139: I would suggest  keeping full bem classes if it's not too much pain.
The upside is that keeps them easily searchable. And from what I perceive it doesn't do that much otimization-wise.

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -1,241 +1,211 @@
...
 
 .va-button {
+  position: relative;
+  margin: var(--va-button-margin);
+  padding: var(--va-button-padding);
   display: var(--va-button-display);
-  align-items: var(--va-button-align-items);
   justify-content: var(--va-button-justify-content);
+  align-items: var(--va-button-align-items);
+  border: var(--va-button-border);
```
asvae, line 194: maybe split border into multiple style params? As you have border color separately and it kinda conflicts with 'border'

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -1,241 +1,211 @@
...
-  transition: var(--va-button-transition, var(--va-swing-transition));
-  background-color: var(--va-button-background-color, var(--va-white));
-  vertical-align: middle;
+  transition: var(--va-button-transition);
   box-sizing: border-box;
-  font-weight: var(--va-button-font-weight);
-  margin: var(--va-button-margin);
-  padding: var(--va-button-padding);
-  position: relative;
+  cursor: pointer;
```
asvae, line 203: We probably can move cursor to variable as well. This is a valid point for customization.

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -251,185 +221,161 @@ export default defineComponent({
...
 
-  &--default {
-    color: var(--va-button-background-color, var(--va-white));
-
-    i {
-      color: var(--va-button-icon-color, var(--va-white));
-    }
-  }
+  &--small {
+    @include va-button(var(--va-button-sm-content-py), var(--va-button-sm-content-px), var(--va-button-sm-font-size), var(--va-button-sm-line-height), var(--va-button-sm-border-radius));
```
asvae, line 225: Can you follow this up by issue (or point me to one if it's created already)? I think we reached an agreement that's not a satisfiable solution long term. Works for this pr though!

packages/ui/src/composables/useDeprecatedProps.ts
```diff
@@ -0,0 +1,16 @@
+import { __DEV__ } from '../utils/global-utils'
```
asvae, line 1: Can we add this strategy to notion? "Deprecation guide" of sorts would be nice to have.

packages/ui/src/composables/useHover.ts
```diff
@@ -1,6 +1,16 @@
...
 
-const getEl = (el: any) => el.$el !== undefined ? el.$el : el
+export const useHoverProps = {
+  hoverBehaviour: {
+    type: String as PropType<'opacity' | 'mask'>,
+    default: 'mask',
+    validator: (value: string) => ['opacity', 'mask'].includes(value),
+  },
+  hoverOpacity: { type: Number, default: 0.15 },
+  hoverMaskColor: { type: String, default: 'white' },
```
asvae, line 10: Can we do 

```
type ColorString = string
```

And then use ColorString here?

But I can create an issue to discuss if you don't find this exciting :D.

asvae (changes requested):
Thanks for excellent work and buttons look crazy good!

packages/ui/src/components/va-button-dropdown/VaButtonDropdown.vue
```diff
@@ -161,27 +147,35 @@ export default defineComponent({
...
+      return { [propName]: computedIcon.value }
     })
 
-    const computedViewStyles = computed(
-      () => pick(props, ['outline', 'gradient', 'rounded', 'flat', 'size', 'color']),
+    const buttonPropsFiltered = computed(() => {
+      const filteredProps = omit(VaButtonProps, ['to', 'href', 'loading', 'icon', 'iconRight', 'disabled'])
+      return Object.keys(filteredProps)
+    })
+    const buttonPropsComputed = computed(
```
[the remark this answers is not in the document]
aluarius: No, this is exactly props for `VaButton`.

packages/ui/src/components/va-button-group/VaButtonGroup.vue
```diff
@@ -1,74 +1,43 @@
...
+  setup: (props) => {
+    // temp
+    useDeprecatedProps(['flat', 'outline'])
 
     return {
-      buttonConfig,
-      computedStyle,
-      computedClass,
+      buttonConfig: computed(() => ({ VaButton: { ...props } })),
+      computedClass: useBem('va-button-group', () => ({
```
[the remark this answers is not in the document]
aluarius: Yeah, `script setup` is an argument. Will fix.

packages/ui/src/components/va-button/VaButton.vue
aluarius: Will fix it here.

packages/ui/src/components/va-button/VaButton.demo.vue
```diff
@@ -1,247 +1,89 @@
 <template>
```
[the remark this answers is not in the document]
aluarius: Current examples cover almost all possible combinations. I'll try to come up with a new examples, but they will be just a variations of the existing ones (via `presets`).

rustem-nasyrov (approved):

m0ksem (approved):

m0ksem:
We need to split composables into utils like useHover, usePressed, useHTMLElement, useClientOnly etc. and mixins like useHoverable, useFormElement, usePressable, useSelectable, useParsable.

Maaaaaaah, I think this must be written somewhere before implementation... Naming is quite important here..

Issue #1859, opened 2022-05-26
VaTimeInput doesn't show picker if manual-input prop is true.


m0ksem:
![image](https://user-images.githubusercontent.com/23530004/170503308-25f485d4-2664-45b3-ad4a-b135b09438fa.png)

PR #1968, opened 2022-06-17
fix#1859 Show dropdown on manul input prop.

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1859
closes https://github.com/epicmaxco/vuestic-ui/issues/1972

firefox_aOmO87H3wM

### DIFF
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -90,6 +90,7 @@
     <va-dropdown-content
       no-padding
       @keydown.esc.prevent="hideDropdown"
+      @keypress.enter.prevent="hideDropdown"
     >
       <va-time-picker
         ref="timePicker"
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -165,8 +165,8 @@ export default defineComponent({
     const { format } = useTimeFormatter(props)
 
     const valueText = computed<string | undefined>(() => {
-      if (!isValid.value) { return format(props.clearValue || new Date()) }
-      if (!modelValueSync.value) { return format(props.clearValue || new Date()) }
+      if (!isValid.value) { return props.clearValue ? format(props.clearValue) : '' }
+      if (!modelValueSync.value) { return props.clearValue ? format(props.clearValue) : '' }
 
       if (props.format) { return format(modelValueSync.value) }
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -77,8 +77,8 @@
             :tabindex="iconsTabIndexComputed"
             :id="componentIconId"
             v-bind="iconProps"
-            @click="dropdownToggle"
-            @keydown.enter.stop="dropdownToggle"
+            @click="dropdownToggle(true)"
+            @keydown.enter.stop="dropdownToggle(true)"
```
RVitaly1978, line 81: we have to do the same in lines 53 and 54

RVitaly1978 (changes requested):

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -77,8 +77,8 @@
             :tabindex="iconsTabIndexComputed"
```
Derranion, line 77: Pressing `Escape` should close the dropdown - identical to default `VaTimeInput` behavior.

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -275,16 +275,18 @@ export default defineComponent({
...
+        if (!props.manualInput || forceFocus) {
+          timePicker.value?.focus()
+        }
       })
     }
 
-    const dropdownToggle = () => {
-      isOpenSync.value ? hideDropdown() : showDropdown()
+    const dropdownToggle = (forceFocus = false) => {
+      isOpenSync.value ? hideDropdown() : showDropdown(forceFocus)
```
Derranion, line 289: Just a thoughts - maybe in such cases it is better to just separate already existing logic instead of adding new one.

For example:
`showDropdown` currently focuses `dropdown`.

We can separate this to three methods.
`showDropdown`
`focusDropdown`
`showAndFocusDropdown`

And use this where it is needed.

Derranion (approved):

packages/ui/src/components/va-time-input/VaTimeInput.vue
m0ksem: Fixed.

m0ksem:
Why don't we close dropdown on enter press? (like, I selected a time). Very confusing for me, personally.

RVitaly1978 (approved):

rustem-nasyrov (approved):

Derranion:
> Why don't we close dropdown on enter press? (like, I selected a time). Very confusing for me, personally.
> 
> How about to close dropdown on enter press?

Yeah I thought that we can do this as a part of this task:
https://github.com/epicmaxco/vuestic-ui/issues/1972

To `close` \ `open` on `enter` key press.

PR #1981, opened 2022-06-22
Fix incorrectly applied progress bar height (#1950)

### LINKED ISSUE #1950: `VaProgressBar` with not large size and slot
**Vuestic-ui version:** 1.4.1

**Steps to reproduce**
  - https://vuestic.dev/en/ui-elements/progress-bar#slots

  - the slot should be visible when using the default size
  - the progress bar should have the correct height

https://user-images.githubusercontent.com/55198465/174010950-91b2b97c-4dba-4b81-81ba-71f7228f5af2.mov

smellyshovel:
Resolves #1950 

What it used to look like:
image

What it looks like with the fix applied:
image

The issue was in that the `height` was applied to the whole block (including the slotted content) whereas it should be applied only to the progress bar's wrapper.


packages/ui/src/components/va-progress-bar/VaProgressBar.vue
```diff
@@ -68,10 +67,13 @@ export default defineComponent({
       rootClass: computed(() => ({
         'va-progress-bar': true,
         'va-progress-bar--square': !props.rounded,
+      })),
+
+      wrapperClass: computed(() => ({
         [`va-progress-bar--${props.size}`]: isTextSize.value,
       })),
```
RVitaly1978, line 74: this class must be in the root wrapper

RVitaly1978 (changes requested):

packages/ui/src/components/va-progress-bar/VaProgressBar.vue
smellyshovel: Fair point, will be addressed.

smellyshovel: Resolved via 96f8f5e

RVitaly1978 (approved):

m0ksem (approved):

PR #1990, opened 2022-06-23
[#1940] Fixed filtering and sorting of `VaDataTable` when using `items` with nested objects

### LINKED ISSUE #1940: `VaDataTable`: sorting and filtering don't work with custom slots
**Vuestic-ui version:** 1.4.1

Sorting and filtering don't work if we have such a custom slot
```vue
<va-data-table :items="items" :columns="columns">
  <template #cell(address)="{ rowData }">{{ rowData.address.street }}</template>
</va-data-table>
```

RVitaly1978:
## Description
close #1940

changes:
- [x] add `name` key to `columns`
- [x] update `key` key in `columns`
- [x] update examples

chrome-capture-2022-5-23

chrome-capture-2022-5-23 (1)

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/data-table/examples/Filtering.vue
+++ b/packages/docs/src/page-configs/ui-elements/data-table/examples/Filtering.vue
@@ -21,7 +21,7 @@
     @filtered="filteredCount = $event.items.length"
   />
 
-  <va-alert class="mt-3" border="left">
+  <va-alert class="mt-3" color="info" outline>
     <span>
       Number of filtered items:
       <va-chip>{{ filteredCount }}</va-chip>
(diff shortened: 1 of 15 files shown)


packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -107,7 +107,7 @@ export default function useSortable (
   // Sets the clicked heading's column as a one to sort by and toggles the sorting order from "asc" to "desc" to `null`
   // (un-sorted) if the same column is clicked again or sets sorting order to "asc" if some other column is chosen.
   function toggleSorting (column: DataTableColumnInternal) {
-    if (column.key === sortBySync.value) {
+    if (column.name === sortBySync.value) {
```
smellyshovel, line 110: name is marked as optional. What if it's not provided? Why is `key` subtituted with `name` everywhere?

aluarius (approved):

packages/ui/src/components/va-data-table/hooks/useSortable.ts
m0ksem: Here we have two types: column and internal column. Internal column means that it is a built column. If name is not provided, then key will be used.

packages/docs/src/page-configs/ui-elements/data-table/examples/Filtering.vue
```diff
@@ -40,40 +40,115 @@ export default defineComponent({
         name: 'Leanne Graham',
         username: 'Bret',
         email: 'Sincere@april.biz',
+        address: {
```
m0ksem, line 43: Do we need this large amount of data? It is hard for user to read demo.

m0ksem (approved):

Issue #1975, opened 2022-06-21
Remove VaPopup


RVitaly1978:

m0ksem:
VaScrollbar and VaTimeLine will be updated in the future. VaPopup can be safety removed.

Issue #1829, opened 2022-05-20
Make composables/index


m0ksem:
![image](https://user-images.githubusercontent.com/23530004/169597362-ff74714b-bd37-4cd4-9cb0-a9adce7bdb95.png)

Can we specify one time path of composable?...

PR #1997, opened 2022-06-23
[#1829] [#1975] `VaPopup` deleted, composables index, `useColor`/`useColors` refactoring

aluarius:
Close: #1829 
Close: #1975 

## Description
`VaPopup` component was removed, created composables index (for more clear composables mass import), `useColors` hook moved to composables, `useColor` hook deleted, changes to associated components and hooks, minor refactoring.

## Types of changes


packages/ui/src/composables/index.ts
```diff
@@ -0,0 +1,43 @@
+export { useAccessTrap } from './useAccessTrap'
```
RVitaly1978, line 1:
```suggestion
export * from './useAccessTrap'
```
and the same below.
And we can avoid explicit type exports in this case.

RVitaly1978 (approved):

m0ksem (approved):
My dark theme pr will require fixing a lot of conflicts...

PR #2011, opened 2022-06-27
Feat/input counter and max length

### LINKED ISSUE #1512: Add example to the docs with textarea size limit
Should be something similar to twitter post. I.e. 
* When length is less than 50 - we display message - `30/50`
* When length is >= 50 - we display same message as error.

rustem-nasyrov:
Close #1512

counter-maxlen

## Description
Added counter and max length features support.

## Types of changes


packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -74,6 +74,15 @@
...
     </div>
 
+    <div v-if="hasCounterOrMaxLengthComputed || $slots.hint" class="va-input-wrapper-bottom">
+      <div v-if="$slots.hint" class="va-input-wrapper-bottom__hint">
+        <slot name="hint" />
+      </div>
+      <div v-if="hasCounterOrMaxLengthComputed" class="va-input-wrapper-bottom__counter">
+        {{ $props.counterOrMaxLength }}
+      </div>
+    </div>
```
RVitaly1978, line 84: I think it would be better to give a slot to the user here
```suggestion
    <div v-if="$props.counter" class="va-input-wrapper__counter">
      <slot name="counter" v-bind="{ counter, maxLength }">
        <span class="...">{{ `${counter}/${maxLength}` }}</span>
      </slot>
    </div>
```

RVitaly1978 (changes requested):

packages/docs/src/page-configs/ui-elements/input/examples/Default.vue
```diff
@@ -43,6 +43,24 @@
...
+      counter
+      class="mb-4"
+    >
+      <template #hint>
+        <i>Counter</i>
+      </template>
+    </va-input>
+    <va-input
+      v-model="value"
+      :max-length="30"
```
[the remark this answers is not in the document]
rustem-nasyrov: In the references they are used with a `counter / max-length` behavior.

packages/ui/src/components/va-input/components/VaInputWrapper.vue
rustem-nasyrov: If I do this, here will empty space.

rustem-nasyrov: I think slot here is will be over-engineering thing, and if we remove the hint slot, here will be an empty space that doesn't look good. :)

aluarius (approved):

RVitaly1978 (changes requested):

packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -74,13 +74,15 @@
...
-      </div>
+    <div v-if="isCounterVisible" class="va-input-wrapper-bottom">
+      <slot name="counter" v-bind="{ counter: counterValueComputed, maxLength: $props.maxLength }">
+        <div v-if="isCounterVisible && !isMaxLengthVisible" class="va-input-wrapper-bottom__counter">
+          {{ counter }}
+        </div>
+        <div v-else class="va-input-wrapper-bottom__counter">
+          {{ `${counter} / ${maxLength}` }}
+        </div>
+      </slot>
```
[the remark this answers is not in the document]
rustem-nasyrov: From `VaInput` to `VaInputWrapper` counter prop is already calculated and using `Boolean($props.counter)` will hide the counter on the zero-like values.

RVitaly1978 (approved):

packages/docs/src/locales/ru/ru.json
```diff
@@ -677,7 +677,9 @@
         "focused": "Применяет `focus` стиль",
         "canBeFocused": "Если `false`, тогда компонент не может получить фокус, используя мышь или клавиатуру",
         "immediateValidation": "Устанавливает выполнение валидации при монтировании компонента",
-        "inputClass": "Передает строку с классами непосредственно для самого `input` (а не его обертки)"
+        "inputClass": "Передает строку с классами непосредственно для самого `input` (а не его обертки)",
+        "counter": "Отображает счетчик введенных символов",
+        "maxLength": "Максимальное количество символов"
```
m0ksem, line 682: How this prop works without counter on?

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -123,6 +125,8 @@ export default defineComponent({
     pattern: { type: String },
     inputmode: { type: String, default: 'text' },
     ariaLabel: { type: String, default: undefined },
+    counter: { type: Boolean, default: false },
```
m0ksem, line 128: How about moving it all in one prop?

```suggestion
    counter: { type: [Boolean, Number], default: false },
```

m0ksem (approved):

packages/docs/src/locales/ru/ru.json
rustem-nasyrov: Well, it's needs to show the max-length number.

packages/docs/src/locales/en/en.json
```diff
@@ -685,7 +685,9 @@
         "immediateValidation": "Sets the validation to be performed when the component is mounted",
```
asvae, line 685: Note should stay on the same line as count.

image

packages/docs/src/locales/en/en.json
```diff
@@ -685,7 +685,9 @@
...
         "inputClass": "Classes for `input` (not its wrapper)",
         "pattern": "The pattern prop specifies a regular expression the input value should match",
-        "inputmode": "The inputmode prop is an enumerated prop that hints at the type of data that might be entered by the user while editing the element or its contents. This alows a browser to display an apropriate virtual keyboard"
+        "inputmode": "The inputmode prop is an enumerated prop that hints at the type of data that might be entered by the user while editing the element or its contents. This alows a browser to display an apropriate virtual keyboard",
+        "counter": "Show a counter of value length",
+        "maxLength": "Specify a max length of value"
```
asvae, line 690: I suggest we use `maxlength` as that name is part of the spec.

packages/docs/src/page-configs/ui-elements/input/examples/Default.vue
```diff
@@ -43,6 +43,36 @@
...
+      class="mb-4"
+    />
+    <va-input
+      v-model="value"
+      :max-length="30"
+      counter
+      class="mb-4"
+    >
+      <template #counter="{ counter, maxLength }">
+        <span>Remaining symbols</span>
```
asvae, line 58: I'm not sure you'll have a case with "Remaining symbols" text. Counter is a wrong slot for this amount of text.

Maybe you can do something like "30 characters left".

packages/docs/src/page-configs/ui-elements/input/examples/Default.vue
```diff
@@ -43,6 +43,36 @@
...
+        <b style="margin-left: auto;">{{ `${counter} / ${maxLength}` }}</b>
+      </template>
+    </va-input>
+    <va-input
+      v-model="value"
+      :max-length="30"
+      counter
+      class="mb-4"
+    >
+      <template #counter="{ counter, maxLength }">
```
asvae, line 68: `{ counter, ` here "counter" name is very confusing. Should be boolean, but it's seemingly not. Maybe use `valueLength` or just pass down `value`?

packages/ui/src/components/va-input/_variables.scss
```diff
@@ -43,4 +43,16 @@
   --va-input-container-label-font-weight: 700;
   --va-input-container-label-text-transform: uppercase;
   --va-input-container-label-transform-origin: top left;
+
+  /* Bottom */
+  --va-input-wrapper-bottom-display: flex;
```
asvae, line 48: Does it make sense to have display types in variables?

Looks like something very fragile: user might shoot their foot or we might screw this over on vuestic update.
Do we use variables like this elsewhere?

packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -106,6 +114,8 @@ export default defineComponent({
   props: {
     ...useFormProps,
     ...useValidationProps,
+    counter: { type: Number, default: undefined },
```
asvae, line 117: So, we discussed this should be Boolean, no?

packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -146,11 +156,18 @@ export default defineComponent({
       typeof messagesComputed.value === 'string' ? messagesComputed.value : messagesComputed.value?.length,
     ))
 
+    const isCounterVisible = computed(() => typeof props.counter !== 'undefined')
```
asvae, line 159: And this one would go away if `counter` is boolean.

asvae (changes requested):
Some small changes, but overall looks close to final line 👍.

packages/ui/src/components/va-input/_variables.scss
rustem-nasyrov: > Do we use variables like this elsewhere?
Yes, it's pretty common.

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <VaInputWrapper
     v-bind="fieldListeners"
+    :model-value="computedValue"
```
RVitaly1978, line 4:
```suggestion
```

RVitaly1978 (changes requested):

packages/ui/src/components/va-input/VaInput.vue
rustem-nasyrov: If I won't pass value to the VaInputWrapper I can't get the value `length`, because is value is `undefined` or something.

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -16,6 +17,8 @@
     :outline="$props.outline"
     :requiredMark="$props.requiredMark"
     :focused="isFocused"
+    :counter="$props.counter"
```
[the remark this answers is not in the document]
rustem-nasyrov: I got your point.

packages/docs/src/page-configs/ui-elements/input/examples/Default.vue
```diff
@@ -43,6 +43,37 @@
...
+    <va-input
+      v-model="value"
+      :max-length="30"
+      counter
+      class="mb-4"
+    >
+      <template #counter="{ valueLength, maxLength }">
+        <b
+          style="margin-left: auto;"
+          :style="{color: valueLength > maxLength ? 'var(--va-danger)' : 'var(--va-success)'}">
```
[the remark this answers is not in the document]
rustem-nasyrov: We need to setup linter (prettier) for suggestions like this. :)

RVitaly1978 (changes requested):

packages/ui/src/components/va-input/components/VaInputWrapper.vue
```diff
@@ -74,10 +74,10 @@
       </div>
     </div>
 
-    <div v-if="$props.counter" class="va-input-wrapper__bottom">
-      <slot name="counter" v-bind="{ valueLength: $props.valueLength, maxLength: $props.maxLength }">
+    <div v-if="isCounterVisible" class="va-input-wrapper__bottom">
+      <slot name="counter" v-bind="{ valueLength: $props.counterValue, maxLength: $props.maxLength }">
         <div class="va-input-wrapper__bottom-counter">
-          {{ $props.maxLength ? `${valueLength}/${$props.maxLength}` : valueLength }}
+          {{ $props.maxLength ? `${$props.counterValue}/${$props.maxLength}` : $props.counterValue }}
```
RVitaly1978, line 80: maybe variable?
```ts
counterString = computed(() => props.maxLength !== undefined
  ? `${props.counterValue}/${props.maxLength}`
  : String(props.counterValue)
)
```

RVitaly1978 (changes requested):

packages/ui/src/components/va-input/components/VaInputWrapper.vue
rustem-nasyrov: I think it's not that necessary, because the counter is `slot` and can be replaced with the custom user content.

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -115,7 +115,7 @@ import VaIcon from '../va-icon/VaIcon.vue'
 import { VaDropdown, VaDropdownContent } from '../va-dropdown'
 
 const VaInputProps = extractComponentProps(VaInput, [
-  'mask', 'returnRaw', 'autosize', 'minRows', 'maxRows', 'type', 'inputmode',
+  'mask', 'returnRaw', 'autosize', 'minRows', 'maxRows', 'type', 'inputmode', 'counter', 'maxLength',
```
m0ksem, line 118: We will need to move VaTimeInput and VaDateInput from input, to VaInputWrapper in other task.

Issue #2018, opened 2022-06-28
Issue with opening va-select element


josip8:
**Vuestic-ui version:** 1.4.4 and 1.4.3

**Steps to reproduce**
Just see one of your examples linked in docs, for example: https://codesandbox.io/s/nutf4c?file=/src/App.vue

![image](https://user-images.githubusercontent.com/26368640/176276968-bbb569e7-0f4a-43ad-a3f5-1f745b5708c6.png)

**What is the expected behavior?**
Dropdown should open underneath the select element without any space in between

**What is the current behavior?**
Dropdown opens down with huge gap between select and dropdown

PR #2019, opened 2022-06-28
[#2018] usePopover - add default position

popov-a-e:
Close https://github.com/epicmaxco/vuestic-ui/issues/2018

## Description

va-dropdown__content-wrapper without `top` and `left` properties might position wrongly, which causes the original issue

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useDomRect.ts
+++ b/packages/ui/src/composables/useDomRect.ts
@@ -24,6 +24,14 @@ export const useDomRect = (target: Ref<HTMLElement | undefined>) => {
     prev = rect
   })
 
+  watch(target, (newVal) => {
+    if (newVal) {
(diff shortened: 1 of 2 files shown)


RVitaly1978 (approved):

packages/ui/src/composables/useDomRect.ts
```diff
@@ -24,6 +24,14 @@ export const useDomRect = (target: Ref<HTMLElement | undefined>) => {
     prev = rect
   })
 
+  watch(target, (newVal) => {
```
m0ksem, line 27: This thing is important. It will update domRect exactly after it appears in DOM.

m0ksem (approved):
Thanks for notice! I made it a bit simpler.

Issue #2006, opened 2022-06-24
Make fixed prop for VaNavbar and VaAppBar


m0ksem:
https://github.com/aluarius/vuestic-ui/commit/286ea45b5be896c3b3cdbfe4ddb216ae8e2f47ed#r76929690

VaHeader and VaAppBar must be relative on default, but fixed on scroll, maybe? Also think about hide-on-scroll prop.

So yes, this two must be reworked.

let's have fixed prop.

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2022

PR #2022, opened 2022-06-29
[#2006] `VaNavbar` & `VaAppBar` refactoring, appearance on scroll logic rework

aluarius:
Close: #2006 

## Description
Fixed & scroll logic rework. Fixed state added for `VaNavbar`. `VaNavbar` & `VaAppBar` refactoring.

## Types of changes


aluarius:
Will reopen merge after `develop` will be merged into `release/1.5.0`.

packages/docs/src/locales/en/en.json
```diff
@@ -144,7 +144,10 @@
         "indeterminateValue": "Overrides a state value that is not set.",
         "borderColor": "Color css style `border` (theme supported options or `HEX`).",
         "clearableIcon": "Sets the cleaning button icon.",
-        "tabindex": "Sets the custom `tabindex`."
+        "tabindex": "Sets the custom `tabindex`.",
```
rustem-nasyrov, line 147: If it is not passed to the inner HTML element inside of component, I think it will be redundant, no?

packages/ui/src/components/va-navbar/_variables.scss
```diff
@@ -2,7 +2,9 @@
   --va-navbar-height: 4.0625rem;
   --va-navbar-padding-left: 1rem;
   --va-navbar-padding-right: 1rem;
-  --va-navbar-transition: background-color 0.3s ease; // sidebar's bg color transitions as well -> consistency
+  --va-navbar-transition: all 0.3s ease;
+  --va-navbar-position: relative;
```
rustem-nasyrov, line 6: Instead of using the relative, I'll suggest you to use the `static` value, because top, right, left and bottom properties won't effect on the HTML-element.

packages/ui/src/composables/useFixedBar.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const transformComputed = computed(() => {
+    if (!props.bottom && !isHiddenComputed.value) { return '' }
+    return props.bottom && isHiddenComputed.value ? 'translateY(100%)' : 'translateY(-100%)'
+  })
+
+  const positionComputed = computed(() => {
+    if (props.fixed) { return 'fixed' }
+    return isHiddenComputed.value ? 'absolute' : undefined
+  })
```
rustem-nasyrov, line 20:
```suggestion
  const positionComputed = computed(
    () => props.fixed ? 'fixed' : isHiddenComputed.value && 'absolute' || undefined
  )
```

packages/ui/src/composables/useScroll.ts
```diff
@@ -1,26 +1,34 @@
...
-  let targetElement: HTMLElement | null
+  let targetElement: HTMLElement | Window | undefined
+
+  const isScrolledDown = ref(false)
+  const prevScrollPosition = ref(0)
+
+  const onScroll = (e: Event) => {
+    const target = e.target as HTMLElement | Window
+    // @ts-ignore
+    const scrollValue = e.target instanceof Window ? target.scrollY : target.scrollTop
```
rustem-nasyrov, line 18: I guess we need to check it with the SSR later.

rustem-nasyrov (changes requested):

packages/ui/src/components/va-navbar/_variables.scss
aluarius: Maxim said that `AppBar` & `NavBar` must be `relative` as default for some reasons.

packages/ui/src/composables/useFixedBar.ts
aluarius: If `props.fixed` -> `fixed`

If `!props.fixed `-> boolean check `isHiddenComputed.value` and, if `true` -> `absolute`. Can't see the way there could be undefined.

May be you have meant double ternary? But it's not good practice.

packages/ui/src/composables/useFixedBar.ts
```diff
@@ -0,0 +1,29 @@
+import { computed, ExtractPropTypes, Ref } from 'vue'
+
+export const useFixedBarProps = {
+  hideOnScroll: { type: Boolean, default: false },
+  fixed: { type: Boolean, default: false },
+  bottom: { type: Boolean, default: false },
+}
+
+export function useFixedBar (props: ExtractPropTypes<typeof useFixedBarProps>, isScrolledDown: Ref<boolean>) {
```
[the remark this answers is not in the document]
aluarius: No, it will be a type. It's not `ExtractComponentsProps`.

rustem-nasyrov (approved):

m0ksem (approved):
We need few docs.

PR #2027, opened 2022-06-30
[#1985] Fix `VaProgressBar` slot position

### LINKED ISSUE #1985: va-progress-bar won't locate the default slot inside itself even if there's enough space for it
image

Per @RVitaly1978 

Currently there are two scenarious for the default's slot behavior for the progress bar. It either is located inside the bar, or above it.

The idea is that if it's possible to fit the slot inside the bar, then it should go there, but if the slot won't fit there (as in case with small- and medium-sized bars), it goes above the bar.

Though the issue is that currently the only possible way to locate the slot inside the bar is to set the bar's size property to `large`. And, as you can see from the screenshot above, when you use some custom size (like 2rem) which is definitely larger than the `large` size, the slot won't go inside the bar. But ideally it should.

AFAICS, there exist 2 potential solutions for this issue:
1. Universal, but clutchy and tricky to implement and support: render the bar first, find its actual rendered size and compare it to the slot's rendered size. If the slot's size is smaller than the bar's size, fit the slot inside. Otherwise, put it on top.
2. The second solution is to provide a new prop, something like boolean `slot-inside`, which tells wheher it's a good idea for Vuestic to try to fit the slot inside the bar instead of locating it above the progress bar (which is, again, the default and the current behavior).

I personally like the 2nd approach more, though there's a fine line between over-implementing things and providing a lot of freedom for the developers. In other words, such a prop might be an overkill. Please, share your thoughts on the topic.

RVitaly1978:
## Description
close #1985

changes:
- [x] add boolean `content-inside` prop
- [x] add boolean `show-percent` prop
- [x] update examples and locales

image

## Types of changes


rustem-nasyrov (approved):

aluarius (approved):

packages/ui/src/components/va-progress-bar/VaProgressBar.vue
```diff
@@ -47,12 +53,12 @@ export default defineComponent({
     buffer: { type: Number, default: 100 },
     rounded: { type: Boolean, default: true },
     reverse: { type: Boolean, default: false },
+    slotInside: { type: Boolean, default: false },
```
m0ksem, line 56: Which slot?

How about to rename it in `contentInside`, `percentInside` or simply `inside`?

m0ksem (approved):

packages/ui/src/components/va-progress-bar/VaProgressBar.vue
RVitaly1978: I choose `contentInside ` ))

PR #2036, opened 2022-07-01
[#2033] Fix `VaDropdown` font

### LINKED ISSUE #2033: Missing fonts for popovers
Popovers use some wierd fonts by default. My browser is configured to use sans- families for missing fonts by default, but it seems to me like popovers are configured for some reason to use serif- fonts by default. Please, check this one.
Screenshot_20220630_182901

RVitaly1978:
## Description
close #2033

changes:
- [x] set font to `va-dropdown__content-wrapper`
- [x] use hook `useBem`

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-dropdown/VaDropdown.vue
+++ b/packages/ui/src/components/va-dropdown/VaDropdown.vue
@@ -87,7 +87,7 @@ export default defineComponent({
       set (val) { statefulVal.value = val },
     })
 
-    const computedClass = useBem('va-dropdown', pick(props, ['disabled']))
+    const computedClass = useBem('va-dropdown', () => pick(props, ['disabled']))
 
     // to be able to select specific anchor element inside anchorRef
     const computedAnchorRef = computed(() => (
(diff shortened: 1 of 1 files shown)


rustem-nasyrov (approved):

packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -85,9 +87,7 @@ export default defineComponent({
       set (val) { statefulVal.value = val },
     })
 
-    const computedClass = computed(() => ({
-      'va-dropdown--disabled': props.disabled,
-    }))
+    const computedClass = useBem('va-dropdown', pick(props, ['disabled']))
```
m0ksem, line 90: How does this reactive? We likely losing reactivity here by passing props value, instead of reference. This must be computed or () => pick(

We also need to double check for this issue in other usages.

m0ksem (changes requested):

aluarius (approved):

Issue #1973, opened 2022-06-20
Impossible to write space in VaTimeInput


m0ksem:
Space opens dropdown with manual input...
Same for enter. How about validate on enter instead of opening modal if manual input?

PR #2060, opened 2022-07-07
feat: allow write space with manual input

rustem-nasyrov:
Close: #1973
image

## Description
Allow to type space symbol when `manual-input` prop truthy.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -13,9 +13,9 @@
     trigger="none"
     @keydown.up.prevent="showDropdown"
     @keydown.down.prevent="showDropdown"
-    @keydown.space.prevent="showDropdown"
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -13,9 +13,9 @@
     trigger="none"
     @keydown.up.prevent="showDropdown"
     @keydown.down.prevent="showDropdown"
-    @keydown.space.prevent="showDropdown"
-    @keydown.esc.prevent="hideDropdown"
+    @keydown.space="showDropdown($event, $props.manualInput, !$props.manualInput)"
     @keydown.enter="!$props.manualInput && showDropdown()"
```
m0ksem, line 17: Not sure if enter must prevent default event, but it looks like it must. No? Have you thought about it?

Anyway, both variants looks good for me.

Issue #1175, opened 2021-10-26
Typography detailed docs 


m0ksem:
Text css utils are missed.

![image](https://user-images.githubusercontent.com/23530004/138876517-0a76ea7b-81e0-4fa7-9884-7bdef235e97e.png)

PR #2064, opened 2022-07-08
docs/typography

m0ksem:
# New docs typogpragy page
image
image

Demo with borders approved by @xx13. It allows user easily find correct code example, instead of searching correct class in long example code. Title heading can also be visible now. It was hard to find it in headers mess before.

Removed useless headings, just watch and copy example!
image

Added text helpers docs.
image
Text helpers must have an interactive example I guess, but we are good to go for now.

# New demo borders
image

For typography it was hard to see where is header and header demo. So I decided to make border around demos like 100% other Frameworks have. It also will fit in interactive playground, which we want to implement later.

# Headers margins mess

We had unlogical margins set to our typography headers to satisfy docs needs. I separated docs styles and our typography module styles. Also, now headers with typography module and va content component looks the same. All values took from Figma. 

# Docs headers margins
In https://github.com/epicmaxco/vuestic-ui/issues/1479 I think what we wanted is to remove big space between title and subtitle. Made this. First subtitle was without margins before.

closes https://github.com/epicmaxco/vuestic-ui/issues/1175


m0ksem:
It still doesn't look perfect. Any thoughts how we can improve these docs and made them easier to read?

rustem-nasyrov:
@m0ksem, maybe we should add a little playground to demonstrate available utility classes?

rustem-nasyrov:
@xx13 we need to increase text saturation for the blockquotes, because it's barely visible, any suggestions? 
image

packages/docs/src/components/DocsNavigation.vue
```diff
@@ -51,66 +65,69 @@ const query = '?query=file=/src/App.vue'
...
-      } catch (e: any) {
-        if (e.message === 'NotAllowedError') {
-          copyButtonState.value = 'error'
-        }
-      }
-
-      setTimeout(() => { copyButtonState.value = 'default' }, 1500)
-    }
+const props = defineProps({
+  code: { type: String as PropType<string>, default: '' },
```
rustem-nasyrov, line 69: As I remember we decided to not use the `PropType`, is it necessary in this property and other ones?

m0ksem: Yeah. We removed them from UI lib, but not docs.

aluarius (approved):

packages/docs/src/components/DocsNavigation.vue
asvae: Probably we want to remove from both places at once. (maybe as a separate issue?)

rustem-nasyrov (approved):

m0ksem:
> @xx13 we need to increase text saturation for the blockquotes, because it's barely visible, any suggestions? image

Yeah. This thing looks inappropriate 😟

xx13:
it should have `secondary` colour which is `#767C88` in designs.

RVitaly1978 (approved):

packages/docs/src/components/DocsExample.vue
```diff
@@ -125,3 +112,27 @@ export default {
   }
 }
 </style>
+
+<style lang="scss" scoped>
+  .docs-example-card {
```
asvae, line 117: Hmmm, why lint didn't work here? Should be no tab.

asvae (approved):
Looks good :).

packages/docs/src/components/DocsExample.vue
m0ksem: Should be two spaces actually, like it is :)

Issue #1725, opened 2022-05-04
Split panel component


m0ksem:
[Live example of split panel](https://sfc.vuejs.org/)
https://github.com/vuejs/repl/blob/main/src/SplitPane.vue

Tag: va-split-pnale, va-panel, va-split etc.

Props: 
- resizable - boolean
- direction- 'horizontal' | 'vertical' - 🤔 maybe even have it as boolean
- startSize, endSize - can be in percents 40%, 60% 

Emits:
- update:startSize, update:endSize

Slots
- start, end (eg. left, right or top, bottom)

We can use VaDivider between start end end parts.

Usage example
```
<va-split-panel direction="horizontal">
   <template #start="{ size }">
      Left
   </template>
   <template #end={ size}">
      Right
   </template>
</va-split-panel>
```

Things to think about:

Maybe this usage example looks more clean. Throw warning if left and bottom slots are present
```
<va-split-panel>
   <template #left> Left </template>
   <template #right> Right </template>
</va-split-panel>
```

m0ksem:
Can be part of https://github.com/epicmaxco/vuestic-ui/issues/1430

m0ksem:
Same as https://github.com/epicmaxco/vuestic-ui/issues/179

asvae:
https://quasar.dev/vue-components/splitter#introduction

PR #2068, opened 2022-07-10
[#1725] `VaSplit` component implementation

aluarius:
Close: #1725 

## Description
New split component implementation.

## Types of changes


packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,189 @@
...
+      <slot name="start" />
+    </div>
+    <div
+      class="va-split__dragger"
+      :style="draggerStyleComputed"
+      @mousedown="startDragging"
+      @dblclick="maximizePanel">
+      <slot name="grabber">
+        <va-divider
+          class="fill-width"
```
m0ksem, line 22: Don't use classes from Grid.css, because user can remove Grid styles from his app.

m0ksem (approved):
Would be nice to have real-world example (like Sidebar and App content.)

Have you tested this on phone? You will need to listen for touchstart, touchend events.

packages/ui/src/components/va-split/VaSplit.demo.vue
```diff
@@ -0,0 +1,145 @@
...
+  components: {
+    VaSplit,
+    VaIcon,
+  },
+}
+</script>
+
+<style lang="scss">
+.va-split-demo {
+  & .va-split {
```
rustem-nasyrov, line 137: The ampersand here is redundant, but you can ignore this suggestion.
```suggestion
  .va-split {
```

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,189 @@
...
+  },
+
+  emits: useStatefulEmits,
+
+  setup: (props, { emit }) => {
+    const { valueComputed } = useStateful(props, emit)
+
+    const splitPanelsContainer = shallowRef<HTMLElement>()
+    const containerSizeComputed = computed(() => {
+      if (!splitPanelsContainer.value) { return undefined }
```
rustem-nasyrov, line 68:
```suggestion
      if (!splitPanelsContainer.value) { return }
```

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,189 @@
...
+
+    const sizePropertyComputed = computed(() => props.vertical ? 'height' : 'width')
+    const getPanelStyle = (position: 'start' | 'end') => ({
+      [sizePropertyComputed.value]: `${position === 'start' ? splitterPositionComputed.value : 100 - splitterPositionComputed.value}%`,
+    })
+
+    const draggerStyleComputed = computed(() => {
+      if (props.disabled) { return {} }
+      let cursor = props.vertical ? 'var(--va-split-vertical-dragger-cursor)' : 'var(--va-split-horizontal-dragger-cursor)'
+      if (isDragging.value) { cursor = 'var(--va-split-dragging-cursor)' }
```
rustem-nasyrov, line 126:
```suggestion
      if (isDragging.value) { return 'var(--va-split-dragging-cursor)' }

      return props.vertical ? 'var(--va-split-vertical-dragger-cursor)' : 'var(--va-split-horizontal-dragger-cursor)'
      
```

rustem-nasyrov (changes requested):
Looks awesome! I have a little suggestion: if the cursor goes out of the splitter container we lose the 'draggable' state, would be better if the user can move the splitter out of the container.

rustem-nasyrov (approved):

packages/docs/src/locales/en/en.json
```diff
@@ -1272,6 +1272,21 @@
       "events": {},
       "methods": {},
       "slots": {}
+    },
+    "VaSplit": {
+      "props": {
+        "vertical": "Changes `va-split` orientation to vertical.",
+        "minSize": "Minimal size of the each panel (%).",
```
m0ksem, line 1279: Do we able to make left minSize as 50%, but right is 16px?

aluarius: No in current state. It's possible to implement this but, honestly, i haven't seen any cases with panels different minimal size.

aluarius: Thanks, now i see.

I will implement this in current pull request scope.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,183 @@
...
+
+    const maximizePanel = () => {
+      if (!props.maximization || props.disabled) { return }
+
+      splitterPosition.value = props.maximizeStart ? 100 - props.minSize : props.minSize
+    }
+
+    watch(valueComputed, (v) => {
+      if (__DEV__ && (v < props.minSize || v > 100 - props.minSize)) {
+        console.warn('Incorrect `modelValue`. Check current `minSize` prop value.')
```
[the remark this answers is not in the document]
aluarius: Changed to the existing util function `warn`. Will update notion a bit later because i have problems with auth.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,261 @@
+<template>
+  <section
+    ref="splitPanelsContainer"
+    aria-label="split panels"
+    class="va-split"
+    :class="classComputed">
+    <div
+      class="va-split__panel"
+      :style="getPanelStyle('start')">
```
RVitaly1978, line 9:
```suggestion
      :style="getPanelStyle('start')"
    >
```
The same below.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,261 @@
...
+    </div>
+  </section>
+</template>
+
+<script lang="ts">
+import { defineComponent, PropType, ref, shallowRef, computed, watch, onMounted } from 'vue'
+import isString from 'lodash/isString.js'
+import isNumber from 'lodash/isNumber.js'
+
+import { useBem, useComponentPresetProp, useStateful, useStatefulEmits, useStatefulProps, useResizeObserver } from '../../composables'
```
RVitaly1978, line 37:
```suggestion
import {
  useBem,
  useComponentPresetProp,
  useStateful, useStatefulEmits, useStatefulProps,
  useResizeObserver,
} from '../../composables'
```

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,261 @@
...
+import { useSplitDragger, useSplitDraggerProps } from './useSplitDragger'
+
+import { warn } from '../../services/utils'
+
+import { VaDivider } from '../va-divider'
+
+type SplitLimit = number | string | string[]
```
RVitaly1978, line 44: maybe it would be better to have the types in a separate file

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,261 @@
...
+          !isNaN(+char) ? numberValue += char : measureValue += char
+        })
+
+      switch (measureValue) {
+        case '%':
+          return +numberValue
+        case 'px':
+          return (+numberValue / containerSize.value!) * 100
+        case 'rem':
+          return ((+numberValue * 16) / containerSize.value!) * 100
```
RVitaly1978, line 100: it might be wrong if rem is not equal to 16 px

RVitaly1978 (changes requested):

packages/ui/src/components/va-split/VaSplit.vue
aluarius: Yes, but there is no another source of truth.

packages/docs/src/page-configs/ui-elements/split/examples/CustomGrabber.vue
```diff
@@ -0,0 +1,40 @@
...
+      </div>
+    </template>
+    <template #end>
+      {{ lorem }}
+    </template>
+  </va-split>
+</template>
+
+<script>
+import { VaSplit } from 'vuestic-ui/src/components/va-split'
```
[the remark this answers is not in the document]
aluarius: For some reasons it's not working without import. I've mentioned it in discussion with you some time ago.

packages/ui/src/components/va-split/VaSplit.vue
aluarius: Mb we can move our base font-size to the var and use it?...

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,281 @@
...
+      if (result?.min && result?.max &&
+        startPanelMinMax.value?.min && startPanelMinMax.value?.max &&
+        (Math.ceil(result.min + startPanelMinMax.value.max) < 100 ||
+          Math.ceil(result.max + startPanelMinMax.value.min) < 100)) {
+        warn('The sum of different panels min and max sizes should be equal to 100% of the container size!')
+        result.min = 100 - startPanelMinMax.value.max
+      }
+
+      return result
+    })
```
RVitaly1978, line 156: what do you think about this:

```ts
    const startPanelLimits = computed(() => getPanelMinMax(props.limits[0]))
    const endPanelLimits = computed(() => getPanelMinMax(props.limits[1]))

    const startPanelMinMax = computed(() => {
      if (startPanelLimits.value?.min + endPanelLimits.value?.min > 100) {
        warn(...)
      }

      return {
        min: startPanelLimits.value?.min,
        max: Math.min(startPanelLimits.value?.max, 100 - endPanelLimits.value?.min),
      }
    }))

    const endPanelMinMax = computed(() => ({
      min: startPanelLimits.value?.min + endPanelLimits.value?.min > 100
        ? 100 - startPanelLimits.value?.min
        : endPanelLimits.value?.min,
      max: Math.min(endPanelLimits.value?.max, 100 - startPanelLimits.value?.min),
    }))
```
and in `getPanelMinMax` we have to check that min/max does not exceed 100

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,281 @@
...
+      startDragging,
+      processDragging,
+      stopDragging,
+      currentSplitterPosition,
+    } = useSplitDragger(containerSize, splitterPositionComputed, props)
+
+    const maximizePanel = () => {
+      if (!props.maximization || props.disabled || !startPanelMinMax.value || !endPanelMinMax.value) { return }
+
+      splitterPosition.value = props.maximizeStart ? 100 - endPanelMinMax.value.min : startPanelMinMax.value.min
```
RVitaly1978, line 184:
```suggestion
      splitterPosition.value = props.maximizeStart ? startPanelMinMax.value.max : 100 - endPanelMinMax.value.max
```

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,281 @@
...
+        (Math.ceil(result.min + startPanelMinMax.value.max) < 100 ||
+          Math.ceil(result.max + startPanelMinMax.value.min) < 100)) {
+        warn('The sum of different panels min and max sizes should be equal to 100% of the container size!')
+        result.min = 100 - startPanelMinMax.value.max
+      }
+
+      return result
+    })
+
+    const splitterPosition = ref(props.modelValue)
```
RVitaly1978, line 158: maybe?
```suggestion
    const splitterPosition = ref(valueComputed.value)
```

packages/ui/src/components/va-split/VaSplit.vue
aluarius: SSR problems. Let's leave it for grooming. Because it seems better to delete rem at all.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,281 @@
...
+
+      // checking min panels size
+      if (splitterPosition.value < startPanelMinMax.value.min) { return startPanelMinMax.value.min }
+      if (splitterPosition.value > 100 - endPanelMinMax.value.min) { return 100 - endPanelMinMax.value.min }
+
+      // checking max panels size
+      if (splitterPosition.value > startPanelMinMax.value.max) { return startPanelMinMax.value.max }
+      if (splitterPosition.value < 100 - endPanelMinMax.value.max) { return 100 - endPanelMinMax.value.max }
+
+      return splitterPosition.value
```
RVitaly1978, line 170:
```suggestion
      return clamp(
        Math.max(startPanelMinMax.value.min, 100 - endPanelMinMax.value.max),
        splitterPosition.value,
        Math.min(startPanelMinMax.value.max, 100 - endPanelMinMax.value.min),
      )
```


where (used from `usePopover`)
```ts
const clamp = (min: number, v: number, max: number) => Math.max(Math.min(v, max), min)
```

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,281 @@
...
+      vertical: props.vertical,
+      dragging: isDragging.value,
+    }))
+
+    return {
+      splitPanelsContainer,
+
+      startDragging,
+      processDragging,
+      stopDragging,
```
RVitaly1978, line 230:
```suggestion
```

RVitaly1978 (changes requested):

packages/ui/src/components/va-split/VaSplit.vue
aluarius: Check again. It's correct logic now.

aluarius: Absolutely the same result. But ok, `max` looks more appropriate.

packages/ui/src/components/va-split/VaSplit.vue
aluarius: Done.

RVitaly1978 (approved):

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,349 @@
...
+      <slot name="end" />
+    </div>
+  </section>
+</template>
+
+<script lang="ts">
+import { defineComponent, PropType, ref, shallowRef, computed, watch, onMounted } from 'vue'
+import isString from 'lodash/isString.js'
+import isNumber from 'lodash/isNumber.js'
+import isUndefined from 'lodash/isUndefined.js'
```
m0ksem, line 44: lol

Check this out: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L12198

For me, using these lodash named methods makes belive that they're not simply checking for undefined, but they are. I would like to remove this.

packages/ui/src/components/va-split/useSplitDragger.ts
```diff
@@ -0,0 +1,58 @@
...
+    const currentPosition = getEventPosition(e, 'mousemove')
+    const distance = currentPosition - dragStartPosition.value
+    currentSplitterPosition.value = dragStartSplitterPosition.value + Math.floor((distance / containerSizeComputed.value!) * 100)
+  }
+
+  const stopDragging = () => {
+    isDragging.value = false
+  }
+
+  onMounted(() => {
```
m0ksem, line 43: useEvent

m0ksem (approved):

m0ksem:
Can we also have some specs like [this](https://github.com/epicmaxco/vuestic-ui/issues/2171)? We will put it later under documentation page. It takes too much time to understand how it works...

packages/ui/src/components/va-split/VaSplit.vue
aluarius: Yea, i know it. In previous implementation this check was used for many times, so i decided to use lodash instead of local check function. Now it's used in only one place and can be removed.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,349 @@
...
+      if (isNumber(v)) { return v }
+
+      v.split('')
+        .filter((char) => char && char !== ' ')
+        .forEach((char) => {
+          !isNaN(+char) ? numberValue += char : measureValue += char
+        })
+
+      switch (measureValue) {
+        case '%':
```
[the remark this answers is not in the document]
aluarius: We have to parse value anyway. So, this suggestion isn't suitable in my opinion.

packages/ui/src/components/va-split/VaSplit.vue
```diff
@@ -0,0 +1,349 @@
...
+          return (+numberValue / containerSize.value!) * 100
+        case 'rem':
+          return ((+numberValue * bodyFontSize.value) / containerSize.value!) * 100
+        case 'any':
+          return ['min', 'snapping'].includes(type) ? 0 : 100
+        case '':
+          return 100
+        default:
+          warn('Invalid limits measure!')
+          return 0
```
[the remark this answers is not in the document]
aluarius: Same as above.

Issue #1646, opened 2022-04-11
 Code split for css


m0ksem:
For now, css from all components bundled in one vuestic-ui.css. So in tree-shaking if user imports only VaTimePicker, then css from all components will be bundled.

For esm build it would be nice components to have separate css imported in js.

dist/VaTimePicker
```
import "./VaTimePicker.css"

export default openBloc(...
```

I think this issue will require writing own rollup plugin or forking rollup-plugin-vue. (Also would be nice to move to vite-plugin-vue)

Currently we use styleinject for each component, what is not working during SSR. (It also likely making double styles in production build).

I propose to import css in dist files from esm build. We 100% sure that webpack and vite can handle css imports from node_modules.

m0ksem:
Can be done with custom rolllup plugin. You can find an example of compiling css with sfc-compiler here: https://github.com/vuejs/repl/blob/main/src/transform.ts#L171

Issue #1545, opened 2022-03-08
Move from rollup to vite


m0ksem:
Since [rollup-plugin-vue](https://github.com/vuejs/rollup-plugin-vue#archived) is no more maintainable, let's move to vite and [@vite/plugin-vue](https://github.com/vitejs/vite/tree/main/packages/plugin-vue).

Should be easy to change bundler after v1.4.0 - this release makes vuestic work with nuxt's vite. Also would be nice to move demos to vite too and remove webpack from source code,

Remember we have bundlers-test package that test project for both vite and webpack. After this FR is complete, we can also remove vite from tests, because it will be used in demos and docs.

Derranion:
Yeah and we can also use the same `demo` inside our `bundlers tests`.

Issue #1489, opened 2022-02-08
Normalize.css -styles should be better scoped


anttimateria:
**Vuestic-ui version:** 1.3.4

**Steps to reproduce**
Add any Vuestic component to a page or register library globally, and you see `normalize.css`-styles applied globally, which is not desired:

![Screenshot 2022-02-07 at 12 54 01](https://user-images.githubusercontent.com/18732832/152934018-522fc3c3-da78-4600-9a9f-feba88cdb075.png)

**What is the expected behavior?**

The normalized styles should apply only to Vuestic-components i.e. they should be **scoped**. 

**What is the current behavior?**
`normalize.css`-styles are applied globally, which might override/conflict some other page styles.

**Other information**

m0ksem:
Changing 
```ts
import 'vuestic-ui/dist/vuestic-ui.css'
```

to

```ts
import 'vuestic-ui/dist/styles/essentials.css'
import 'vuestic-ui/dist/styles/grid/grid.css'
import 'vuestic-ui/dist/styles/resources/typography.css'
```
should make a trick. 



For new project, you can use `vue add vuestic`, use treeshaking and unselect `Normalize CSS` option in prompts.

asvae:
We have do document this, as far as I see users won't have a way to know we have that functionality in place already.

Further development on optional/core is [here](https://github.com/epicmaxco/vuestic-ui/issues/1426), we plan to tackle it next milestone.

PR #1801, opened 2022-05-18
feat/ CSS code split. Move to vite from rollup.

aluarius:
close: #1545 
close: #1646 
closes https://github.com/epicmaxco/vuestic-ui/issues/1489

## Description
Rollup production build has been fully changed to Vite (`yarn build` to test it). CSS spliting added (for ESM build, including import to separate components).

New average build time is 27 s (vs previous 156 s for my pc).

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

PR #2081, opened 2022-07-12
[#1545] Copy of the Vite build for devel branch

aluarius:
Moved from https://github.com/epicmaxco/vuestic-ui/pull/1801

close: #1545 
close: #1646 
close: #1489

## Description
Rollup production build has been fully changed to Vite (`yarn build` to test it). CSS spliting added (for ESM build, including import to separate components).


packages/ui/build/vite/build.mjs
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * parallel build for all formats
+ */
+await Promise.all([
+  $`vite build --config ./build/vite/configs/vite.cjs.js`,
+  $`vite build --config ./build/vite/configs/vite.iife.js`,
+  $`vite build --config ./build/vite/configs/vite.esm.js`,
+  $`vite build --config ./build/vite/configs/vite.styles.js`,
+  $`yarn build:types`,
```
m0ksem, line 40:
```suggestion
  $`npm run build:types`,
```

Maybe we will move to pnpm, so we will not face potential issue.

packages/ui/build/vite/build.mjs
```diff
@@ -0,0 +1,91 @@
...
+])
+
+/**
+ * renaming css files from esm-node build (deleting hash from name)
+ */
+await readDirRecursive('./dist/esm-node/src/components').forEach((filePath) => {
```
m0ksem, line 46:
```suggestion
/** Renaming css files from esm-node build (deleting hash from name) */
await readDirRecursive('./dist/esm-node/src/components').forEach((filePath) => {
```

This will work too.

packages/ui/build/vite/utils.mjs
```diff
@@ -0,0 +1,36 @@
...
+  const pathBase = splitPath.slice(0, splitPath.length - 1).join('/')
+
+  return { splitFileName, pathBase }
+}
+
+export const kebabToPascalCase = (name) => {
+  return name.split('-').map((el) => el.charAt(0).toUpperCase() + el.slice(1)).join('')
+}
+
+// vite/rollup, for example, generates assets names as `[path]/[name].[hash].[extension]`
```
m0ksem, line 28: Don't you able to remove them with rollup?

```
      entryFileNames: `[name].[ext]`,
      chunkFileNames: `[name].[ext]`,
```

m0ksem: I mean, you can fix it with vite's `assetFileNames`

packages/ui/build/vite/configs/vite.cjs.js
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vite'
+import getViteConfig from '../common-config'
+
+export default ({ mode }) => {
+  return defineConfig({
+    ...getViteConfig(mode === 'production', 'cjs'),
```
m0ksem, line 6: Isn't mode always production? We use these configs for build only.

packages/ui/build/vite/common-config.js
```diff
@@ -0,0 +1,79 @@
...
+      lib: {
+        entry: resolver(process.cwd(), 'src/main.ts'),
+        fileName: () => 'main.js',
+        formats: [format],
+
+        // only for iife/umd
+        name: 'vuestic',
+      },
+
+      ...commonOptions,
```
m0ksem, line 66: You can use lodash/deepMerge and use getViteConfig everywhere.

packages/ui/build/vite/common-config.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // disable mangling class names (for vue class component)
+    keep_classnames: true,
+
+    // disable mangling functions names
+    keep_fnames: true,
+  },
+}
+
+export default function getViteConfig (isProduction, format) {
```
m0ksem, line 46:
```suggestion
export default function createViteConfig (isProduction, format) {
```

It is a fabric.

packages/ui/build/vite/common-config.js
```diff
@@ -0,0 +1,79 @@
...
+export const vuePlugin = vue({
+  isProduction: true,
+  exclude: [/\.md$/, /\.spec\.ts$/, /\.spec\.disabled$/],
+})
+
+export const resolve = {
+  alias: {
+    '@': resolver(process.cwd(), 'src'),
+    '~/': resolver(process.cwd(), 'src'),
+    '~normalize.css/normalize.css': 'normalize.css/normalize.css',
```
m0ksem, line 21: Would be nice to get rid of it. But I don't know how, yet.

packages/ui/package.json
```diff
@@ -51,6 +51,10 @@
     "normalize.css": "^8.0.1"
   },
   "devDependencies": {
+    "vite": "^2.9.8",
```
m0ksem, line 54: Let's have deps sorted, how package manager does.

In VS Code select deps, run `Ctrl + Shift + P` and `Sort`

m0ksem (changes requested):
This thing looks a bit complicated, I believe we can make it easier. At least a bit.

Have not tried it to run, yet.

packages/ui/build/vite/common-config.js
m0ksem: normalize.css can be included in final bundle

packages/ui/build/vite/build.mjs
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env zx
```
asvae, line 1: couldn't that use `.ts` similar to release script?

Issue #715, opened 2021-05-10
Checkbox visual update


xx13:
- [x] Change padding between a checkbox and a label from `4px` to `8px`
- [x] `Mixed` option was added with a minus icon
- [x] Default font style should match `regular medium` style which is `15/20`
- [x] #2092
- [x] Set border-radius to 0 (0.125 rem according to the figma)
- [x] Change `display` to `inline-block`
- [x] Remove horizontal paddings

The reset items are for 1.5.0.

[Visual references](https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/vuestic-ui?node-id=188%3A0)

m0ksem:
We also need to remove extra paddings
![image](https://user-images.githubusercontent.com/23530004/178080482-c0ef6d8c-ae8e-44c2-8c7e-e029ca5a932f.png)

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2346

PR #2085, opened 2022-07-12
[#715] `VaCheckbox` visual updates

aluarius:
Related: #715

## Description
Minor `VaCheckbox` visual updates.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-checkbox/VaCheckbox.demo.vue
+++ b/packages/ui/src/components/va-checkbox/VaCheckbox.demo.vue
@@ -1,5 +1,5 @@
 <template>
-  <VbDemo class="va-input-demo">
+  <VbDemo class="va-checkbox-demo">
     <VbCard title="Default">
       <va-checkbox
         v-model="value"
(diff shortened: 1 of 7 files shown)


rustem-nasyrov (approved):

m0ksem:
image

We will destroy users layout which rely on this padding on right!
---

image

Is it planned to remove color from error messages?
---
image
Small ugly space

m0ksem:
Revert display inline-block.

rustem-nasyrov (approved):

packages/ui/src/components/va-checkbox/VaCheckbox.demo.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <VbDemo>
+  <VbDemo class="va-input-demo">
```
m0ksem, line 2:
```suggestion
  <VbDemo class="d-flex direction--column">
```

But, do we actually need it if checkbox is block?

m0ksem (approved):

packages/ui/src/components/va-checkbox/VaCheckbox.demo.vue
aluarius: I've decided not delete it - we will need it after we make checkbox inline.

Issue #2073, opened 2022-07-11
Range slider is changeable even without v-model.


asvae:
**Vuestic-ui version:** 1.4.6

**Steps to reproduce**

* Use this syntax:

  ```html
  <va-slider :model-value="[20, 80]" range />
  ```

* Click on slider in range
* See:

  <img width="443" alt="image" src="https://user-images.githubusercontent.com/5394573/178250319-f4413d50-5087-47a9-8c72-ec54bcfd1d8d.png">

**What is the expected behavior?**

* Click shouldn't change anything visually.

arkmadj:
@asvae I don't really understand. Can you explain a bit more?

arkmadj:
Yeah. I can see the issue now. Thanks

PR #2087, opened 2022-07-13
[#2073] `VaSlider` range static `modelValue` fix

aluarius:
Close: #2073

## Description
It wasn't trivial at all :)

For some reason first update emit, even in case it's static, changed `modelValue`.

## Types of changes


RVitaly1978 (approved):

aluarius:
> I'm assuming this is a problem because we have a `setCurrentValue` method that mutate `currentValue` (used `splice` array method). Maybe we can change this method?

Discussed. Not optimal for this issue, but interesting for discussion about value-by-link variables mutations.

RVitaly1978 (approved):

rustem-nasyrov (approved):

Issue #2088, opened 2022-07-13
Multi select does not handle text properly when defining value-by


Luffyyy:
Basically using multi select yields an unexpected result that doesn't happen with the single select, I expect the text to show up and not the values.

**Vuestic-ui version:** 1.4.6

**Steps to reproduce**
Try using the component like so:
```vue
<template>
    <!-- Works -->
    <va-select v-model="test1" :options="options" value-by="value"/>
    <!-- Doesn't work -->
    <va-select v-model="test2" :options="options" value-by="value" multiple/>
</template>

<script setup>
const test1 = ref();
const test2 = ref([]);
const options = [
    { 
        text: 'Hello',
        value: 1
    },
    { 
        text: 'World',
        value: 2
    },
    { 
        text: 'Test',
        value: 3
    }
];
</script>
```
**What is the expected behavior?**
(Works if we remove value-by, but I am interested in defining it)
![image](https://user-images.githubusercontent.com/11510099/178777725-6b11eff5-5cb7-4ded-912f-f83cb315e09c.png)

**What is the current behavior?**
![image](https://user-images.githubusercontent.com/11510099/178777545-bba8840b-06a1-4bc1-8c1a-7c901db1d6e7.png)

PR #2101, opened 2022-07-14
Fix multiple select value-by prop & nuxt version bump...

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2088

image

Also, fixed issue when multiple select ignore clearable prop, so any multiple select adds clear icon.
image


m0ksem:
And... I forget to push nuxt's bump :)

Issue #2090, opened 2022-07-13
Add nuxt-docs readme


m0ksem:
Write here that we don't use it for now...

PR #2103, opened 2022-07-14
chore: remove nuxt docs

asvae:
Temporarily removing nuxt docs as they give us a lot of duplication and are generally hard to support.

closes https://github.com/epicmaxco/vuestic-ui/issues/2090


m0ksem:
I will save here current commit hash `10043bd249f38116b9dda0ec789b94cb1643c978` so we can easily come back from PR.

m0ksem (approved):

PR #2105, opened 2022-07-14
Css code split docs and essentail css in build

m0ksem:
Added essential.css to dist:
image

Added small docs.
image


aluarius (approved):

packages/docs/src/page-configs/getting-started/tree-shaking/code-examples/css-code-split.ts
```diff
@@ -0,0 +1,8 @@
+export const cssCodeSplit = `
+// Required
+import 'vuestic-ui/styles/essential.css'
+// Optional
+import 'vuestic-ui/styles/typography.css'
+import 'vuestic-ui/styles/grid.css'
```
asvae, line 6: `.mx-0` is not exactly grid, rather spacers. But probably fine for now as we're gonna overhaul these regardless.

m0ksem: Yeah. Let's make more CSS modules later.

Issue #1969, opened 2022-06-17
VaDropdown has a lot of divs


m0ksem:
![image](https://user-images.githubusercontent.com/23530004/174395510-a2951ca9-ca4f-4cc5-a56c-2c3a3351913b.png)

Would be nice to have simply

```vue
<template>
   <slot v-bind="{ bind, show, toggle, hide }" />
   <teleport>
      ....
</template>
```

This will remove a lot of extra divs, so it is just Dropdown :)

We don't have VaDropdown docs yet, some anything can be changed!

Make sure update all components where VaDropdown is used.

Issue #1494, opened 2022-02-09
contextual popover on right click


ralpheichelberger:
**Is your feature request related to a problem? Please describe.**
I like to provide a simple context popover in e.g. my va-data-tabel that shows more info, options with buttons, etc.

**Describe the solution you'd like**
Similar to the default context menu of e.g. chrome, it would open a popover starting at the postion of the mouse, or where it is in the viewport.
This could be done in the same way va-popover is designed by wraping the content of that "context-popover".

I imagine this component like this:

```vue
<template>
  <va-context-popover
    color="primary" 
    text-color="white"
    title="Product options"
  >
    <va-button>
        @click="doOption1">
            Option 1
    </va-button>
    <va-button>
        @click="doOption1">
            Option 1
    </va-button>    
  </va-context-popover>
</template>
```

So it is basically a popover but positioned where you right click the mouse and is space on the viewport.

asvae:
Thanks. Interesting.

Thinking if we can reuse existing popover, while adding `anchor="cursor"` slot or something similar.

ralpheichelberger:
Hey! How are you guys! Are you considering to add this feature some time soon?

asvae:
Would be great for mini release after we're done with current one.

Issue #1154, opened 2021-10-09
Dropdown docs


m0ksem:
Need docs about VaDropdown component with references to PopperJS

---

In nuxt we remove poperjs, so need more details for this issue from me

wossig:
Is there missing Dropdown doc on [https://vuestic.dev/en/ui-elements](url)?  I only see `Button Dropdown`.

otatar:
Really need docs about VaDropdown...

Issue #890, opened 2021-06-25
Close Dropdown on window blur


m0ksem:
Reason: 
When a user clicks on the iframe no window click event does not emit. Instead, there is a focus on iframe and blur for a window. Also, if we can we need to check if the new focus target is an iframe. Because blur emits when you can click on the Windows Task bar. 

It's not important but will be nice to have.

DieselTripleX:
I could do it :)

asvae:
Sounds reasonable!

PR #2109, opened 2022-07-16
docs: dropdown

m0ksem:
While I was making docs found a lot of issues... Fixed them here.

closes https://github.com/epicmaxco/vuestic-ui/issues/1494
closes https://github.com/epicmaxco/vuestic-ui/issues/1969
closes https://github.com/epicmaxco/vuestic-ui/issues/1154
closes https://github.com/epicmaxco/vuestic-ui/issues/890
closes https://github.com/epicmaxco/vuestic-ui/issues/849
closes https://github.com/epicmaxco/vuestic-ui/issues/2125

image


packages/docs/package.json
```diff
@@ -68,6 +68,7 @@
     "stylelint": "^13.13.1",
     "stylelint-config-standard": "^22.0.0",
     "ts-jest": "^26.5.4",
+    "ts-node": "^10.9.1",
```
m0ksem, line 71: We need ts-node for plop to generate docs pages.

packages/ui/src/components/va-input/components/VaInputWrapper/_variables.scss
```diff
@@ -3,7 +3,7 @@
   --va-input-wrapper-background: var(--va-background, #f5f9fb);
   --va-input-wrapper-background-opacity: 1;
   --va-input-wrapper-border-color: var(--va-input-bordered-color);
-  --va-input-wrapper-min-width: 280px;
+  --va-input-wrapper-min-width: 240px;
```
m0ksem, line 6: @rustem-nasyrov . I changed min width to 240px so it perfectly fit DatePicker in DateInput.

Do you see anything wrong here?

packages/ui/src/composables/useMousePosition.ts
```diff
@@ -0,0 +1,25 @@
+import { useClientOnly } from './useClientOnly'
+import { onMounted, onBeforeUnmount, Ref, reactive } from 'vue'
+
+export const useMousePosition = (el?: Ref<GlobalEventHandlers | undefined | null>) => {
+  if (!el) {
+    el = useClientOnly(() => window as GlobalEventHandlers)
```
m0ksem, line 6: This part can be changed to useWindow.

packages/ui/src/composables/useMousePosition.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const mouse = reactive({ x: 0, y: 0 })
+
+  const onMouseMove = (e: MouseEvent) => {
+    mouse.x = e.clientX
+    mouse.y = e.clientY
+  }
+
+  onMounted(() => {
+    el?.value?.addEventListener('mousemove', onMouseMove)
```
m0ksem, line 17: Better watch on target if it is not undefined.

packages/ui/src/components/va-input/components/VaInputWrapper/_variables.scss
rustem-nasyrov: No, I don't.

packages/ui/src/composables/usePopover.ts
```diff
@@ -109,14 +109,23 @@ const getAutoPlacement = (placement: Placement, coords: Coords, content: DOMRect
...
 }
 
+const findFirstRelativeParent = (el: Element | null) => {
+  while (el) {
+    if (window.getComputedStyle(el).position === 'relative') { return el }
+    el = el.parentElement
+  }
+
+  return document.body
+}
```
RVitaly1978, line 119:
```suggestion
const findFirstRelativeParent = (documentRef: Ref<Document>, root: string) => {
  let el = documentRef.value.querySelector(root)

  while (el) {
    if (window.getComputedStyle(el).position === 'relative') { return el }
    el = el.parentElement
  }

  return documentRef.value.body
}
```

RVitaly1978 (changes requested):
- bug (or feature) on hover: content doesn't disappear after hover:

chrome-capture-2022-6-20

- `stickToEdges` doesn't work for all components (select, time-input, date-input, popover):

https://user-images.githubusercontent.com/55198465/179948159-42da832c-6123-4849-8749-9969bcc1c3f6.mov

packages/ui/src/components/va-dropdown/hooks/useAnchorSelector.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const computedAnchorRef = computed<HTMLElement | undefined>({
+    set (v) { anchorRef.value = v },
+    get () {
+      if (props.innerAnchorSelector) { return anchorRef.value?.querySelector(props.innerAnchorSelector) || anchorRef.value }
+
+      return props.anchorSelector
+        ? document.value?.querySelector(props.anchorSelector) || anchorRef.value
+        : anchorRef.value
+    },
```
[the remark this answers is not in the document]
m0ksem: I believe computed things must be cached, so it doesn't run get multiple times. But, yeah, the previous idea was different, so I will refactor it.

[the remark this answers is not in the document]
m0ksem: So. Computed values are actually cached until some of it's references is changed, so there will not be additional calculation on get. BUT! The important thing is that `set` [will not be recalled ](https://sfc.vuejs.org/#eNp9UcFugzAM/RUrF6hKobt2pdJu233HXDrqdnSQRE5gk1D+fU4C1bpJExKK7efn5+dJPBlTjgOKndjbhlrjwKIbzEGqtjeaHExAeC6g0b0ZHJ7Aw5l0Dxk3ZVJJ1WhlHZcH5ZCgDuh8u3oMJWZ6Cdnx2OX5CuoDTFLBgi05PSCsa3hgtC/gYbuNjQALa6te8cvNpNkzdp3OAiJVXaotyvJIzjPzcZUGwUwwT6phhDVk/K3vNcSZvgj/C7ffuomdIHVHsiB9FCqZ4we9FGb4kCJtsK+Sn+wkBw570x0dcgTSTVMS733A3WqiEMn0TX805dVqxWeJWuRcsFLsFnVS8AlCLMW7c8buqsqem3DMqy01XSp+lcRrtj2WaPvNG+lPyztfmSUuO3NUnByRNoTqhIT0H+cv6B/e6A/bI/w3vmjNOQ== )when it's dependencies updated.

It is not much important there, but if user will have a dynamic anchor prop the reactivity will be lost.

m0ksem:
We need to make user able to remove "prevent-overflow" feature. Prevent overflow actually force us to use requestAnimationFrame which is not usually needed so we make up performance.

m0ksem:
Dropdown must react different on events if it touch events. For example:
- trigger hover must open on touch click, since there is no hover. 
- trigger right-click must work if it is long press.

PR #2113, opened 2022-07-19
VaCounter validation

### LINKED ISSUE #1976: Validation `min`, `max` and `step` props of `VaCounter`
We can use the same approach as in `va-slider`.

damirJa:
Fix #1976
## Description
- quick fix input auto width
- add counter validation

## Types of changes


packages/ui/src/components/va-counter/validateCounter.ts
```diff
@@ -0,0 +1,31 @@
+import { warn } from '../../services/utils'
+
+export const validateCounter = (value: string | number, step: number, min?: number, max?: number) => {
+  const val = Number(value)
+  if (!Number.isNaN(val)) {
```
aluarius, line 5:
```suggestion
  if (!Number.isNaN(val)) {
    warn('The value is not a number or cannot be reduced to a number.')
    return
  }
```

To decrease `if` nesting level.

aluarius (approved):

packages/ui/src/components/va-counter/validateCounter.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (min && max) {
+      if (max < min) {
+        warn(`The maximum value (${max}) can not be less than the minimum value (${min}).`)
+      }
+
+      if (min > max) {
+        warn(`The minimum value (${min}) can not be greater than the maximum value (${max}).`)
+      }
```
RVitaly1978, line 13:
```suggestion
      if (max < min) {
        warn(`The maximum value (${max}) can not be less than the minimum value (${min}).`)
      }
```
one of these warnings is enough

packages/ui/src/components/va-counter/validateCounter.ts
```diff
@@ -0,0 +1,31 @@
...
+
+      if ((max - min) % step !== 0) {
+        warn(`Step ${step} is illegal. Counter is non-divisible (Min:Max-${min}:${max}).`)
+      }
```
RVitaly1978, line 17:
```suggestion
```
we don't need this condition because the counter has some logic to round up to the nearest valid value.
But maybe we should check if the `step` is greater than `max - min`

packages/ui/src/components/va-counter/validateCounter.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+    if (min && val < min) {
+      warn(`The value of the counter is ${val}, the minimum value is ${min}, the value of this counter can not be less than the minimum value`)
+    }
+    if (max && val > max) {
+      warn(`The value of the counter is ${val}, the maximum value is ${max}, the value of this counter can not be greater than the maximum value`)
+    }
+    if (val % step !== 0) {
```
RVitaly1978, line 25:
```suggestion
    if ((val - min) % step !== 0) {
```

packages/ui/src/components/va-slider/validateSlider.ts
```diff
@@ -6,15 +6,15 @@ export const validateSlider = (value: number | number[], step: number, min: numb
...
   }
 
   if (min > max) {
-    warn('The minimum value can not be greater than the maximum value.')
+    warn(`The minimum value (${min}) can not be greater than the maximum value (${max}).`)
   }
 
   if ((max - min) % step !== 0) {
-    warn('Step is illegal. Slider is nondivisible.')
+    warn(`Step ${step} is illegal. Counter is non-divisible (Min:Max-${min}:${max}).`)
```
RVitaly1978, line 17:
```suggestion
    warn(`Step ${step} is illegal. Slider is non-divisible (Min:Max-${min}:${max}).`)
```

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -294,6 +295,8 @@ export default defineComponent({
 
     const marginComputed = computed(() => safeCSSLength(props.margins))
 
+    validateCounter(props.modelValue, props.step, props.min, props.max)
```
RVitaly1978, line 298: This checks the props only once. But props can change dynamically

RVitaly1978 (changes requested):

RVitaly1978 (approved):

m0ksem (approved):

Issue #2132, opened 2022-07-22
Get rid of VaContent extra style dependencies. 


m0ksem:
I mean... Do we even need VaContent?

asvae:
<img width="1146" alt="image" src="https://user-images.githubusercontent.com/5394573/181485531-6562272e-45f4-4101-9e92-0cf66b9c1b43.png">


We decided to keep class, but remove component.

PR #2145, opened 2022-07-25
Fix/isolate typography

### LINKED ISSUE #2131: Typography missing font-family
image

For now, user need to wrap with `.va-content` so typography will work. Need to add font-family everywhere in typography module. For example, simply add 
```css
body, * { font-familty: var(--va-font-family); }
```

m0ksem:
image

Changed code-snippet color to css variables instead of some old color.

Updated color contrast in blockqoute.
image

There was an issue that VaContent import all global styles, so if you want to tree-shake typography, you also need to remove VaContent.

closes https://github.com/epicmaxco/vuestic-ui/issues/2132
closes #2131


m0ksem:
We can make docs later, while typography service is implemented. So this is hot fix for nuxt - we need to make user able to easy remove typography CSS.

Issue #1944, opened 2022-06-14
Roadmap out of date


Leon0402:
Version 1.4.0 was already released, but is still marked as "in development". 

![image](https://user-images.githubusercontent.com/33933142/173626741-5bf4eed4-d452-491a-8441-d98ea61adc94.png)

Would be nice to see what is planned for 1.5 additionally.

asvae:
![image](https://user-images.githubusercontent.com/5394573/180246556-ceba434f-d822-4cff-a2e3-8be1392af135.png)

asvae:
Would be great to handle this on high level of quality. We can show our plans similarly to what developer do in games.

On screenshot you can see some of the plans we're very committed for. So these could be easily shaped into timeline.

PR #2147, opened 2022-07-25
docs/roadmap-1-5-0

m0ksem:
image

closes https://github.com/epicmaxco/vuestic-ui/issues/1944


packages/docs/src/components/DocsComponent.vue
```diff
@@ -1,6 +1,6 @@
 <template>
```
asvae, line 1: Perhaps we can also mark version for now? And also group items somehow?

image

Would be also nice to link issue, so that user can hover/click on card.

PR #2162, opened 2022-07-26
Refactroing/cleaning

m0ksem:
Remove vue-cli-example package. We do not support it, it is a bit outdated. Not sure why we need it.

Removed data folder from packages/ui.

I also think it would be nice to move assets to .github in root and vue-book to different package. Making packages/ui cleaner will make easier new contributors to join !


asvae (approved):
Awesome 😎

PR #2165, opened 2022-07-27
[#2163] Fix `VaInputWrapper` `min-width`

### LINKED ISSUE #2163: Issue to set `min-width` and `width` rules to `VaInput`, `VaSelect`, `VaTimePicker`, `VaDatePicker`
**Vuestic-ui version:** 1.4.7

This not works:
```ts
<va-input style="min-width: 150px; width: 100%" />
<va-select style="min-width: 150px; width: 100%" />
<va-time-input style="min-width: 150px; width: 100%" />
<va-date-input style="min-width: 150px; width: 100%" />
```

RVitaly1978:
## Description
close #2163 

changes:
- [x] add global css-variable for the form-elements `min-width`
- [x] update `min-width` rule for components
- [x] change `display: flex` with `display: inline-block` for the `va-date-input` (according to other "input" components)

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -19,6 +19,7 @@
   >
     <template #anchor>
       <va-input-wrapper
+        class="va-time-input__anchor"
         v-bind="computedInputWrapperProps"
         @click="toggleDropdown"
         @keydown.enter.stop="toggleDropdown"
@@ -78,7 +79,6 @@
           <va-icon
             v-else-if="!$props.leftIcon"
             role="button"
-            class="va-dropdown__icons__reset"
             aria-label="toggle dropdown"
             aria-hidden="false"
             :tabindex="iconTabindexComputed"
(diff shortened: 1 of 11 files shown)


rustem-nasyrov (approved):

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -78,7 +79,6 @@
           <va-icon
             v-else-if="!$props.leftIcon"
             role="button"
-            class="va-dropdown__icons__reset"
```
aluarius, line 81: Won't it remove focus-outline state?

aluarius (approved):

packages/ui/src/components/va-time-input/VaTimeInput.vue
RVitaly1978: This state sets in `va-icon` component:
```ts
.va-icon {
  vertical-align: var(--va-icon-vertical-align);
  user-select: var(--va-icon-user-select);

  &[tabindex]:not([tabindex^="-"]) {
    &:focus {
      @include focus-outline;
    }
  }

  ...
```

m0ksem (changes requested):
I think we better remove min-width completely and use size of icons and subtract it from input-wrapper width. This way we can achieve removing jumping while validation icons are added to input, so input is not become bigger.  That would be a different task and different PR I think.

240px just sucks in practice.

PR #2166, opened 2022-07-27
feat: release improvements

### LINKED ISSUE #2102: Release script improvements
- [x] Only use "upstream" as remote name.

We want to have a number of early checks to prevent manual fixes:
- [x] Check node has correct version. Attempt to nvm it if not.
- [x] Check github doesn't have the tag already
- [x] Check npm doesn't have the version already
- [x] Check "upstream" is valid remote

asvae:
close #2102

Mostly went through the issue.

One thing of interest is that I added `.nvmrc` to project root, which would allow us to consistently use the same version of node as on circleci.


m0ksem (approved):

m0ksem:
@asvae, I would also make `npm publish` the last part of release script. Because `publish` can not be undone, but we can remove git tag and commit from remote. Looks like it is not importantant since we have checks, But...

Issue #2168, opened 2022-07-28
Remove `min-width` completely from `va-input-wrapper` css rules


RVitaly1978:
I think we better remove min-width completely and use size of icons and subtract it from input-wrapper width. This way we can achieve removing jumping while validation icons are added to input, so input is not become bigger.  That would be a different task and different PR I think.

240px just sucks in practice.

_Originally posted by @m0ksem in https://github.com/epicmaxco/vuestic-ui/pull/2165#pullrequestreview-1053168229_

PR #2179, opened 2022-07-29
Fix/input field layout

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2168

# Cool hack

I assume it works like this:
```scss
.parent {
  max-width: 100%;
  display: inline-flex;
  
  .child {
     width: var(--va-default-width);
     flex: 1;
  }
}
```

flex: 1 makes child to follow parent width. But, if parent doesn't have width or `auto`, then child width will be used.

# Before/after

image
image

---

image

We still have the same width for inputs in form even with icon like we had with min-width, but we are able to make them smaller

Set .va-input width 100px before:
image

Now: (last two inputs have margin-top - it's OK in this demo)
image

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/date-input/examples/View.vue
+++ b/packages/docs/src/page-configs/ui-elements/date-input/examples/View.vue
@@ -1,26 +1,25 @@
 <template>
   <div class="row">
-    <div class="mr-4">
-      <va-date-input
-        v-model="value"
-        v-model:view="dayView"
-        label="Day"
-      />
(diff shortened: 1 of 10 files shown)


RVitaly1978 (approved):
All good!

Don't forget to tweak styles in `va-time-input` and `va-date-input`:
- instead
```scss
&__anchor {
  min-width: auto;
  width: 100%;
}
```
- should be
 ```scss
&__anchor {
  flex: 1;
}
```

PR #2186, opened 2022-07-30
fixed roadmap responsiveness on mobile

### LINKED ISSUE #2183: Roadmap mobile version is not awesome
**Vuestic-ui version:** 1.4.8

image

ameer-khan-ashraf:
Close #2183 

## Description
Added two media query in the roadmap.vue file to fix the roadmap in mobile view

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/introduction/roadmap/components/roadmap.vue
+++ b/packages/docs/src/page-configs/introduction/roadmap/components/roadmap.vue
@@ -82,10 +82,18 @@ $gap: 1rem;
     display: flex;
     width: 100%;
 
+    @media screen and (max-width: 479px) {
+      flex-direction: column;
+    }
+
     .docs-roadmap__column {
       &-wrapper {
         width: 33%;
         padding: calc($gap / 2);
+
+        @media screen and (max-width: 479px) {
+          width: 100%;
+        }
       }
 
       & > *:not(h6) {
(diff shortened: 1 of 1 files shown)


packages/docs/src/page-configs/introduction/roadmap/components/roadmap.vue
```diff
@@ -82,10 +82,18 @@ $gap: 1rem;
...
+    }
+
     .docs-roadmap__column {
       &-wrapper {
         width: 33%;
         padding: calc($gap / 2);
+
+        @media screen and (max-width: 479px) {
+          width: 100%;
+        }
```
m0ksem, line 96:
```suggestion
    .docs-roadmap__column {
      &-wrapper {
        width: 100%;
        min-width: 300px;
        flex: 1;
        padding: calc($gap / 2);
```

packages/docs/src/page-configs/introduction/roadmap/components/roadmap.vue
```diff
@@ -82,10 +82,18 @@ $gap: 1rem;
     display: flex;
     width: 100%;
 
+    @media screen and (max-width: 479px) {
+      flex-direction: column;
+    }
```
m0ksem, line 87:
```suggestion
    display: flex;
    flex-wrap: wrap;
    width: 100%;
```

m0ksem (changes requested):
Thanks!

I would suggest you make these columns responsive to any screen size, instead of anchoring to a specific screen size.

Having flex-wrap will move column that not fit downside, flex 1 will force column to take as much space as it can, and min width will force columns to be readable and move other columns to down.

ameer-khan-ashraf:
Thanks for the heads up. I have updated the code and pushed another commit.

m0ksem (approved):

PR #2198, opened 2022-08-02
feat: add markdown support to the doc pages

### LINKED ISSUE #2112: .md support for docs
The idea is to have version of documentation, that doesn't support translations and is less powerful, but is very easy to work with as a developer. This documentation could be used for internal things, such as composables, tests, analytics etc. We could also move some things from notion.

For usage - it won't be that different from our current documentation. It would have a route on https://vuestic.dev/, maybe slightly hidden.

- [ ] Provide a way to use md documentation (maybe as a documentation block?)
- [ ] Connect existing md files to that documentation
- [ ] document this approach in contribution docs

rustem-nasyrov:
Close #2112
## Description
Added markdown files support for the documentation pages.

You can import whole markdown file to the `PageGenerationHelper.markdown` method:

```ts
import { ApiDocsBlock } from '@/types/configTypes'
import { PageGenerationHelper } from '@/helpers/DocsHelper'

// import example markdown file:
import docs from './docs.md'

const block = new PageGenerationHelper(__dirname)

const config: ApiDocsBlock[] = [
  block.markdown(docs),
]

export default config
```

Or you can provide markdown code directly:

```ts
import { ApiDocsBlock } from '@/types/configTypes'
import { PageGenerationHelper } from '@/helpers/DocsHelper'

const block = new PageGenerationHelper(__dirname)

// markdown string:
const docs = `
# Hello, World!
Example text
|Column|Column|
|---|---|
|Text|Text|
`

const config: ApiDocsBlock[] = [
  block.markdown(docs),
]

export default config
```

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/dropdown/specs.md
+++ b/packages/docs/src/page-configs/ui-elements/dropdown/specs.md
@@ -1,0 +1,44 @@
+### Map
+![Example how to get `content.left`](https://user-images.githubusercontent.com/23530004/181629414-5e40211a-7a3c-4d3c-8e78-8808b956d98b.png)
+_Example how to get `content.left`_
+
+### Blocks
+
+- browser window - edge from where `getBoundingClientRect` returns `.left`
+- viewport - container with overflow hidden. Content shouldn't be visible outside the viewport. Used in auto-placement, if overflows viewport - flip placement.
+- root - a container relative to which content will be positioned. We need to set content `.left` which is relative to `root.left`.
+- anchor - content must be directly near anchor, but then offset can be added.
+- content - block which we need to position.
+
+### Placement
+All placement names the same as in CSS.
+
+- verical - cross direction `left`, main:
+    - `bottom`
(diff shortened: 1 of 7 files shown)


packages/docs/src/page-configs/ui-elements/dropdown/specs.md
```diff
@@ -0,0 +1,44 @@
+### Map
+![Example how to get `content.left`](https://user-images.githubusercontent.com/23530004/181629414-5e40211a-7a3c-4d3c-8e78-8808b956d98b.png)
```
rustem-nasyrov, line 2: I didn't find a solution to import local assets. Any suggestions?

m0ksem: We'll need a markdownit plugin, that will handle relative paths. Let's have it on hold before massive docs migration to nuxt.

m0ksem (approved):

Issue #2224, opened 2022-08-12
Problems with Nuxt in 1.5.0 branch


aluarius:
For any component:

<img width="934" alt="image" src="https://user-images.githubusercontent.com/64714442/184320381-82ea8a20-c81c-489c-9182-b60ee186f95d.png">

aluarius:
https://github.com/epicmaxco/vuestic-ui/pull/2261

PR #2261, opened 2022-08-18
(1.5.0) Fix: stack overflow issue in Nuxt 3

aluarius:
Closes: #2224 

## Description
Subject fix.

## Types of changes

### DIFF
--- a/packages/ui/src/services/component-config/component-config.ts
+++ b/packages/ui/src/services/component-config/component-config.ts
@@ -1,7 +1,7 @@
 import { useLocalConfig } from '../../components/va-config/VaConfig'
 import { useGlobalConfig } from '../global-config/global-config'
-import { ComponentInternalInstance, computed } from 'vue'
-import type { VuesticComponentsMap } from '../../vuestic-plugin/global-components'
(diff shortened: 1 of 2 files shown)


packages/ui/src/services/component-config/component-config.ts
```diff
@@ -22,6 +22,9 @@ export const useComponentConfigProps = <T extends DefineComponent>(component: T,
   const localConfig = useLocalConfig()
   const { globalConfig } = useGlobalConfig()
 
+  const instancePreset = ref(instance.props?.preset)
+  watchEffect(() => (instancePreset.value = instance.props?.preset))
```
m0ksem, line 26:
```suggestion
  const instancePreset = computed(() => instance.props?.preset)
```

Isn't it the same?

m0ksem (changes requested):
Makes no sense for me completely. Need some comments in code.

m0ksem (approved):
@aluarius, I updated PR made it is less weird. 

Checked with nuxt if preset is reactive prop - all good. Result here: https://github.com/epicmaxco/vuestic-ui/issues/2269

Double check this and we are ready for merge.

Issue #2250, opened 2022-08-15
Dropdown position is broken after SPA route change in nuxt


m0ksem:
Reproduction: https://codesandbox.io/s/loving-hoover-phqo6c?file=/pages/index.vue

Luffyyy:
Been pretty annoying lately, hope it gets fixed soon enough

m0ksem:
> Been pretty annoying lately, hope it gets fixed soon enough

Sorry for that! It is going to be fixed ASAP, since we haven't released anything this week. For now, as a workaround, you can turn off `prevent-overflow`  prop for now.
```
  vuestic: {
    config: {
      components: {
        VaDropdown: {
          preventOverflow: true,
        }
      }
    },
  },
```

PR #2278, opened 2022-08-22
fix: floating dropdown position

rustem-nasyrov:
Close: #2250

## Description
Fix dropdown position is broken after SPA route change in nuxt apps.

## Types of changes


packages/ui/src/composables/useDropdown.ts
```diff
@@ -169,6 +151,25 @@ export const useDropdown = (
   }
 
   watchPostEffect(() => {
+    const rootRef = computed(() => {
```
m0ksem, line 154: Computed must be created in setup function, not in sub function. This cause creating new computed each time anchor or content DOM rect change.

m0ksem (changes requested):
I think the problem here is that root passed from VaDropdown is not reactive ref because of useTemplateRef.

packages/ui/src/composables/useDropdown.ts
rustem-nasyrov: What if replace it with a ref?

rustem-nasyrov:
External bug, reproduces only in the nuxt. Here is a link to the bug: https://github.com/nuxt/framework/issues/6845

m0ksem (approved):

PR #2281, opened 2022-08-22
Fix/unit tests

### LINKED ISSUE #2133: Fix unit tests
- [ ] Some tests went broken at some point because of either changed functionality or regression. You may want to handle that carefully (see `.skip`).
- [ ] Config is mocked in not ideal way. Too much duplication, we can think of a better solution.
```js
global: {
          provide: {
            [GLOBAL_CONFIG]: createGlobalConfig(),
          },
        },
```
- [ ] composables with provide that were tested without instance (as a function) are broken, we should probably use wrappers + setup to test these.

Tests were reenabled in context of https://github.com/epicmaxco/vuestic-ui/pull/2126.

Azgiliat:
fix #2133

## Description
Fix some unit tests

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-carousel/tests/VaCarousel.spec.ts
+++ b/packages/ui/src/components/va-carousel/tests/VaCarousel.spec.ts
@@ -6,17 +6,23 @@ import {
   GLOBAL_CONFIG,
   createGlobalConfig,
 } from '../../../services/global-config/global-config'
+import { mountWithGlobalConfig } from '../../../composables/tests/mountWithGlobalConfig'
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
   },
   "workspaces": {
     "packages": [
-      "packages/*"
+      "packages/*",
+      "packages/extensions/*"
```
asvae, line 37: extensions are not needed right now. (folder is removed)

asvae (approved):
All right!

Issue #2269, opened 2022-08-19
(1.5.0) Button text color doesn't change after preset changed


m0ksem:
![firefox_CQMsYpsgik](https://user-images.githubusercontent.com/23530004/185616257-ff1163f9-0708-4b06-a51b-454855300666.gif)

Repo code in nuxt:

```vue
<template>
  <va-button :preset="preset" @click="onClick">Aaa {{ preset }}</va-button>
</template>

<script setup lang="ts">
const preset = ref('text');

const onClick = () => {
  preset.value =  preset.value !== 'secondary' ? 'secondary' : 'text';
}
</script>
```

Problem can be related to presets, buttons or useTextColor.

aluarius:
Guess it's a presets service problem - it must switch preset to `default` if non-existing preset was passed ( it's just setting it and nevermind if it doesn't exist).

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2283

PR #2283, opened 2022-08-22
`VaPagination` rework

aluarius:
Additionally closes: #2269

## Description
- Pagination rework according to the new `VaButton` implementation.
- New visual states and examples.
- Component moved out from `VaButtonGroup`.
- Docs updates.

image

## Types of changes


packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -224,20 +216,45 @@ export default defineComponent({
...
+
+    const classComputed = useBem('va-pagination', () => ({
+      ...pick(props, ['gapped', 'rounded']),
+      bordered: !!props.borderColor,
+    }))
+
+    const currentPageButtonStyle = computed(() => ({
+      outline: props.currentPageBorderColor ? `1px solid ${getColor(props.currentPageBorderColor)}` : 'inherit',
+      outlineOffset: props.currentPageBorderColor ? '-1px' : 'inherit',
+    }))
```
RVitaly1978, line 239: In this case `currentPageButton` will not have any style defined unless the user passes `currentPageBorderColor`.
Maybe it's better to have an `active` state for `va-button`.

RVitaly1978 (changes requested):
Buttons have confusing styles. For example:
<details>
chrome-capture-2022-7-23
</details>

packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -104,16 +104,23 @@ export default defineComponent({
...
     directionIconRight: { type: String, default: 'chevron_right' },
+    gapped: { type: Boolean, default: false },
+    borderColor: { type: String, default: '' },
+    rounded: { type: Boolean, default: false },
+    currentPageBorderColor: { type: String, default: '' },
   },
 
   setup (props, { emit }) {
+    // TODO(1.6.0): Remove deprecated props
+    useDeprecatedProps(['flat', 'outline'])
```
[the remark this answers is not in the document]
aluarius: They are already removed. We have decided that we should throw warnings to help understand what to fix after update. We can't remove them during the other update because it's a breaking change.

packages/ui/src/components/va-pagination/VaPagination.vue
aluarius: I agree, but still don't know how to implement this - props to slotted buttons are passed via common button group props object - can't specify props for separate button.

aluarius:
> Buttons have confusing styles. 

Damn, looks strange, will investigate.

packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -304,12 +292,68 @@ export default defineComponent({
       opacity: 0.4;
     }
 
-    &--outline.va-button--disabled {
-      opacity: 1;
+    &.va-button--disabled {
+      opacity: 0.8;
```
rustem-nasyrov, line 296: Doesn't here will fit the `va-disabled` global mixin?

aluarius: It's already used in `VaButton` which is used by `VaButtonGroup` which is used by `VaButton` :).

But i just need to overwrite opacity, provided by mentioned mixin.

rustem-nasyrov:
Current page number is barely visible:
image
image

packages/ui/src/components/va-pagination/VaPagination.vue
aluarius: > Current page number is barely visible

Will move `VaButton` out of `VaButtonGroup` can't set separate button state without it. Moreover, no benefits, only problems.

aluarius:
> Current page number is barely visible

Will move VaButton out of VaButtonGroup can't set separate button state without it. Moreover, no benefits, only problems.

rustem-nasyrov (approved):

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
+        aria-hidden="true"
+        type="number"
       />
     </VbCard>
     <VbCard>
       <table class="table table-bordered">
         <tr>
           <th>Description</th>
-          <th>Va Pagination</th>
+          <th>VaPagination</th>
```
asvae, line 15: image

When you click on page - you get this orange border. Very annoying :).

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
         </tr>
         <tr>
-          <td>Default Pagination</td>
```
asvae, line 17: Pretty sure button should lose selection when it becomes active.

image

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
+          <td>Default</td>
```
asvae, line 18: image

default preset doesn't work too well as pagination default. primary/secondary make more sense for me.

For default it looks like 4 items are selected and one is disabled.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
               />
             </VbCard>
           </td>
         </tr>
         <tr>
-          <td>Pagination Sizes</td>
+          <td>Bordered</td>
           <td>
             <VbCard>
               <va-pagination
```
asvae, line 105: image

Left and right buttons stop working if modelValue is out of bounds (i.e. for pages="5", :modelValue="20")

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
         <tr>
-          <td>Colored Pagination</td>
+          <td>Colored</td>
           <td>
             <VbCard>
               <va-pagination
                 v-model="activePage"
-                :visible-pages="4"
                 :pages="15"
+                :visible-pages="4"
```
asvae, line 232: Maybe we can clean-up demos a bit. To remove unnecessary props. But up to you.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
+                v-model="activePage"
+                :pages="20"
+                :visible-pages="7"
+                boundary-numbers
+              />
+            </VbCard>
+          </td>
+        </tr>
+        <tr>
+          <td>Plain with input</td>
```
asvae, line 344: We probably want to preselect input value, so that it's faster to override.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
                 :pages="15"
+                :visible-pages="4"
                 color="warning"
               />
             </VbCard>
           </td>
         </tr>
         <tr>
-          <td>Disabled Pagination</td>
+          <td>Disabled</td>
```
asvae, line 251: image

Disabled pagination has no distinction from non disabled.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
           <td>
             <VbCard>
               <va-pagination
```
asvae, line 254: Except for cursor - disabled buttons have no distinction from active ones.

image

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
-      <va-config :components="{
+        <tr>
+          <td>Boundary Icons</td>
+          <td>
+            <VbCard>
+              <va-pagination
+                v-model="activePage"
+                :pages="20"
+                :visible-pages="5"
+                boundary-numbers
```
asvae, line 332: image

Only bottom demo moves boundary (`...`) on page selection. I assume both should move it.

Here also we might want an ability to remove navigation arrows, as they're somewhat redundant.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
+                direction-icon-right="volume_down"
+              />
+            </VbCard>
+          </td>
+        </tr>
+        <tr>
+          <td>Input</td>
+          <td>
+            <VbCard>
+              <va-pagination
```
asvae, line 316: Here would be nice to have an ability to remove navigation arrows, so that it looks like this

image

and this

image

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
-        :visible-pages="5"
-        input
-      />
-    </VbCard>
-    <VbCard
-      title="Hide on single page"
-    >
-      <va-pagination
-        :pages="1"
-        hide-on-single-page
```
asvae, line 198: This demo disappeared.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,216 +1,376 @@
...
+        <tr>
+          <td>Plain with input</td>
+          <td>
+            <VbCard>
+              <va-pagination
+                v-model="activePage"
+                :pages="20"
+                :visible-pages="5"
+                input
+                plain
```
asvae, line 352: Plain prop is at very weird spot right now. It works only with input (and it actually forces input), maybe we can remove it and somehow make visuals rely on preset prop?

Here's what it does for `default` pagination demo. Doesn't look intended.
image

packages/ui/src/composables/useHover.ts
```diff
@@ -1,13 +1,17 @@
...
 import { useHTMLElement, useEvent } from './'
 
-export function useHover (el?: Ref<HTMLElement | null | undefined>) {
+export function useHover (el?: Ref<HTMLElement | null | undefined>, disabled?: Ref<boolean>) {
   const isHovered = ref(false)
 
   const onMouseEnter = () => { isHovered.value = true }
   const onMouseLeave = () => { isHovered.value = false }
 
+  disabled && watch(disabled, (v) => {
```
asvae, line 11: Is this necessary? Not seeing the case where this is needed? (why not keep it hovered?)

packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -1,53 +1,52 @@
...
       >
         {{ n }}
       </va-button>
     </slot>
     <input
       v-else
+      v-model="inputValue"
       ref="htmlInput"
       class="va-pagination__input va-button"
       aria-label="enter the page number to go"
```
asvae, line 47: Hmm, this is necessary and we don't allow to pass translation?

packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -104,26 +109,31 @@ export default defineComponent({
...
 
   setup (props, { emit }) {
+    // TODO(1.6.0): Remove deprecated props
+    useDeprecatedProps(['flat', 'outline'])
+
     const htmlInput = shallowRef<HTMLInputElement>()
 
     const inputValue = ref('')
 
-    const usedTotal = computed(() => !!((props.total || props.pageSize === 0) && props.pageSize))
```
asvae, line 121: `usesTotal`?

asvae (changes requested):
Still some work to do.

Very happy with component quality wise, felt perfection while scrolling through code.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
aluarius: It's already solved. Just not in this issue branch. Will merge 1.5.0 to it and everything will be nice :)

packages/ui/src/components/va-pagination/VaPagination.demo.vue
aluarius: Agree. Will remove it from examples and demos.

packages/ui/src/composables/useHover.ts
aluarius: Without it we will get this:

https://user-images.githubusercontent.com/55198465/186181812-3604e6c6-a74a-46ce-a260-0b524a6e7152.gif

packages/ui/src/components/va-pagination/VaPagination.vue
aluarius: Will move `aria-attributes translations` to different issue - it's not related only to this component.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
aluarius: Active button is `disabled` so it can't be focused. I can implement 'personal' disabled state for `VaPagination` active button with blackjack and focus but honestly that doesn't make sense for me.

packages/ui/src/components/va-pagination/VaPagination.vue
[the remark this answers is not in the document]
m0ksem: @aluarius, deprecated usually means that it is something that still exists, but not recommended using and will be removed in feature updates.

I think we need just remove these props and write it in migration guide.

m0ksem (changes requested):
<details>
image
</details>

Buttons must be disabled.

---

Stange input

<details>
image
</details>

packages/ui/src/components/va-pagination/VaPagination.demo.vue
asvae: > Active button is disabled so it can't be focused

Why do we need that? Maybe let it be focusable?

Checked refs - everywhere buttons are focusable.

aluarius: Because current page button isn't interactive element. But i don't insist - will make it focusable.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
aluarius: It's seems to be old bug. Deal with it.

packages/ui/src/components/va-pagination/VaPagination.vue
```diff
@@ -224,20 +239,60 @@ export default defineComponent({
...
 
+    const buttonPropsComputed = computed(() => ({
+      size: props.size,
+      color: props.color,
+      borderColor: props.borderColor,
+      preset: props.preset || 'primary',
+    }))
+
+    const currentPageButtonProps = computed(() => ({
+      preset: props.preset === 'default' ? 'primary' : 'default',
```
asvae, line 250: That's not how presets are meant to be used. Let's discuss.

packages/ui/src/components/va-pagination/VaPagination.demo.vue
```diff
@@ -1,260 +1,428 @@
...
               />
             </VbCard>
           </td>
         </tr>
         <tr>
-          <td>Colored Pagination</td>
+          <td>Colored</td>
           <td>
             <VbCard>
               <va-pagination
```
asvae, line 229: I would assume colored would cover cursor color as well. The way it works for button etc.

image

asvae (approved):
Looks good!

Issue #2282, opened 2022-08-22
On hover - "Open in Github" button is barely visible


asvae:
All other buttons are ok.

<img width="277" alt="image" src="https://user-images.githubusercontent.com/5394573/185955026-336fbc25-3c0c-4d82-9594-d08ed0449c5b.png">

m0ksem:
> From which file is this exactly from ?

Hey, @Pathholder1806. This issue is related to DocsExample component. You can find reproduction on any demo we have in docs: https://vuestic.dev/en/ui-elements/alert#default

Pathholder1806:
The issue is persisting with all the 'a tags'

Pathholder1806:
So the issue is in the following file:
https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/styles/typography/_mixins.scss
Line number - 183

The value of brightness filter on hover is too high.

Pathholder1806:
@m0ksem 
**filter: brightness(100%)** should fix the issue.
File - https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/styles/typography/_mixins.scss
Line - 183

asvae:
Might be related to #2260. Global typography is giving us issues.

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2288

PR #2288, opened 2022-08-23
Updated Fixing issue: On hover - "Open in Github" button is barely vi…

Pathholder1806:
…sible #2282

So as in the description of the issue, the 'Open in GitHub' button becomes barely visible on hover.

Issue file location
File - https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/styles/typography/_mixins.scss
Line number - 183

Actual Issue: The actual issue was that the value of 'brightness' filter was way to much high(125%) which was making the button barely visible on hover

Fix: The value of **115%** for the 'brightness' filter should fix the issue and makes the 'Open in GitHub' button decently visible on hover.

**Note**
Hey @aluarius , this is a new PR with changes that you suggested in the brightness value from **103% -> 115%**
Kindly review the changes and I think we can go for merging it then.

## Types of changes

### DIFF
--- a/packages/ui/src/styles/typography/_mixins.scss
+++ b/packages/ui/src/styles/typography/_mixins.scss
@@ -180,6 +180,6 @@ $va-display-margins: (
 
   &:hover {
     color: var(--va-link-color);
-    filter: brightness(125%);
+    filter: brightness(115%);
   }
 }


aluarius:
Thank you!

Issue #2299, opened 2022-08-25
Codesandbox should use vite instead of vue-cli


asvae:
Vite is default environment, so we should use it. It's also faster.

Issue #2257, opened 2022-08-17
Codesandbox examples doesn't work


rustem-nasyrov:
1. Open any documentation of any component
2. Click any codesandbox example
3. After initizalisation facing issue given in the screenshot below.

![image](https://user-images.githubusercontent.com/29167241/185137847-05652e69-2b4e-4cef-bd40-8ae99730cf2d.png)

rustem-nasyrov:
@asvae can we close this?

m0ksem:
> @asvae can we close this?

Same as https://github.com/epicmaxco/vuestic-ui/issues/2313 now.

PR #2302, opened 2022-08-26
feat: migrate from vue-cli to the vite

rustem-nasyrov:
Close: #2299 #2257
image

## Description
Migration from the vue-cli solution to the vite for the codesandbox examples. 

## Types of changes

### DIFF
--- a/packages/docs/src/helpers/CodeSandboxExample.ts
+++ b/packages/docs/src/helpers/CodeSandboxExample.ts
@@ -7,16 +7,20 @@ import packageUi from 'vuestic-ui/package.json'
 const main = `import { createApp } from "vue"
 import App from "./App.vue"
 import { createVuestic } from "vuestic-ui"
-import 'vuestic-ui/css'
+import 'vuestic-ui/dist/vuestic-ui.css'
 
 const app = createApp(App)
 app.use(createVuestic())
 app.mount("#app")
 `
 
-const babel = `export const presets = [
-  '@vue/cli-plugin-babel/preset'
-];`
+const viteConfig = `import { defineConfig } from 'vite'
+import vue from '@vitejs/plugin-vue'
+
+export default defineConfig({
+  plugins: [vue()]
+})
+`
 
 const defaultExample = `<template>
   <div class="pa-4">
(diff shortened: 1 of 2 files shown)


packages/docs/src/helpers/CodeSandboxExample.ts
```diff
@@ -37,31 +41,30 @@ const getCodeSandboxHtml = ({ requireIcons = false }: CodesandboxConfig): string
...
-    'core-js': '^3.6.5',
     vue: '^3.0.0',
     'vuestic-ui': `${packageUi.version}`,
   }
   const commonDevDeps = {
-    '@vue/cli-plugin-babel': '~4.5.0',
-    '@vue/cli-service': '~4.5.0',
-    '@vue/compiler-sfc': '^3.0.0',
+    '@vitejs/plugin-vue': '^3.0.3',
+    vite: '^3.0.7',
```
rustem-nasyrov, line 61: @m0ksem, @asvae do I need to implement the auto picking the version from the `vuestic-ui/package.json`? Or better replace it with a `package-name@latest`?

packages/docs/src/helpers/CodeSandboxIconsHelper.ts
```diff
@@ -46,9 +45,7 @@ export const iconsConfig = `
   const app = createApp(App)
     .use(createVuestic({
       config: {
-        {
-          icons: createIconsConfig({ fonts }),
-        }
+        icons: createIconsConfig({ fonts }),
```
m0ksem, line 48: We use fontawesome in one single demo. I think we need to add a small alert to this demo that user needs to install fa font manually.

This is not related to this pr. All good here. Just noticed that here.

packages/docs/src/helpers/CodeSandboxExample.ts
m0ksem: Latest will be good. We will now that something is broken :)

m0ksem (approved):
We need to fix cjs build anyway, even if we move codesandbox to vite. Likely, vite uses cjs build in SSR.

packages/docs/src/helpers/CodeSandboxIconsHelper.ts
rustem-nasyrov: We already have a notification.
image

packages/docs/src/helpers/CodeSandboxExample.ts
```diff
@@ -37,31 +41,30 @@ const getCodeSandboxHtml = ({ requireIcons = false }: CodesandboxConfig): string
...
-    'core-js': '^3.6.5',
-    vue: '^3.0.0',
+    vue: 'latest',
     'vuestic-ui': `${packageUi.version}`,
   }
   const commonDevDeps = {
-    '@vue/cli-plugin-babel': '~4.5.0',
-    '@vue/cli-service': '~4.5.0',
-    '@vue/compiler-sfc': '^3.0.0',
+    '@vitejs/plugin-vue': 'latest',
```
asvae, line 60: Maybe cap to major version? `latest` seems too risky for me.

asvae (approved):
Looks good!

packages/docs/src/helpers/CodeSandboxExample.ts
m0ksem: ~3.0.0 I think would be nice. It will be a place where we can ensure that build doesn't work with last vite version.

packages/docs/src/helpers/CodeSandboxExample.ts
```diff
@@ -3,6 +3,10 @@ import { iconsStyles, iconsConfig } from './CodeSandboxIconsHelper'
 import { CodesandboxConfig } from '../types/configTypes'
 // @ts-ignore
 import packageUi from 'vuestic-ui/package.json'
+// @ts-ignore
+import exampleTsconfig from './codesandbox-example-config/example-tsconfig.json'
+// @ts-ignore
+import exampleTsConfigNode from './codesandbox-example-config/example-tsconfig.node.json'
```
rustem-nasyrov, line 9: About this `ts-ignore`, maybe better to declare the *.json modules or enable the `resolveJsonModule` flag in the `packages/docs` typescript configuration?

packages/docs/src/helpers/CodeSandboxExample.ts
```diff
@@ -37,31 +50,32 @@ const getCodeSandboxHtml = ({ requireIcons = false }: CodesandboxConfig): string
...
-    vue: '^3.0.0',
+    vue: 'latest',
     'vuestic-ui': `${packageUi.version}`,
   }
   const commonDevDeps = {
-    '@vue/cli-plugin-babel': '~4.5.0',
-    '@vue/cli-service': '~4.5.0',
-    '@vue/compiler-sfc': '^3.0.0',
+    '@vitejs/plugin-vue': '~3.0.0',
+    typescript: 'latest',
```
m0ksem, line 70: Can we have `~x.x.x` here as well? Let's stick to major version.

m0ksem (approved):

Issue #2303, opened 2022-08-26
Types. VaRadio `option` prop type is missing


m0ksem:
Must be string or boolean and likely required.

packages/ui/src/components/va-radio/VaRadio.vue

Udaay:
Also, what would be the default value for this ?

m0ksem:
> Also, what would be the default value for this ?

@Udaay, I think this mustn't have a default value. I can't see any case where you might need a default value... It always must be passed. So this is looks like required prop without default value.

Usage example:
```vue
<va-radio option="yes" />
<va-radio option="no" />
```

asvae:
Options is actually `any` :). You can pass whatever and `object`/`null` is pretty common case.

m0ksem:
> Options is actually `any` :). You can pass whatever and `object`/`null` is pretty common case.

Hm. Make sense.

Udaay:
I this PR related this is already merged , should i undo it ??

m0ksem:
> I this PR related this is already merged , should i undo it ??

No need to undo, would be nice to create new one. Sorry for that.

Udaay:
Yeah no problem I will do that

PR #2318, opened 2022-08-28
Fix: Types. VaRadio option prop type is missing [#2303]

Udaay:
## Description
Added a propType to option in vaRadio.vue

## Types of changes


m0ksem (approved):
Thanks!

PR #2322, opened 2022-08-29
feat: scroll-container

m0ksem:
image

Since vuestic is not exporting scss mixins for now, I just need this minor component.


packages/docs/src/page-configs/ui-elements/scroll-container/examples/Color.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <va-scroll-container vertical color="primary" style="max-height: 300px;">
+    <va-list>
+      <va-list-item v-for="i in 30" :key="i">
+        {{ i}}
```
rustem-nasyrov, line 5: We definitely need to enable linter for the docs examples too.

rustem-nasyrov (approved):

rustem-nasyrov:
It's not necessary, but I suggest you to add example where is horizontal and vertical scroll bars are enabled.

packages/docs/src/page-configs/ui-elements/scroll-container/examples/Color.vue
m0ksem: I just committed without linter cuz there was a lot of lint changes in docs, see: https://github.com/epicmaxco/vuestic-ui/pull/2321

Issue #2292, opened 2022-08-23
Virtual scroller component


aluarius:
**Live demo:**

- https://quasar.dev/vue-components/virtual-scroll#basic
- https://vuetifyjs.com/en/components/virtual-scroller/#usage

**References:**

- https://quasar.dev/vue-components/virtual-scroll
- https://vuetifyjs.com/en/components/virtual-scroller/
- https://github.com/Akryum/vue-virtual-scroller
- https://github.com/tangbc/vue-virtual-scroll-list
- https://github.com/zuolei828/vue-virtual-infinite-scroll

**Name:**

- `VaVirtualScroller`

**Slots:**

- `default` - slot for list item template, provides `item` and `index` variables for the scope

**Props:**

- `horizontal` <Boolean> - switches component to the horizontal mode
- `itemSize` <Number> - item average (`px` or `rem`) width or height (depends on `horizontal` property)
- `bench` <Number> - amount of items rendered before and after visible area (for smooth scroll)
- `items` <Array<any>> - array of data for scrollable items

**Example:**

```vue
<va-virtual-scroller :items="hugeArray" v-slot="{ item, index }">
  <div :key="index">{{ item }}</div>
</va-virtual-scroller>
```

aluarius:
Close via https://github.com/epicmaxco/vuestic-ui/pull/2331

PR #2331, opened 2022-08-31
[#2292] virtual scroller component

aluarius:
Closes: #2292 

## Description
Virtual scroller component implementation.

## Types of changes


packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
+import { PropType, ExtractPropTypes, ref, Ref, shallowRef, computed, watch, onMounted } from 'vue'
+
+import { useResizeObserver } from '../../composables'
+
+export const useVirtualScrollerSizesProps = {
+  horizontal: { type: Boolean, default: false },
+  itemSize: {
+    type: [Number, String] as PropType<string | number>,
+    default: 24,
```
rustem-nasyrov, line 9: I think here is default value is redundant, because it recalculates anyway.

rustem-nasyrov (changes requested):

packages/docs/src/locales/en/en.json
```diff
@@ -4293,5 +4309,35 @@
         "text": "`disabled` prop restricts to change panels size (including option via `maximization` prop)."
       }
     }
+  },
+  "virtualScroller": {
+    "title": "Virtual Scroller",
+    "summaryText": "`va-virtual-scroller` component allows you to use long lists without need to render every list child element but only ones visible depending on the current state of scroll.",
```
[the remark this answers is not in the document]
aluarius: Comma is not suitable before `but` in english.

packages/docs/src/locales/en/en.json
```diff
@@ -4293,5 +4309,35 @@
...
+    "examples": {
+      "default": {
+        "title": "Default",
+        "text": ""
+      },
+      "horizontal": {
+        "title": "Smooth scrolling",
+        "text": "`horizontal` prop allows you to change component's orientation."
```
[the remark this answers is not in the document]
aluarius: > How does the "Smooth scrolling" title connect to its description?

Yeah, misswrited.

P.S. Second `the` is not suitable.

packages/docs/src/page-configs/ui-elements/virtual-scroller/examples/Bench.vue
```diff
@@ -0,0 +1,20 @@
...
+  >
+    <va-badge class="pb-2" color="success" :text="`item ${item}`" />
+  </va-virtual-scroller>
+</template>
+
+<script>
+export default {
+  data: () => {
+    const hugeArrayBase = new Array(10000)
+    const hugeArray = hugeArrayBase.fill(null).map((_, index) => index + 1)
```
rustem-nasyrov, line 16: Redundant variable `hugeArrayBase` here and in the other examples.

rustem-nasyrov (changes requested):

rustem-nasyrov (approved):

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.demo.vue
```diff
@@ -0,0 +1,100 @@
...
+      >
+        <div :class="index % 2 ? 'pb-1' : 'pb-5'">
+          <va-button :size="index % 2 ? 'small' : 'large'">{{ item.value }}</va-button>
+        </div>
+      </va-virtual-scroller>
+    </VbCard>
+  </VbDemo>
+</template>
+
+<script>
```
asvae, line 78: We can use script setup for demos, as it's more convenient.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.demo.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <VbDemo class="va-virtual-scroller-demo">
+    <VbCard title="default">
+      <va-virtual-scroller
```
asvae, line 4: Default demo should have absolute minimum of props defined. From my perspective, that's `items` and `v-slot`

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.demo.vue
```diff
@@ -0,0 +1,100 @@
...
+      >
+        <div>{{ index }} - {{ item }}</div>
+      </va-virtual-scroller>
+    </VbCard>
+    <VbCard title="custom key">
+      <va-virtual-scroller
+        :items="hugeObjectsArray"
+        :bench="10"
+        :item-size="24"
+        custom-key="value"
```
asvae, line 42: That prop is renamed and shouldn't work?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.demo.vue
```diff
@@ -0,0 +1,100 @@
...
+  data: () => {
+    const hugeArrayBase = new Array(10000)
+    const hugeArray = hugeArrayBase.fill(null).map((_, index) => index)
+    const hugeObjectsArray = hugeArrayBase.fill(null).map((el, index) => ({ value: index }))
+    return { hugeArray, hugeObjectsArray }
+  },
+}
+</script>
+
+<style lang="scss">
```
asvae, line 98: generally styles are unneeded in demos, so we can remove. (I'm nitpicking here 😅)

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
+<template>
+  <div ref="wrapper" class="va-virtual-scroller">
```
asvae, line 2: Do we truly need 3 wrappers?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+  props: {
+    ...useVirtualScrollerSizesProps,
+    items: { type: Array, required: true, validator: (v: any[]) => !!v.length },
+    bench: { type: Number, default: 10, validator: (v: number) => v >= 0 },
+    trackBy: { type: String, default: '' },
+    disabled: { type: Boolean, default: false },
+  },
+
+  setup: (props) => {
+    const currentListScroll = ref<number>(0)
```
asvae, line 44: `<number>` does nothing, TS can figure the type from `0`. (also nitpicking :>)

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+import { useVirtualScrollerSizes, useVirtualScrollerSizesProps } from './useVirtualScrollerSizes'
+
+import { warn } from '../../services/utils'
+
+export default defineComponent({
+  name: 'VaVirtualScroller',
+
+  props: {
+    ...useVirtualScrollerSizesProps,
+    items: { type: Array, required: true, validator: (v: any[]) => !!v.length },
```
asvae, line 37: Items could definitely be an empty array.

Maybe we can also integrate some empty list handling, but up to you.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
...
+
+import { useResizeObserver } from '../../composables'
+
+export const useVirtualScrollerSizesProps = {
+  horizontal: { type: Boolean, default: false },
+  itemSize: {
+    type: [Number, String] as PropType<string | number>,
+    default: 0,
+    validator: (v: number | string) => {
+      if (typeof v === 'string') { return (!isNaN(+v) || v.endsWith('px') || v.endsWith('rem')) && parseInt(v) >= 0 }
```
asvae, line 11: Maybe we can provide more descriptive error messages via `warn` function we have in utils?

validator won't tell much of what's wrong

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+    }
+    if (!props.disabled) { useEvent('scroll', handleScroll, true) }
+
+    const { list, wrapper, wrapperSize, itemSize } = useVirtualScrollerSizes(props, currentListScroll)
+
+    const uniqueKey = (item: unknown, index: number) => {
+      if (props.trackBy && item && typeof item === 'object' && !isArray(item)) {
+        const key = (item as Record<string, unknown>)[props.trackBy]
+
+        if (key || key === 0) { return `virtual-scroller-item-${key}` }
```
asvae, line 57: Probably doesn't matter, but maybe shorter key? like `va-${key}`. Btw, why did you even use string for that? key could be an object.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+import { warn } from '../../services/utils'
+
+export default defineComponent({
+  name: 'VaVirtualScroller',
+
+  props: {
+    ...useVirtualScrollerSizesProps,
+    items: { type: Array, required: true, validator: (v: any[]) => !!v.length },
+    bench: { type: Number, default: 10, validator: (v: number) => v >= 0 },
+    trackBy: { type: String, default: '' },
```
asvae, line 39: Can't we reuse `trackBy` from `useSelectableList` composable?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+
+import { warn } from '../../services/utils'
+
+export default defineComponent({
+  name: 'VaVirtualScroller',
+
+  props: {
+    ...useVirtualScrollerSizesProps,
+    items: { type: Array, required: true, validator: (v: any[]) => !!v.length },
+    bench: { type: Number, default: 10, validator: (v: number) => v >= 0 },
```
asvae, line 38: offset maybe?

Or did you find this `bench` word somewhere?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+    trackBy: { type: String, default: '' },
+    disabled: { type: Boolean, default: false },
```
asvae, line 40: Not seeing a demo with that prop :)

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
+import { PropType, ExtractPropTypes, ref, Ref, shallowRef, computed, watch, onMounted } from 'vue'
+
+import { useResizeObserver } from '../../composables'
+
+export const useVirtualScrollerSizesProps = {
+  horizontal: { type: Boolean, default: false },
+  itemSize: {
+    type: [Number, String] as PropType<string | number>,
+    default: 0,
```
asvae, line 9: Does 0 even work?

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
...
+    validator: (v: number | string) => {
+      if (typeof v === 'string') { return (!isNaN(+v) || v.endsWith('px') || v.endsWith('rem')) && parseInt(v) >= 0 }
+      return v >= 0
+    },
+  },
+}
+
+export const useVirtualScrollerSizes = (
+  props: ExtractPropTypes<typeof useVirtualScrollerSizesProps>,
+  currentScroll: Ref<number>,
```
asvae, line 19: `scrollPosition`?

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
...
+) => {
+  const wrapper = shallowRef<HTMLElement>()
+  const list = shallowRef<HTMLElement>()
+
+  const itemSizeCalculated = ref(0)
+  const wrapperSize = ref(0)
+  const bodyFontSize = ref(16)
```
asvae, line 26: ability to set sizes in refs and px should be represented in demo, also stated in docs.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
asvae, line 11: Would hope to see this logic being moved elsewhere (to utils maybe), but up to you.
Wondering if that's a concept we can reuse for other components.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
...
+  const itemSize = computed(() => {
+    if (typeof props.itemSize === 'string') {
+      const sizeParsed = parseInt(props.itemSize)
+      return props.itemSize.endsWith('rem') ? sizeParsed * bodyFontSize.value : sizeParsed
+    }
+
+    const result = Math.max(props.itemSize, itemSizeCalculated.value, 1)
+    const diff = Math.abs(((oldItemSize / result) * 100) - 100)
+
+    if (diff > 5 || oldItemSize === 0) {
```
asvae, line 69: Magic number? :D
Can you explain?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+
+    const { list, wrapper, wrapperSize, itemSize } = useVirtualScrollerSizes(props, currentListScroll)
+
+    const uniqueKey = (item: unknown, index: number) => {
+      if (props.trackBy && item && typeof item === 'object' && !isArray(item)) {
+        const key = (item as Record<string, unknown>)[props.trackBy]
+
+        if (key || key === 0) { return `virtual-scroller-item-${key}` }
+
+        warn(`[va-virtual-scroller] Key ${props.trackBy} wasn't found in ${JSON.stringify(item)}!`)
```
asvae, line 59: Wouldn't recommend stringifying item, things can go wrong in a lot of ways. (like if you have recursive links)

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,126 @@
...
+    const handleScroll = () => {
+      if (!wrapper.value) { return }
+      currentListScroll.value = wrapper.value[props.horizontal ? 'scrollLeft' : 'scrollTop']
+    }
+    if (!props.disabled) { useEvent('scroll', handleScroll, true) }
+
+    const { list, wrapper, wrapperSize, itemSize } = useVirtualScrollerSizes(props, currentListScroll)
+
+    const uniqueKey = (item: unknown, index: number) => {
+      if (props.trackBy && item && typeof item === 'object' && !isArray(item)) {
```
asvae, line 54: In this case `trackBy` should be more universal, item could easily be an array.

Maybe we can reuse the logic from selectable composable?

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,78 @@
...
+    calcAverageItemsSize()
+    handleWrapperResize()
+  })
+  watch(currentScroll, calcAverageItemsSize)
+
+  const handleWrapperResize = () => {
+    if (!wrapper.value) { return }
+
+    wrapperSize.value = props.horizontal ? wrapper.value.offsetWidth : wrapper.value.offsetHeight
+    bodyFontSize.value = parseFloat(getComputedStyle(document.documentElement).fontSize)
```
asvae, line 54: That's not `body`, that's `html` element.

packages/docs/src/pages/ui-elements/virtual-scroller.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <docs-content :config="configComputed" />
+</template>
+
+<script lang="ts">
+import { Vue, Options } from 'vue-class-component'
```
asvae, line 6: Let's not use class component (for new stuff at least)

I added and issue to remove it for good - #2345

packages/docs/src/page-configs/ui-elements/virtual-scroller/examples/DifferentContent.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <va-virtual-scroller
+    :items="hugeArray"
+    style="height: 400px;"
```
asvae, line 4: That doesn't look nice, if hight/width is mandatory - we should make it to a prop for convenience.

asvae:
We're pretty close!

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: Discussed in Slack.

Yes, we need (but i'm also not happy about this).

First div - wrapper, visible list size. Second - virtual container with size as if all items are rendered. Third - 'sledge', riding on second container in visible zone of the first.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: Will discuss tomorrow - can't imagine the purpose of this case.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: It's not logically suitable for this component. Also `useSelectableList` has not exactly the same what i need here.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: In vuetify docs :) And i like it because of its meaning. Offset is also suitable but worse (in my opinion) because it's not actually offset.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
aluarius: We calculate items size ourselves, so, this prop is optional.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
asvae: > Key should be unique. There is may be another `va-1` key on page.

Actually it should be unique only for current list, not on page. I don't see a way for you can have same items by key in list. I actually even don't see why prefix is needed at all.

> object key

Checked docs - https://vuejs.org/api/built-in-special-attributes.html.
Seems you're right, can't be object :D (only string, number or symbol). Still don't see a need to convert or validate it - vue can do that instead.

I used objects as keys before, but that is kinda undocumented feature. :)

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: Answered above.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
aluarius: I can reuse it in split component.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
aluarius: Will add comment into this part of code.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
asvae: Let's not use it in demos and examples then.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
asvae: @aluarius hmm, maybe we can move trackBy to different composable or something?

trackBy logic fits our case very nicely, no?

asvae (approved):
👍

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
```diff
@@ -0,0 +1,145 @@
...
+      overflow: hidden;
+    }
+
+    &__list {
+      display: flex;
+      flex-direction: column;
+    }
+
+    &--horizontal {
+      &.va-virtual-scroller__list {
```
rustem-nasyrov, line 140: Isn't it will be `va-virtual-scroller--horizontal va-virtual-scroller__list`?

packages/ui/src/services/utils.ts
```diff
@@ -99,3 +99,14 @@ const getRandomString = (stringLength = 4): string => {
 export const generateUniqueId = () => {
   return `${getRandomString(8)}-${getRandomString(4)}-${getRandomString(4)}`
 }
+
+export const isParsablePositiveMeasure = (value: unknown) => {
```
rustem-nasyrov, line 103: Will it be often case, where we need a positive measure? In my opinion will be nice to have something like `isParsableMeasure` and its dependant `isPositiveMeasure`.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
```diff
@@ -0,0 +1,97 @@
...
+export const useVirtualScrollerSizes = (
+  props: ExtractPropTypes<typeof useVirtualScrollerSizesProps>,
+  scrollPosition: Ref<number>,
+) => {
+  const list = shallowRef<HTMLElement>()
+  const wrapper = shallowRef<HTMLElement>()
+
+  const parseSizeValue = (value: number | string) => {
+    if (typeof value === 'string') {
+      const parsedValue = parseInt(value)
```
rustem-nasyrov, line 38: What would be if we get here `NaN` value? Is it possible?

rustem-nasyrov (changes requested):

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: And what's the problem? BEM doesn't welcome structures with multiply elements or modifiers.

packages/ui/src/components/va-virtual-scroller/useVirtualScrollerSizes.ts
aluarius: It's possible only if `value` contains no digits at all. This situation is regulated via props validator but i'll add one more check here just in case.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.vue
aluarius: Yes it's worked because i have forgotten to move root modifier class to the root div :)

PR #2340, opened 2022-09-01
fix:vaRadio option propType and added default null #2303

Udaay:
## Description

- Added PropType to options in vaRadio, now it has 4 types String, Boolean, Object and Number and default value will be null.
- Added type Number to modelValue, as passing any number dataType in options , causing a warning of type mismatch.

## Types of changes


Udaay:
@m0ksem @asvae  have a look

m0ksem (approved):

Issue #2092, opened 2022-07-14
Error examples added (with an icon for better accessibility, and without icon


m0ksem:

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2346

PR #2346, opened 2022-09-05
[#715] `VaCheckbox` visual updates

aluarius:
Closes: #715 
Closes: #2092 

## Description
`VaCheckbox` visual updates. Agreed with Vasiliy.

image

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/components/VaMessageList/VaMessageList.vue
+++ b/packages/ui/src/components/va-input/components/VaMessageList/VaMessageList.vue
@@ -10,7 +10,7 @@
       class="va-message-list__message"
     >
       <va-icon
-        v-if="isError"
+        v-if="hasError"
         class="va-message-list__icon"
         name="warning"
         :size="16"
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/va-input/components/VaMessageList/VaMessageList.vue
```diff
@@ -9,25 +9,36 @@
...
+  components: { VaIcon },
+
   props: {
     modelValue: {
       type: [String, Array] as PropType<string | string[]>,
       default: '',
     },
     limit: { type: Number, default: 1 },
     color: { type: String },
+    isError: { type: Boolean, default: false },
```
rustem-nasyrov, line 41: Why it's named as `isError` instead of `hasError` or something similar?

aluarius: It's state naming and seems naturally to me. But `hasError` is also ok. Will rename.

Azgiliat (approved):

PR #2364, opened 2022-09-06
Add missing types for array-value prop

skyroses:
## Description

There is no type for the array-value property in the checkbox component

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useSelectable.ts
+++ b/packages/ui/src/composables/useSelectable.ts
@@ -27,7 +27,7 @@ export const useSelectableProps = {
   ...useStatefulProps,
   ...useLoadingProps,
   ...useValidationProps,
-  arrayValue: { type: [String, Boolean, Object, Number] as PropType<any>, default: null },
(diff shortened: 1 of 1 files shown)


packages/ui/src/composables/useSelectable.ts
```diff
@@ -27,7 +27,7 @@ export const useSelectableProps = {
   ...useStatefulProps,
   ...useLoadingProps,
   ...useValidationProps,
-  arrayValue: { type: null as any as PropType<unknown>, default: null },
+  arrayValue: { type: [String, Boolean, Object, Number] as PropType<any>, default: null },
```
m0ksem, line 30:
```suggestion
  arrayValue: { type: [String, Boolean, Object, Number], default: undefined },
```

Vue can extract this as `PropType<string, boolean, Record<string, any>, number>` automatically.

m0ksem (changes requested):

m0ksem (approved):

PR #2374, opened 2022-09-08
`VaButton` fixes

aluarius:
## Description
- [x] Fixed bugs with `VaButtons` after merging dark theme functionality.
- [x] Fixed minor bug for `VaToggle` (different buttons height for icon-only and not icon-only states in one toggle group block).

## Types of changes

### DIFF
--- a/packages/docs/src/components/header/Header.vue
+++ b/packages/docs/src/components/header/Header.vue
@@ -22,13 +22,12 @@
       <algolia-search />
     </div>
     <div class="header__nav">
-
-      <div class="header__prefences sm-hidden flex">
+      <div class="header__prefences sm-hidden">
         <va-button
           v-for="(link, index) in links"
           :key="index"
           preset="secondary"
-          class="mr-1"
+          class="header__prefences__button"
           :to="link.to"
           :href="link.url"
           :target="link.target"
@@ -70,6 +69,7 @@ const PropsMixin = Vue.with(Props)
     AlgoliaSearch,
   },
 })
+
 export default class Header extends mixins(PropsMixin) {
   get locale () {
     return this.$root?.$i18n?.locale
@@ -109,11 +109,11 @@ export default class Header extends mixins(PropsMixin) {
   .header {
     width: 100%;
     height: 4rem;
-    box-shadow: 0 2px 8px var(--va-shadow);
+    padding: 0 calc(1.2rem + 4px);
     display: flex;
     justify-content: space-between;
     align-items: center;
-    padding: 0 calc(1.2rem + 4px);
+    box-shadow: 0 2px 8px var(--va-shadow);
 
     @include media-breakpoint-down(sm) {
       .sm-hidden {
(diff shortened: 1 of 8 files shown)


rustem-nasyrov:
Facing issues with colors in the sources.
image

rustem-nasyrov:
Blue buttons have a black text. 
image

rustem-nasyrov:
`Open in GitHub` button lead to the 404 page.
image

rustem-nasyrov (changes requested):
Needs to be compared with a current version on vuestic.dev and fix the differences.

m0ksem:
> `Open in GitHub` button lead to the 404 page. image

It is okay, since it is not on GitHub yet, but in 1.5.0 branch :)

packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -170,7 +170,7 @@ export default defineComponent({
     }))
 
     // styles
-    const isLowContrastBg = computed(() => props.plain || isLightBackground(colorComputed.value, props.backgroundOpacity))
+    const isLowContrastBg = computed(() => props.plain || isLightBackground(colorComputed.value, props.backgroundOpacity, 200))
```
m0ksem, line 173:
```suggestion
    const isTransparentBg = computed(() => props.plain || props.backgroundOpacity < 0.5)
```

packages/ui/src/services/color-config/color-functions.ts
```diff
@@ -6,10 +6,13 @@ export const colorToRgba = (color: ColorInput, opacity: number) => {
   return new ColorTranslator(color).setA(opacity).RGBA
 }
 
-export const isLightBackground = (color: ColorInput, opacity = 1) => {
-  // TODO: replace with color mixin (doesn't play any role in dark theme)
+export const isLightBackground = (color: ColorInput, opacity = 0.6, edge = 120) => {
```
m0ksem, line 9: Do not use opacity here at all. Better use applyColors(actuallBg, bg)

Opacity 0.6 doesn't mean that is it light bg...

packages/docs/src/components/header/Header.vue
```diff
@@ -127,48 +127,18 @@ export default class Header extends mixins(PropsMixin) {
...
-          justify-content: center;
-        }
-      }
     }
 
-    &__links {
+    &__prefences {
       display: flex;
+      align-items: center;
+      font-weight: $font-weight-bold;
```
m0ksem, line 135: Do not scss variables. Use css variables or hardocde :)

m0ksem (changes requested):

m0ksem:
> Blue buttons have a black text. image

I'll handle it.

aluarius:
> Facing issues with colors in the sources. 

Not related to this pr, will move to different issue.

aluarius:
> Needs to be compared with a current version on vuestic.dev and fix the differences.

It's impossible because current version contains old buttons implementation.

packages/ui/src/components/va-button/VaButton.vue
aluarius: I planned to solve this while reimplementing buttons to after-before. But ok, let's make it here :)

packages/docs/src/components/header/Header.vue
aluarius: It was here before - will remove.

m0ksem (approved):

PR #2377, opened 2022-09-08
Feat/kitchensink

### LINKED ISSUE #1168: Kitchensink page
https://vuesax.com/
https://get.foundation/sites/docs/kitchen-sink.html
https://framework7.io/vue/kitchen-sink.html

What purpose does it serve:
* Component test 5/5
* Framework preview for user 3/5

- [ ] You can also check if we need something in kitchensink apart from components.
- [ ] We would also need some feedbpack on state of Default examples.

Azgiliat:
## Description

Add kitchensink page

Close #1168

## Types of changes


packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
```diff
@@ -0,0 +1,36 @@
...
+  /page-config.ts$/,
+  'lazy',
+)
+
+export const config = shallowReactive<ApiDocsBlock[]>([])
+
+importedConfigsContext.keys().forEach(async (filename) => {
+  let importedConfig = null
+  try {
+    importedConfig = (await importedConfigsContext(filename)).default
```
aluarius, line 16:
```suggestion
    importedConfig = (await importedConfigsContext(filename))?.default
```
If here is no way it will return null or undefined, ignore this. Or you rely on try-catch?

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
```diff
@@ -0,0 +1,36 @@
...
+  } catch {
+    return
+  }
+
+  config.push({ ...importedConfig[0], type: BlockType.SUBTITLE }) // title
```
aluarius, line 21: I'd prefer here array-length check but up to you.

aluarius (approved):

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -59,6 +59,10 @@ export const navigationRoutes: NavigationRoute[] = [
         name: 'ssr-guide',
         displayName: 'menu.ssrGuide',
       },
+      {
```
asvae, line 62: I would suggest building this in context of 1.5.0 branch.

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -59,6 +59,10 @@ export const navigationRoutes: NavigationRoute[] = [
...
+        name: 'kitchensink',
+        displayName: 'menu.kitchensink',
```
asvae, line 64: Affix and backtop are pretty horrible, can we do something about them?

image

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -59,6 +59,10 @@ export const navigationRoutes: NavigationRoute[] = [
...
+      },
       // GENERATOR_ADD - gettingStarted
```
asvae, line 66: image

I don't see a case where you'd need code example for kitchensink, at least we can link docs (but that's also dubiuous)

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
```diff
@@ -0,0 +1,36 @@
...
+  true,
+  /page-config.ts$/,
+  'lazy',
+)
+
+export const config = shallowReactive<ApiDocsBlock[]>([])
+
+importedConfigsContext.keys().forEach(async (filename) => {
+  let importedConfig = null
+  try {
```
asvae, line 15: Why do we need try/catch?

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
asvae, line 21: > importedConfig[0]

Maybe you can instead find title, it's not immediately clear what you're trying to do.

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  config.push({ ...importedConfig[0], type: BlockType.SUBTITLE }) // title
+
+  const example =
+    importedConfig.find(
+      (blockElement: any) =>
+        blockElement.type === BlockType.EXAMPLE &&
+        blockElement.component === 'Default',
+    ) ||
+    importedConfig.find(
```
asvae, line 29: What's the point of this rollback?

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
```diff
@@ -0,0 +1,36 @@
...
+      (blockElement: any) => blockElement.type === BlockType.EXAMPLE,
+    )
+
+  if (example) {
```
asvae, line 33: That logic looks confusing for me. Why would you have 2 checks?

I.e you can have title, but no example? Then for user it still would be broken.

packages/docs/src/pages/getting-started/kitchensink.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <div>
+    <docs-content :config="config" />
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { config } from '@/page-configs/getting-started/kitchensink/page-config'
```
asvae, line 8: Needs relative paths :X.

asvae (changes requested):
Feels pretty good.

But there are still some things to fix. Also we need to do a follow up on improving examples (could you create issue on that). These are too low quality.

m0ksem (changes requested):
image

Let's have one input here.

---

image
Let's show this icons in one line

---

Affix target must be changed to scroll container.

---

I would like to have a link to component to see another examples.

packages/docs/src/components/sidebar/navigationRoutes.ts
Azgiliat: We can create good examples :)

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
Azgiliat: We want to grab _only_ default examples? And if there is no default example don't show component at all? So then we need to create default examples for all our components (right now some components don't have it)

packages/docs/src/components/sidebar/navigationRoutes.ts
asvae: @Azgiliat I mean maybe replacing default example (pretty sure both backtop and affix have containerized versions).

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
asvae: > right now some components don't have it

Let's figure out which components don't have it and think about it. :)

Why would we have components without default example?

asvae:
@m0ksem we might want to merge it without modifying demos, then follow it up with demo fixes as a separate PR.

packages/docs/src/components/sidebar/navigationRoutes.ts
Azgiliat: Backtop excluded, affix is placed in container now

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
Azgiliat: Yep, looking for title block now

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
Azgiliat: Only default examples now, no exmaple === don't show

packages/docs/src/page-configs/getting-started/kitchensink/page-config.ts
Azgiliat: fixed

packages/docs/src/pages/getting-started/kitchensink.vue
Azgiliat: Maybe we can use absolute path here? Here I need to go upstairs to `src` folder and then downstairs. Why can't we just start immediatly from `@ === src` ?

packages/docs/src/components/sidebar/navigationRoutes.ts
Azgiliat: removed

aluarius (approved):

asvae (approved):
Looks almost finished. The only thing that's missing from my perspective - is an indication of what the page is supposed to do.

When user lands on page - they might have no idea what "kitchensink" means, so we have to explain.

PR #2386, opened 2022-09-11
Import ScrollContainer component and fix plugin path

skyroses:
## Description
Added a forgotten import of the ScrollContainer component and fix path to the plugin 

## Types of changes

### DIFF
--- a/packages/nuxt/src/composables/use-plugin.ts
+++ b/packages/nuxt/src/composables/use-plugin.ts
@@ -8,7 +8,7 @@ import type { VuesticOptions } from './../types';
 /** Register vuestic nuxt plugin */
 export const useVuesticPlugin = (options: VuesticOptions) => {
   addPluginTemplate({
-    src: resolve(distDir, './runtime/plugin.mjs'),
+    src: resolve(distDir, './runtime/plugin.ts'),
     filename: 'plugin.mjs',
 
     // Use JSON.stringify() here, because it will be inserted in ejs template as string. Then we will JSON.parse it.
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/composables/use-plugin.ts
```diff
@@ -8,7 +8,7 @@ import type { VuesticOptions } from './../types';
 /** Register vuestic nuxt plugin */
 export const useVuesticPlugin = (options: VuesticOptions) => {
   addPluginTemplate({
-    src: resolve(distDir, './runtime/plugin.mjs'),
+    src: resolve(distDir, './runtime/plugin.ts'),
```
m0ksem, line 11: Must stay `.mjs`, cuz `ts` file is compiled to `mjs`.

Were there any issues with that before?

m0ksem (changes requested):

packages/nuxt/src/composables/use-plugin.ts
skyroses: Without this, the module is not compiled

aluarius: It's compiled and with `.ts` here will be a wrong path (dist module.mjs on the screen):

image

aluarius (changes requested):

packages/nuxt/src/composables/use-plugin.ts
skyroses: I didn't change the filename. 
With src dir (`resolve(distDir, "./runtime/plugin.mjs")`) we can't get types using `dev:prepare`. We get an error that there is no such file. With an extension .ts everything is fine

skyroses: And the `filename` option specifies how the file will be named after compilation - there should be `mjs`, yes

m0ksem:
Moved to https://github.com/epicmaxco/vuestic-ui/pull/2610

PR #2389, opened 2022-09-12
[#2295] `VaDataTable` scroll events

### LINKED ISSUE #2295: Emit events in `va-data-table` when the scroll exceeds the top or bottom

RVitaly1978:
## Description
close #2295

changes:
- [x] add props `scrollTopMargin` and `scrollBottomMargin` (type `number`)
- [x] add `scroll:bottom` and `scroll:top` emits
- [x] update docs example

chrome-capture-2022-8-12

## Types of changes


packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,35 @@
+interface useTableScrollProps {
```
aluarius, line 1: Maybe will be better to create exported object `useTableScrollProps`, then `props: ExtractPropTypes<typeof useTableScrollProps>` and  import-destruct it in `VaDataTable` props.

aluarius (approved):

rustem-nasyrov (approved):

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,35 @@
...
+      return
+    }
+
+    if (props.scrollEventsThreshold === undefined) {
+      return
+    }
+
+    const { scrollTop, scrollHeight, clientHeight } = evt.target as HTMLTableElement
+
+    if (scrollHeight - clientHeight - scrollTop <= props.scrollEventsThreshold) {
```
m0ksem, line 23: I think `scrollEventsThreshold` should be renamed to "margin"... Or not?
If this made for lazy-loading I suggest following [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -329,6 +333,7 @@ export default defineComponent({
     height: { type: [String, Number] },
     rowBind: { type: null as unknown as PropType<DataTableRowBind> },
     cellBind: { type: null as unknown as PropType<DataTableCellBind> },
+    scrollEventsThreshold: { type: Number },
```
m0ksem, line 336:
```suggestion
    scrollEventsThreshold: { type: Number, default: 0 },
```

m0ksem (approved):

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
RVitaly1978: reworked with [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

packages/ui/src/components/va-data-table/VaDataTable.vue
RVitaly1978: The main idea is to not fire scroll events if set to `undefined`.

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,101 @@
...
+    target && observer.unobserve(target)
+  }
+
+  const initObserver = () => {
+    observer = new IntersectionObserver(intersectionHandler, {
+      root: scrollContainerHTML.value,
+      rootMargin: margins.value,
+    })
+
+    props.scrollTopMargin !== undefined && observeTarget(topTriggerHTML.value)
```
m0ksem, line 66: Do we need to init observer while these two are undefined?

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  watch(() => props.scrollBottomMargin, (newValue) => {
+    if (isObservable.value) {
+      newValue !== undefined
+        ? nextTick(() => observeTarget(bottomTriggerHTML.value))
+        : unObserveTarget(bottomTriggerHTML.value)
+    }
+  })
+
+  onMounted(() => isObservable.value && initObserver())
```
m0ksem, line 92: This can be handled in separated composable useIntersectionObserver. Same as we have with ResizeObserver.

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,101 @@
...
+  const scrollContainer = shallowRef<HTMLDivElement>()
+  const scrollContainerHTML = useHTMLElement(scrollContainer)
+
+  const topTrigger = shallowRef<HTMLDivElement>()
+  const topTriggerHTML = useHTMLElement(topTrigger)
+
+  const bottomTrigger = shallowRef<HTMLDivElement>()
+  const bottomTriggerHTML = useHTMLElement(bottomTrigger)
+
+  const isScrollable = computed(() => !!(props.height))
```
m0ksem, line 28: Can't table be scrollable without `height` prop? I mean, smby can set table height with CSS. At least, I've made it.

m0ksem (commented):
Why do we need to make it with IntersectionObserver, not just onscroll? We stick to margin props anyway. 

Also, we have infinite scroll component. If you're not able to use it in table, this means that infinite scroll component not quite good.

packages/ui/src/components/va-data-table/VaDataTable.vue
m0ksem: Why not? It feels unnatural syncing threshold value (number) with doWeNeedToFireScrollEvent (boolean).

RVitaly1978: done

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,65 @@
...
+  scrollBottomMargin: { type: Number, default: 0 },
+}
+
+export const useTableScrollEmits = ['scroll:top', 'scroll:bottom']
+
+export default function useTableScroll (
+  props: ExtractPropTypes<typeof useTableScrollProps>,
+  emit: (event: 'scroll:bottom' | 'scroll:top', ...args: any[]) => void,
+) {
+  // TODO: replace 'vNodeProps' with 'attrs' when we bump vue to ^3.2
```
m0ksem, line 16: Will it work with 3.2 now?

packages/ui/src/composables/useIntersectionObserver.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const initObserver = () => {
+    observer.value = new IntersectionObserver(cb, unref(options))
+  }
+
+  let unwatchOptions: WatchStopHandle
+
+  const watchOptions = () => {
+    unwatchOptions = watch(options, () => {
```
m0ksem, line 33: Do not use watch inside method, this creates memory leak.

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,65 @@
...
+  const vNodeProps = getCurrentInstance()?.vnode.props
+  const isTopTriggerListener = vNodeProps?.['onScroll:top'] !== undefined
+  const isBottomTriggerListener = vNodeProps?.['onScroll:bottom'] !== undefined
+
+  const scrollContainer = useElementRef()
+  const topTrigger = useElementRef()
+  const bottomTrigger = useElementRef()
+
+  const isObservable = computed(
+    () => (isTopTriggerListener || isBottomTriggerListener) && !!scrollContainer.value,
```
m0ksem, line 27: I think it's okay to fire an event if there is no listener. It looks like you don't need this...

Firing event even might be more performant then having reactive variable :) And code much easier to read and understand.

m0ksem (changes requested):

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
RVitaly1978: The vue 3 [docs](https://v3-migration.vuejs.org/breaking-changes/listeners-removed.html) say listeners are available in `attrs`, but they are still not available in our current version of vue.
We have decided to use `getCurrentInstance()?.vnode.props` now and track listener availability in future vue versions.

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,63 @@
...
+    rootMargin: `${props.scrollTopMargin ?? 0}px 0px ${props.scrollBottomMargin ?? 0}px 0px`,
+  }))
+
+  useIntersectionObserver(intersectionHandler, options, targets)
+
+  return {
+    scrollContainer,
+    topTrigger,
+    bottomTrigger,
+    isTopTriggerListener,
```
m0ksem, line 60: Not important: `doShowTopTrigger` or `doRenderTopTrigger`

packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,63 @@
...
+          ? emit('scroll:top')
+          : emit('scroll:bottom')
+      }
+    })
+  }
+
+  const targets = computed<HTMLElement[]>(() => {
+    const list: HTMLElement[] = []
+
+    if (isObservable.value) {
```
m0ksem, line 41: Do we need this check? This will be bound to tor and bottom triggers which is also DOM elements.

m0ksem (approved):

PR #2393, opened 2022-09-13
[#2316] Update `VaDataTable` filtering example in docs with debounced input

### LINKED ISSUE #2316: Datatable filter must be debounced
> datatable filter should wait for user to finish typing or backspaceing (e.g. for 0.7 seconds) before re-initiating filtering -> makes filter unresponsive on big tables or slow devices

RVitaly1978:
## Description
close #2316 

chrome-capture-2022-8-13

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/data-table/examples/Filtering.vue
+++ b/packages/docs/src/page-configs/ui-elements/data-table/examples/Filtering.vue
@@ -1,16 +1,23 @@
 <template>
   <div class="row">
     <va-input
-      class="flex mb-2 md6"
+      class="flex mb-2 md6 xs12"
       placeholder="Filter..."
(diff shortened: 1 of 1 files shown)


aluarius (approved):
Asva has said that we should create examples with script setup because we're Vue 3, not Vue 2 UI Kit. But up to you - anyway, there will be a huge task to rewrite all examples.

rustem-nasyrov (approved):

m0ksem (approved):
It is a common to copy-paste example demo to user's project. It is annoying to remove "settings" stuff like this "debounced" checkbox. So, it might be two different demos instead one with settings.

It is OK now, let's split this in two demos after table of contents component will be available in docs.

Issue #2349, opened 2022-09-05
Change `useKeyboardOnlyFocus` composable to `:focus-visible` CSS selector


aluarius:
It makes the same but in more native and easy way.

- [x] Integrate to devel branch.
- [x] Delete composable from 1.5.0 branch.

PR #2394, opened 2022-09-13
[#2349] Keyboard-only focus-outline composable changed to CSS selector

aluarius:
Closes: #2349 

## Description
Changes keyboard-only focus style state from composable to native CSS selector.

## Types of changes


packages/ui/src/styles/resources/_mixins.scss
```diff
@@ -200,8 +200,8 @@
   outline-offset: $offset;
 }
 
-@mixin keyboard-focus($radius: 2px, $width: 2px, $offset: null, $style: solid) {
-  &--keyboard-focus {
+@mixin keyboard-focus-outline($radius: 2px, $width: 2px, $offset: null, $style: solid) {
```
RVitaly1978, line 203: Maybe having the default `$radius` equal to `'inherit'` is more convenient?

RVitaly1978 (approved):

Issue #2403, opened 2022-09-18
Vitest doesn't work with vuestic ui


asvae:
**Vuestic-ui version:** 1.4.10

Here's repo with repro
https://github.com/asvae/my-vue-app-tmp2

There are 2 things that go wrong:

### 1.4.6 and prior:

```
[Vue warn]: injection "Symbol(GLOBAL_CONFIG)" not found.
  at <VaButton>
  at <VueExample ref="VTU_COMPONENT" >
  at <VTUROOT>
```

Injection is missing for some reason. I tried different ways to provide config, nothing worked. I tested vue-test-utils provide/inject - it worked fine.

### 1.4.7 and later:

createVuestic is undefined

Which breaks everything

---

So I would recommend to go in this order:
* Figure out why createVuestic is undefined (we can do that by comparing with <=1.4.6 bundle) and fix it.
* Then work on fixing injects.
----

Worth noting that `dev` version of vite works just fine, it's just vitest that doesn't work.

PR #2409, opened 2022-09-19
Fix/test utils

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2403

Missed injection cause cuz plugin was not registered, because it was undefined.


asvae (approved):
Thanks for effort!

PR #2418, opened 2022-09-20
refactor search to use vue-router

Azgiliat:
## Types of changes


packages/docs/src/components/header/components/algolia-search/AlgoliaSearch.vue
```diff
@@ -2,34 +2,53 @@
...
+            event.preventDefault()
+            router.push({
+              path: hit.url.replace(/^https.*\/[a-z]{2}\//, `/${locale.value}/`),
+            })
+          },
+          children,
+        },
+        __v: null,
+      }
+    },
```
m0ksem, line 41: You can get rid of ref and key, group contructor and __v. 

```suggestion
    // absolutely kekw but docsearch is based on React, so we simulate React.createElement()
    // @ts-ignore
    hitComponent ({ hit, children }) {
      const url = new URL(hit.url)

      url.hostname = window.location.hostname

      return {
        type: 'a',
        props: {
          href: url,
          target: '_blank',
          onClick: (event: MouseEvent) => {
            event.preventDefault()
            router.push({
              path: hit.url.replace(/^https.*\/[a-z]{2}\//, `/${locale.value}/`),
            })
          },
          children,
        },
        // Need to create empty contructor and __v as null.
        __v: null,
        constructor: undefined,
      }
```

packages/docs/src/components/header/components/algolia-search/AlgoliaSearch.vue
```diff
@@ -2,34 +2,53 @@
...
+        constructor: undefined,
+        key: undefined,
+        props: {
+          href: hit.url,
+          target: '_blank',
+          onClick: (event: MouseEvent) => {
+            event.preventDefault()
+            router.push({
+              path: hit.url.replace(/^https.*\/[a-z]{2}\//, `/${locale.value}/`),
+            })
```
m0ksem, line 35: This repeated router.push can be moved to method. Name of method can give as a hit what a `/^https.*\/[a-z]{2}\//` does.

m0ksem (approved):

PR #2421, opened 2022-09-20
Fix tests and refactor colors case notation

rustem-nasyrov:
## Description
Fixed tests, did a little refactoring.

image

## Types of changes

### DIFF
--- a/packages/docs/src/config/theme-config.ts
+++ b/packages/docs/src/config/theme-config.ts
@@ -12,43 +12,43 @@ export const ThemeNameIterator = [
 
 export const COLOR_THEMES: Record<ThemeName, Record<string, string>> = {
   [ThemeName.DEFAULT]: {
-    primary: '#2550C0',
-    secondary: '#767C88',
-    success: '#40E583',
-    info: '#2C82E0',
-    danger: '#DE1041',
-    warning: '#FFAC0A',
+    primary: '#2550c0',
+    secondary: '#767c88',
+    success: '#40e583',
+    info: '#2c82e0',
+    danger: '#de1041',
+    warning: '#ffac0a',
 
-    backgroundPrimary: '#FFFFFF',
-    backgroundTertiary: '#FFFFFF',
-    backgroundSecondary: '#F4F8FA',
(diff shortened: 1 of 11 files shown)


packages/ui/src/composables/useColors.ts
```diff
@@ -88,6 +88,10 @@ export const useColors = () => {
       return prop
     }
 
+    if (isCSSVariable(prop)) {
```
aluarius, line 91: Why not:

image

Or it wa changed before?

aluarius (approved):

packages/ui/src/composables/useColors.ts
aluarius: It's wrong test condition, not composables. 

was: [['var(--va-primary)', undefined, false], '#2C82E0'],
now: [['var(--va-info)', undefined, false], 'var(--va-info)'],

It shouldn't return variable but hex.

packages/docs/src/config/theme-config.ts
```diff
@@ -14,29 +14,29 @@ export const COLOR_THEMES: Record<ThemeName, Record<string, string>> = {
...
     primary: '#2550C0',
     secondary: '#767C88',
-    success: '#40e583',
-    info: '#2c82e0',
+    success: '#40E583',
+    info: '#2C82E0',
     danger: '#DE1041',
     warning: '#FFAC0A',
 
-    backgroundPrimary: '#ffffff',
```
asvae, line 22: Can't we use lowercase colors?

In all other places it's lowercase

m0ksem:
```ts
{
 danger: '#f0f',
 'background-primary': '#fff
}
```

It doesn't look too nice. BTW, it is not case-sensitive.

aluarius (approved):

packages/docs/src/config/theme-config.ts
asvae: @m0ksem meant hexes 😄, so `#ffac0a` instead of `#FFAC0A`.

Issue #2420, opened 2022-09-20
VaDatePicker shouldn't have yellow outline on hover


asvae:
**Vuestic-ui version:** 1.4.11

<img width="356" alt="image" src="https://user-images.githubusercontent.com/5394573/191277072-38c35943-a9e0-4e0a-97b7-984fd1db422a.png">

r-Dev03:
@asvae can I take on this issue?

vrindavan:
This file?
https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-date-picker/components/VaDatePickerCell.vue

VBoudreau55:
Can I be assigned to this issue, I think it would be a great first issue for me

rustem-nasyrov:
> This file? [VaDatePickerCell.vue](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-date-picker/components/VaDatePickerCell.vue)

Yeah, this file.

rustem-nasyrov:
> @r-Dev03, @vrindavan, @VBoudreau55 hi! Of course you can assign yourselves, but before you start, please read the [contribution guide](https://vuestic.dev/en/contribution/guide). :)

My bad, I'll assign someone.

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/2422

PR #2422, opened 2022-09-20
[#2420] Removed outline on datepicker

VBoudreau55:
## Description
Removed the @include focus-outline; line to remove outline when the element is focused

## Types of changes


aluarius:
@m0ksem we want to remove it at all, or change to keyboard-only?

PR #2425, opened 2022-09-21
`VaButton` background color state should rely on CSS, no on JS

aluarius:
## Description
`VaButton` background state colors now relies on CSS, not JS. Related minor fixes.

## Types of changes


m0ksem (approved):
Idk, plain button looks good for me, but too sharp. How about transition?

Found a few bugs:

Strange reaction
firefox_KqFB2qsngy

No reaction
firefox_CZNhbPSorI

aluarius:
> No reaction

Plain preset -> only text is affected by focus-press -> there're no text. No bug, but bad example, will remove it.

> Strange reaction

Just an extremal opacity hover state was set here, will change example.

Issue #2419, opened 2022-09-20
Flashbang effect on docs


asvae:
**Vuestic-ui version:** 1.4.11

<img width="1509" alt="image" src="https://user-images.githubusercontent.com/5394573/191276339-59600d06-80ae-40b9-a049-0fd5704241e2.png">

When you open docs on dark theme - it shows you white screen briefly, then goes to dark mode. That hurts.
Ideally there shouldn't be any brightness transition.

m0ksem:
This is caused because of SPA. We're not able to make page dark before SPA is loaded. There will be a dark or white flash bang in any case. We might make some hacks here, but let's discuss it before.

PR #2428, opened 2022-09-21
Fix/dark theme flashbang

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2419

Optimized only for production. Idk how to fix this in SPA... We need to wait until all js loaded and parsed and then bg can be changed.

### DIFF
--- a/packages/docs/src/components/header/Header.vue
+++ b/packages/docs/src/components/header/Header.vue
@@ -104,7 +104,6 @@ export default class Header extends mixins(PropsMixin) {
 
 <style lang="scss" scoped>
   @import "~vuestic-ui/src/styles/resources";
-  @import "~vuestic-ui/src/styles/global";
 
   .header {
     width: 100%;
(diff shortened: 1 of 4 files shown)


asvae:
@m0ksem can you check this? https://caniuse.com/prefers-color-scheme

m0ksem:
> @m0ksem can you check this? https://caniuse.com/prefers-color-scheme

I could hardcode colors for docs, but it is not solution for vuestic-ui. User must hardcode it too. 

Flash caused because body has white BG until we register CSS variables. Unfortunately, we need to wait for all components to be registered and then CSS variables can be rendered in DOM.

I thought about running CSS variables update before registering components. Registering components takes 90% of all rendering time.

m0ksem:
Require https://github.com/epicmaxco/vuestic-ui/pull/2438 to be merged before.

asvae:
Can't we just hack a small script into docs `index.html`'s head?

m0ksem:
> Can't we just hack a small script into docs `index.html`'s head?

Might be fixable after https://github.com/epicmaxco/vuestic-ui/issues/2623

m0ksem:
Hope to fix this in nuxt later.

Issue #2362, opened 2022-09-05
Select text-by broken if value-by isn't empty


Luffyyy:
**Vuestic-ui version:** 1.4.10

Basically the select will show the value rather than text-by's text, possibly due to relying on the value-by.

**Steps to reproduce**
Same as before:
```vue
<template>
    {{test}}
    <div>
        <va-select v-model="test" :options="options" value-by="value" multiple/>
    </div>
</template>

<script setup>
const test = ref([]);
const options = [
    { 
        text: 'Hello',
        value: 1
    },
    { 
        text: 'World',
        value: 2
    },
    { 
        text: 'Test',
        value: 3
    }
];
</script>
```

**What is the expected behavior?**
![image](https://user-images.githubusercontent.com/11510099/188510781-bcb814c9-727d-4e57-af2c-c745126eb40c.png)

**What is the current behavior?**
![image](https://user-images.githubusercontent.com/11510099/188510794-9c15782f-a4f5-413b-8b00-aaa40b213f48.png)

**Other information**
😅 hopefully last issue report regarding the select

stuk88:
I get it too

stuk88:
So for anyone struglling with that as myself.
when textBy is "number" or "boolean" it will act strange. 
change the name of the label to a different name, like `label` or `text` 
And than  valueBy="value" is must.

m0ksem:
> So for anyone struglling with that as myself. when textBy is "number" or "boolean" it will act strange. change the name of the label to a different name, like `label` or `text` And than valueBy="value" is must. Oh and it acts wierd when the model is object

Hi, can you reproduce it in https://ui.vuestic.dev/play/ and share here?

Issue #2136, opened 2022-07-22
Select track-by make fallback to value-by or text-by


m0ksem:
`track-by` shoudl work similarly to `value-by` (if not set - should be set by object identity).

Look at useSelectableList.ts

asvae:
I suggest we just deprecate track-by altogether. No point having it.

PR #2429, opened 2022-09-22
[#2395] [#2362] [#2136] `VaSelect` bugfix, minor refactoring

aluarius:
Closes: #2395 
Closes: #2362 
Closes: #2136

## Description
Separate bugfixes related to the `VaSelect` component.

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useSelectableList.ts
+++ b/packages/ui/src/composables/useSelectableList.ts
@@ -36,7 +36,10 @@ export function useSelectableList (props: ExtractPropTypes<typeof useSelectableL
 
   const getText = (option: SelectableOption) => getOptionProperty(option, props.textBy)
   const getDisabled = (option: SelectableOption) => getOptionProperty(option, props.disabledBy)
-  const getTrackBy = (option: SelectableOption) => getOptionProperty(option, props.trackBy)
+  const getTrackBy = (option: SelectableOption) => {
+    if (props.trackBy) { return getOptionProperty(option, props.trackBy) }
+    return getValue(option)
+  }
   const getGroupBy = (option: SelectableOption) => getOptionProperty(option, props.groupBy)
 
   return {
--- a/packages/ui/src/services/utils.ts
+++ b/packages/ui/src/services/utils.ts
@@ -74,21 +74,22 @@ export const getValueByPath = <T extends Record<string, unknown>>(option: T, pro
 }
 
 /**
- * Finds value of nested property inside of an object.
+ * Finds value of nested property inside an object.
  *
- * @param option - Object to look properties inside
- * @param prop - string or function used to find nested property
+ * @param option - Object to look properties inside.
+ * @param prop - String or function used to find nested property.
  */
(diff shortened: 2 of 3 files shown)


m0ksem (approved):

packages/ui/src/services/utils.ts
```diff
@@ -74,21 +74,22 @@ export const getValueByPath = <T extends Record<string, unknown>>(option: T, pro
...
-  // Can't access not object option
-  if (typeof option !== 'object' || !option) { return undefined }
+  if (!option || typeof option !== 'object' || Array.isArray(option)) { return undefined }
 
   if (!prop) { return option }
   if (typeof prop === 'string') { return getValueByPath(option, prop) }
   if (typeof prop === 'function') { return prop(option) }
+
+  // if `prop` has different to string or function type and can't be processed
   return option
```
RVitaly1978, line 93: I think we should return `undefined` if we can't get the value by the key

packages/ui/src/composables/useSelectableList.ts
```diff
@@ -36,7 +36,10 @@ export function useSelectableList (props: ExtractPropTypes<typeof useSelectableL
 
   const getText = (option: SelectableOption) => getOptionProperty(option, props.textBy)
   const getDisabled = (option: SelectableOption) => getOptionProperty(option, props.disabledBy)
-  const getTrackBy = (option: SelectableOption) => getOptionProperty(option, props.trackBy)
+  const getTrackBy = (option: SelectableOption) => {
+    if (props.trackBy) { return getOptionProperty(option, props.trackBy) }
+    return getValue(option)
+  }
```
RVitaly1978, line 42: I think we should treat other methods in the same way.

But maybe it would be better to rework the function:
```ts
const getOptionProperty = (option: SelectableOption, selector: StringOrFunction) => {
  if (isStringOrNumber(option)) {
    return option
  }

  return selector
    ? getValueByKey(option, selector)
    : getValue()
}
```

RVitaly1978 (changes requested):
Perhaps we should set the default value of `'valueBy`' to `'value'`?
```ts
export const useSelectableListProps = {
  options: { type: Array as PropType<SelectableOption[]>, default: () => [] },
  textBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'text' },
  valueBy: { type: [String, Function] as PropType<StringOrFunction>, default: '' },
  trackBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'value' },
  disabledBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'disabled' },
  groupBy: { type: [String, Function] as PropType<StringOrFunction>, default: 'group' },
}
```

packages/ui/src/composables/useSelectableList.ts
aluarius: Discussed.

packages/ui/src/services/utils.ts
aluarius: Discussed.

PR #2431, opened 2022-09-22
feat: added stub tsconfig script

m0ksem:
Now after `yarn install` `.nuxt/tsconfig.json` will be created, so there will not be an error while running `yarn build` in packages/nuxt.

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -40,6 +40,7 @@
     "@nuxt/schema": "latest",
     "@nuxt/vite-builder": "latest",
     "@nuxtjs/eslint-config-typescript": "latest",
-    "eslint": "latest"
+    "eslint": "latest",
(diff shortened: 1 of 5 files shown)


aluarius (approved):

packages/nuxt/package.json
```diff
@@ -40,6 +40,7 @@
     "@nuxt/schema": "latest",
     "@nuxt/vite-builder": "latest",
     "@nuxtjs/eslint-config-typescript": "latest",
-    "eslint": "latest"
+    "eslint": "latest",
+    "when-dependencies-installed": "^1.0.0"
```
RVitaly1978, line 44: Should we add the same script in this package as in the sandbox package?
`"dependenciesInstalled": "node ./scripts/stud-nuxt.mjs"`

RVitaly1978 (approved):

Issue #2354, opened 2022-09-05
Themes config


m0ksem:
```
type ThemeOptions = DeepPartial<{
   colors: {
      // Accents:
      primary: '#f00'
      success
      ...

      // Backgrounds
      background-primary,
      background-secondary

      // Elements background
      background-el-primary
      background-el-secondary

      // Text color
      text-light // Text used on dark background
      text-dark // Text used on light background

      // Other
      focus, shadow
   },
   lightThreshold: 140 // 0-225
   variantDarkness: {
      hover: 0.8, // 0.2 lighter
      focus: 1.2, // 0.2 darken
      press: 1.5 // SUPER-mega-dark
   },
   componentColors: {
      VaNavbar: '#eee',
      VaSidebar: 'primary'
   }
}>
```

asvae:
lightThreshold and variantDarkness are free for us, as these values are already hardcoded.

asvae:
Suggested change:

```
colors :{
  variables: {
      // Backgrounds
      background-primary,
      background-secondary

      // Elements b
  },
  lightThreshold: 140 // 0-225
   variantDarkness: {
      hover: 0.8, // 0.2 lighter
      focus: 1.2, // 0.2 darken
      press: 1.5 // SUPER-mega-dark
   }
}
```

```
theme
  colors
  icons
  components
  breakpoints
```

PR #2438, opened 2022-09-25
feat: added threshold to colors config

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2354

@aluarius, should we handle hover and focus color modificators in colors config?


aluarius:
> should we handle hover and focus color modificators in colors config?

@m0ksem i'm not sure if it will work with components presets.

packages/ui/src/services/color-config/color-functions.ts
```diff
@@ -6,14 +6,14 @@ export const colorToRgba = (color: ColorInput, opacity: number) => {
   return new ColorTranslator(color).setA(opacity).RGBA
 }
 
-export const isLightBackground = (color: ColorInput) => {
+export const isLightBackground = (color: ColorInput, threshold: number) => {
```
aluarius, line 9: Where is it used inside `isLightBackground`?

aluarius (approved):

packages/ui/src/services/color-config/color-functions.ts
```diff
@@ -6,14 +6,14 @@ export const colorToRgba = (color: ColorInput, opacity: number) => {
...
   // TODO: replace with color mixin (doesn't play any role in dark theme)
   const { R, G, B } = new ColorTranslator(color)
   return Math.sqrt(R * R * 0.241 + G * G * 0.691 + B * B * 0.068) > 120
```
m0ksem, line 12:
```suggestion
  return Math.sqrt(R * R * 0.241 + G * G * 0.691 + B * B * 0.068) > threshold
```

packages/ui/src/services/color-config/color-functions.ts
aluarius, line 9: Mb we want to set threshold default value to 120 here in purpose not to pass it everywhere?

aluarius (approved):

packages/ui/src/composables/useColors.ts
```diff
@@ -110,6 +116,10 @@ export const useColors = () => {
       }, {})
   }
 
+  const getTextColor = (color: ColorInput, darkColor = 'textDark', lightColor = 'textLight') => {
+    return isLightBackground(color, 120) ? darkColor : lightColor
+  }
```
RVitaly1978, line 121: Can't we use `getTextColor` from `'../services/color-config/color-functions'`?

packages/ui/src/services/color-config/color-functions.ts
```diff
@@ -6,14 +6,14 @@ export const colorToRgba = (color: ColorInput, opacity: number) => {
...
   // TODO: replace with color mixin (doesn't play any role in dark theme)
   const { R, G, B } = new ColorTranslator(color)
-  return Math.sqrt(R * R * 0.241 + G * G * 0.691 + B * B * 0.068) > 120
+  return Math.sqrt(R * R * 0.241 + G * G * 0.691 + B * B * 0.068) > threshold
 }
 
-export const getTextColor = (color: ColorInput, darkColor = 'textDark', lightColor = 'textLight') => {
-  return isLightBackground(color) ? darkColor : lightColor
+export const getTextColor = (color: ColorInput, darkColor: string, lightColor: string, threshold: number) => {
+  return isLightBackground(color, threshold) ? darkColor : lightColor
```
RVitaly1978, line 16: maybe we can set default values to `darkColor = 'textDark'`, `lightColor = 'textLight'`.
Shouldn't we use `threshold` from `globalConfig.colors.threshold` by default?

packages/ui/src/services/color-config/plugin/create-color-config-plugin.ts
```diff
@@ -32,20 +33,17 @@ export const createColorConfigPlugin = (app: App) => {
...
+      setCSSVariable(`on-${key}`, getTextColor(newValue[key], newValue.textDark, newValue.textLight, globalConfig.value.colors!.threshold), root)
     })
   }
 
-  updateColors(globalConfig.value.colors)
+  updateColors(globalConfig.value.colors!.variables)
 
   watch(() => globalConfig.value.colors, (newValue) => {
-    updateColors(newValue)
+    updateColors(newValue?.variables)
```
RVitaly1978, line 43: Should we be looking at changes to the entire color configuration here, or are just the `variables` enough?
`watch(() => globalConfig.value.colors.variables, (newValue) => {`

RVitaly1978 (approved):

packages/ui/src/services/color-config/color-functions.ts
m0ksem: It is used in useColors, cuz it we can use globalConfig only in setup context.

Issue #2308, opened 2022-08-26
Dropdown doesn't ignore overflow from parent


asvae:
![image](https://user-images.githubusercontent.com/5394573/186943173-f53f23e0-2122-4571-aa96-aa8a6af13789.png)

In [these demos](http://vuestic-ui-demo.sub.asva.by/#/demo/va-dropdown/VaDropdown.demo.vue) - dropdown should overflow container. 

The way I understand it - dropdown by default should be appended to `body`. This helps to avoid overflow issues (which is very tricky for users to deal with).

But we also have (had?) `DisableAttachment` flag, which allows to render dropdown inside a parent.

Ideally we want to support both ways (via a flag), but `body` should go by default.

@m0ksem your thoughts on this would be appreciated. 😼

m0ksem:
Old demo. This can be resolved by adding `prevent-overflow` prop to dropdown.

asvae:
I think we should fix demo then.

PR #2439, opened 2022-09-26
Fix#2270/html element as dropdown target

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2308
closes https://github.com/epicmaxco/vuestic-ui/issues/2270


aluarius (approved):
Useful composable!

PR #2442, opened 2022-09-27
release(1.4.12): fix missing essential.css file in build.

m0ksem:
Made release/1.4.12 branch, so we still can support 1.4.*.

We had 1.4.11 without essenatil.css file.
Then I made release without types....
Now we have 1.4.13 :)


asvae (approved):
👍

PR #2443, opened 2022-09-27
chore: update contribution guide documentation

m0ksem:
* removed useless documentation blocks.
  like npm scripts which user likely even do not need to use.
* added packages description.
* reorder guides so they are easier to read.

Later in md docs we can make an advanced guide, but I don't think we need it for now.

BTW, help me find if I missed here.


rustem-nasyrov:
Looks like we have broke padding for the code-blocks:
develop: 
image
production:
image

rustem-nasyrov (approved):

RVitaly1978 (approved):

packages/docs/src/locales/en/en.json
```diff
@@ -3659,6 +3659,45 @@
   "guide": {
     "title": "Contribution guide",
     "description": "This guide describes the possible ways in which you can become a part of the ongoing development of Vuestic UI. We are really excited that you are interested in helping us make Vuestic better!",
+    "quickStart": {
```
asvae, line 3662: Maybe we can use md for that?

to not bother with translations etc?

PR #2461, opened 2022-10-05
Add "Meet the Team" page, add blog link to the landing header

zvenigorodskaia:
## Description

- add  Meet the Team page
- add Blog link to the landing header

## Markup:

image

<details>

```vue
// Your code
```
</details>

## Types of changes


m0ksem (changes requested):
image

Images has strange dots caused by border in Firefox. I would prefer making two different borders instead of playing with background gradient.

Let's also make images squared and smaller (from 700px to 200px) in source. You can run them through https://tinypng.com/.

packages/docs/src/components/landing/LandingEpicmaxBanner.vue
```diff
@@ -50,10 +50,7 @@ export default class EpicmaxBanner extends Vue {
       display: flex;
```
asvae, line 50: I suggest we compress images a bit, 240/240 might be enough

image

PR #2462, opened 2022-10-05
Fix building issues

rustem-nasyrov:
## Description
Refactoring, fixed types.

## Types of changes


packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -641,7 +643,7 @@ export default defineComponent({
       ...filterComponentProps(props, VaDropdownProps).value,
       closeOnContentClick: closeOnContentClick.value,
       stateful: false,
-      offset: [1, 0],
+      offset: [1, 0] as DropdownOffsetProp,
```
m0ksem, line 646: Do we need this cast? [1, 0] should be valid DropdownOffsetProp without casting.

You probably wanted to extract props from VaDropdown and make this type as return type of computed.

m0ksem (approved):

packages/ui/src/components/va-select/VaSelect.vue
m0ksem: > @m0ksem, nope, typescript check says that `number[]` isn't incompatible with VaDropdown prop `number | [number, number]`.

This will not be an issue, if you cast whole object to VaDropdownProps. 

It is not important, BTW.

Issue #2434, opened 2022-09-23
i18n config


m0ksem:
We have the couple of texts used in app (for example, https://github.com/epicmaxco/vuestic-ui/issues/2430).

```ts
createVuestic({
   i18n: {
      ok: 'Так',
      no: 'Ні'.
      search: 'Пошук'
      // ...
   }
})
```

It must be easily integrated to vue-i18n if user has few languages across app.

Points for research:
* fallback
* integration with vue-i18n (if possible)

m0ksem:
See https://github.com/epicmaxco/vuestic-ui/issues/2381

Issue #2381, opened 2022-09-09
Translations for WAI-ARIA attributes


aluarius:
<img width="867" alt="image" src="https://user-images.githubusercontent.com/64714442/189336330-d5d8e66e-650c-432a-af18-b749ac04d578.png">

asvae:
Investigate how it's done in other frameworks.

PR #2471, opened 2022-10-08
Feat/i18n

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2434
closes https://github.com/epicmaxco/vuestic-ui/issues/2381

I18n support for Vuestic components props and aria-label attributes.

Need docs after review, but I wrote enough in comments of useI18nConfig composable to make it clear.

Take a look commit by commit.


packages/ui/src/composables/useTranslation.ts
```diff
@@ -0,0 +1,36 @@
...
+  return key
+}
+
+export const useTranslation = () => {
+  const { globalConfig, mergeGlobalConfig } = useGlobalConfig()
+
+  const config = computed(() => globalConfig.value.i18n)
+
+  return {
+    /** Translate prop. Translate only if key has `t:` prefix */
```
m0ksem, line 22:
```suggestion
    /** Translate prop. Translate only if key has `$t:` prefix */
```

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -40,23 +40,23 @@
...
             v-bind="{ ...iconProps, ...clearIconProps }"
+            :aria-label="t('resetTime')"
             @click.stop="reset"
             @keydown.enter.stop="reset"
             @keydown.space.stop="reset"
           />
           <va-icon
             v-else-if="!$props.leftIcon"
-            aria-label="toggle dropdown"
+            :aria-label="t('toggleDropdown')"
```
m0ksem, line 59: I might miss something to translate. Just run thought :aria-label and props. @aluarius, do you think something else must be translated too?

aluarius:
Closes https://github.com/epicmaxco/vuestic-ui/issues/2381

packages/ui/src/components/va-time-input/VaTimeInput.vue
aluarius: Some aria attributes are set via computed attributes variable. Additional check for 'aria-label' & ariaLabel patterns is needed. Will try to find time next week for this.

All in all, aria attributes and default props values translations is enough probably (can't remember any other cases).

rustem-nasyrov (approved):

asvae (approved):
Don't see any issues 👍.

Flat translation strings are a bit troubling, but probably all right.

Issue #2336, opened 2022-09-01
Aspect ratio for VaCarousel component


asvae:
Should have similar api as image.

PR #2472, opened 2022-10-09
fix(#2336): proxy image props to carousel

m0ksem:
Proxy props for VaImage, so now user can specify ratio.

Previously it has 300px height by default, but now it has 16/9 ratio by default instead. You still can specify height BTW.

It will break current carousel layout for some people.

image
image


rustem-nasyrov (changes requested):
In the carousel docs images became really big. Some of the given examples are broken.
<details>
<summary>Some broken demos</summary>
image
image
</details>

rustem-nasyrov (approved):
Great job! Keep it up. :)

Issue #1936, opened 2022-06-13
Move all vuestic icons to aliases.


m0ksem:
We have an aliases feature, which allow us to set custom icon name for any icon.

We can provide list of icons that we use in our components as aliases with 'va' prefix.

Then we can use them in VaSelect, VaModal, VaDatePicker etc.

User must be able to easy change aliases if he disslike material icons.

For example, 

va-expand-arrow, va-collapse-arrow, va-left-arrow, va-clocks etc.

Must be a page in configuration docs.

m0ksem:
We also need docs with our icon names.

PR #2473, opened 2022-10-09
fix#1936: va icon aliases

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1936

image


rustem-nasyrov (approved):

packages/docs/src/locales/en/en.json
```diff
@@ -3844,6 +3844,10 @@
         "title": "Example aliases config",
         "about": "Below we can see how the icons would be transformed",
         "explain": "As you can see alias config is more important than font config. And `va-icon` props take precedence over both."
+      },
+      "vaAliases": {
+        "title": "Vuestic component alises",
+        "text": "We use some icons in our components by default. You can redefined them by changing it's alias."
```
m0ksem, line 3850:
```suggestion
        "text": "We use some icons in our components by default. You can redefine them by changing it's alias."
```

PR #2476, opened 2022-10-10
Fix/minor docs bugs

m0ksem:
Fix button grow:
image
Margins:
image

Added prefixes to all typography classes
We might need to rename va-display-1 to va-h1 to make it more clear.
image

### DIFF
--- a/packages/docs/src/config/vuestic-config.ts
+++ b/packages/docs/src/config/vuestic-config.ts
@@ -16,6 +16,14 @@ export const VuesticConfig: Partial<GlobalConfig> = {
       VaButton: {
         addToCart: { size: 'large', round: true },
         deleteFromCart: { size: 'small', plain: true },
+        landingHeader: {
+          target: '_blank',
+          size: 'large',
+          plain: true,
+          'hover-behaviour': 'mask',
+          'hover-mask-color': 'textDark',
+          'hover-opacity': '1',
+        },
       },
     },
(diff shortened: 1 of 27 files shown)


packages/docs/src/config/vuestic-config.ts
```diff
@@ -16,6 +16,14 @@ export const VuesticConfig: Partial<GlobalConfig> = {
       VaButton: {
         addToCart: { size: 'large', round: true },
         deleteFromCart: { size: 'small', plain: true },
+        landingHeader: {
+          target: '_blank',
```
rustem-nasyrov, line 20: Do we really need to open 'Overview' a.k.a docs pages in the new tab?

packages/docs/src/locales/en/en.json
```diff
@@ -3354,6 +3354,10 @@
         "title": "Sizes",
         "text": "You can set different sizes."
       },
+      "grow": {
+        "title": "Grow",
+        "text": "Makes button group grow to the width of its container."
```
rustem-nasyrov, line 3359: Missing translation in the russian, it's not necessary for now, because we have no translations in some pages too. I'll move it to the new issue, where we can check all pages for missing translations.

m0ksem: Yeah. This is a thing I don't care for now.

PR #2484, opened 2022-10-11
fix: multiple select options value-by

m0ksem:
Now ValueComputed returns SelectOption and emits valueBy correctly.

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.vue
+++ b/packages/ui/src/components/va-select/VaSelect.vue
@@ -294,19 +294,19 @@ export default defineComponent({
 
     const valueComputed = computed<SelectOption | SelectOption[]>({
       get () {
-        const value = getOptionByValue(props.modelValue)
-
         if (props.multiple) {
-          if (!value) {
+          if (!props.modelValue) {
             return []
           }
 
-          if (!Array.isArray(value)) {
-            return [value]
+          if (!Array.isArray(props.modelValue)) {
+            return [getOptionByValue(props.modelValue)]
           }
 
-          return value.map(getValue)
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -294,20 +294,20 @@ export default defineComponent({
...
           }
 
-          if (!Array.isArray(value)) {
-            return [value]
+          if (!Array.isArray(props.modelValue)) {
+            return [getOptionByValue(props.modelValue)]
           }
 
-          return value.map(getValue)
+          return props.modelValue.map((v) => getOptionByValue(v))
```
rustem-nasyrov, line 306: We can pass the `getOptionByValue` method directly into the map's callback. It doesn't accept any other arguments rather than value.
```suggestion
          return props.modelValue.map(getOptionByValue)
```

packages/ui/src/composables/useMaxSelections.ts
```diff
@@ -23,9 +23,8 @@ export function useMaxSelections (
     return selections.value.length >= maxSelections.value
   }
 
-  const addOption = (optionToAdd: any): void => {
-    const newSelectedOptions = [...selections.value, optionToAdd]
-    emit('update:modelValue', newSelectedOptions)
```
rustem-nasyrov, line 28: Also would be great to remove `emit` from the function arguments. :) And also remove emits where it was passed.

rustem-nasyrov (changes requested):

rustem-nasyrov (approved):

PR #2490, opened 2022-10-11
Fix/form reset behavior

### LINKED ISSUE #2441: Form or form items reset (clear the value) triggers validation
**Vuestic-ui version:** 1.4.11

**Steps to reproduce**
```vue
<template>
  <button @click="$refs.field.reset()">Clear input</button>
  <va-input
    ref="field"
    v-model="value"
    :rules="[(value) => !!value || 'Field is required']"
  />
</template>

<script>
  export default {
    components: { VaInput },
    data: () => ({ value: '' })
  }
</script>
```
The form or its field resets.
Getting validation error.
**Other information**
https://codesandbox.io/s/relaxed-gianmarco-3iv3f7?file=/src/App.vue

rustem-nasyrov:
Closes #2441

## Description
- Added support to reset the form items without triggering the validation.

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useSelectable.ts
+++ b/packages/ui/src/composables/useSelectable.ts
@@ -1,4 +1,4 @@
-import { PropType, computed, SetupContext, ShallowRef, ExtractPropTypes } from 'vue'
+import { PropType, computed, ref, SetupContext, ShallowRef, ExtractPropTypes } from 'vue'
 
 import { useStateful, useStatefulProps, StatefulProps } from './useStateful'
 import { useLoadingProps, LoadingProps } from './useLoading'
@@ -62,11 +62,29 @@ export const useSelectable = (
 ) => {
   checkDuplicates(props)
 
-  /** @public */
-  const reset = () => emit('update:modelValue', false)
-  const focus = () => input.value?.focus()
+  const isValueReset = ref(false)
 
-  const { computedError, computedErrorMessages, validate, validationAriaAttributes } = useValidation(props, emit, reset, focus)
+  /** @public */
+  const reset = () => {
+    isValueReset.value = true
+    emit('update:modelValue', false)
+  }
+
+  const focus = () => {
+    isValueReset.value = false
+    input.value?.focus()
+  }
+
+  const {
+    computedError,
+    computedErrorMessages,
+    validate,
+    validationAriaAttributes,
+  } = useValidation(props, emit, {
+    reset,
+    focus,
+    isValueReset,
+  })
   const { valueComputed } = useStateful(props, emit)
   const { isFocused } = useFocus()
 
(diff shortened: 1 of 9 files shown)


packages/ui/src/composables/useSelectable.ts
```diff
@@ -62,11 +62,29 @@ export const useSelectable = (
 ) => {
   checkDuplicates(props)
 
+  const isValueReset = ref(false)
+
   /** @public */
-  const reset = () => emit('update:modelValue', false)
-  const focus = () => input.value?.focus()
+  const reset = () => {
+    isValueReset.value = true
```
m0ksem, line 69: This looks overall confusing for me. We use reset method in useValidation. How about handle this issue in scope of this method? I mean, if value changed with `resetValidation` method from `useValidation`, then we skip validation.

This way you will not need to pass additional options, etc.

m0ksem (changes requested):

packages/ui/src/composables/useSelectable.ts
rustem-nasyrov: Reset validation in the `useValidation` composable resets only validation, not the component.

packages/ui/src/composables/useValidation.ts
```diff
@@ -50,8 +65,16 @@ export const useValidation = <V, P extends ExtractPropTypes<typeof useValidation
     computedErrorMessages.value = []
   }
 
+  const withoutValidation = (cb: () => any): void => {
+    canValidate.value = false
+    cb()
+    resetValidation()
```
m0ksem, line 71: Why would a function that executes callback without validation resetValidation? This likely must be named withValidationRest().

This must look like this.
```suggestion
    canValidate.value = true
```

packages/ui/src/composables/useValidation.ts
```diff
@@ -37,10 +51,11 @@ export const useValidationEmits = ['update:error', 'update:errorMessages']
 export const useValidation = <V, P extends ExtractPropTypes<typeof useValidationProps>>(
   props: P,
   emit: (event: any, ...args: any[]) => void,
-  reset: () => any,
-  focus: () => any,
+  options: UseValidationOptions,
 ) => {
+  const { reset, focus } = options
   const { isFocused, onFocus, onBlur } = useFocus()
+  const canValidate = ref(true)
```
m0ksem, line 58: This might be simple `let`

m0ksem (changes requested):

packages/ui/src/composables/useValidation.ts
rustem-nasyrov: The suggested change won't work, I've tested it.

packages/ui/src/composables/useSelectable.ts
```diff
@@ -63,10 +63,21 @@ export const useSelectable = (
   checkDuplicates(props)
 
   /** @public */
-  const reset = () => emit('update:modelValue', false)
-  const focus = () => input.value?.focus()
+  const reset = () => withValidationReset(() => {
+    emit('update:modelValue', false)
+  })
```
m0ksem, line 68:
```suggestion
  const reset = () => withValidationReset(() => emit('update:modelValue', false))
```

packages/ui/src/composables/useValidation.ts
m0ksem: You likely need to do something like this

```ts
const reset = withoutValidation(() => {
   // do stuff here
   resetValidation()
})
```

Issue #2480, opened 2022-10-11
Check prefixes for typography helpers


asvae:
Discuss what we decide to do with typography helpers.

We discussed we want to rename `--` to `-`,

PR #2504, opened 2022-10-13
fix: typography double dash

m0ksem:
image

closes https://github.com/epicmaxco/vuestic-ui/issues/2480

### DIFF
--- a/packages/ui/src/styles/typography/_module.scss
+++ b/packages/ui/src/styles/typography/_module.scss
@@ -14,15 +14,15 @@
   .va-title {
     @include va-title();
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/styles/typography/_module.scss
```diff
@@ -15,80 +14,76 @@
   .va-title {
     @include va-title();
 
-    &--va-info {
```
rustem-nasyrov, line 18: Here is the result will be `.va-title-va-info`.

rustem-nasyrov (approved):

PR #2518, opened 2022-10-14
Fixed demos spacing.

### LINKED ISSUE #2509: Demo spacing
## Checkbox

image

image

image

## Select

image

## List
Lost spacing between items
image

Strange padding in list
image

Sadnessa:
## Description
Closes #2509.

## Types of changes

### DIFF
--- a/packages/docs/src/page-configs/ui-elements/list/examples/Clickable.vue
+++ b/packages/docs/src/page-configs/ui-elements/list/examples/Clickable.vue
@@ -1,10 +1,11 @@
 <template>
-  <va-list>
+  <va-list id="clickable-list-example">
     <va-list-label>
       Clickable
     </va-list-label>
 
     <va-list-item
+      class="list__item"
       v-for="(contact, index) in contacts"
       :key="index"
       to="/"
@@ -49,3 +50,13 @@ export default {
   },
 }
 </script>
+
+<style lang="scss">
+  #clickable-list-example {
+    padding: 0;
+
+    .list__item:not(:last-child) {
+      margin-bottom: 20px;
+    }
+  }
+</style>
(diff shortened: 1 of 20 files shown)


packages/docs/src/page-configs/ui-elements/list/examples/Clickable.vue
```diff
@@ -49,3 +50,13 @@ export default {
...
+
+<style lang="scss">
+  #clickable-list-example {
+    padding: 0;
+
+    .list__item:not(:last-child) {
+      margin-bottom: 20px;
+    }
+  }
+</style>
```
rustem-nasyrov, line 62: Let's rewrite it (in other demos too) like this:
```suggestion
<style>
.list-item + .list-item {
  margin-top: 20px;
}
</style>
```

packages/ui/src/components/va-list/VaList.vue
```diff
@@ -30,7 +30,6 @@ export default defineComponent({
   @import "variables";
 
   .va-list {
-    padding: var(--va-list-padding);
```
rustem-nasyrov, line 33: This change might break styles for end users.

rustem-nasyrov (changes requested):
Awesome! Thank you for your contribution, but there are still things to do.

packages/ui/src/components/va-list/VaList.vue
m0ksem: @rustem-nasyrov, should we care about it, if we are going to release 1.5.0?

rustem-nasyrov: @m0ksem, I think for now we don't, but I'm wondering is users will notice it or not. :)

m0ksem (approved):

rustem-nasyrov (approved):

PR #2535, opened 2022-10-18
fix: partial fix for va-image

### LINKED ISSUE #2488: VaImage doesn't show image on ssr
**Vuestic-ui version:** 1.5.0

image

In component `loading` is set to `true` which sets `display: none`. `loading = false` is set on image load, but for some reason image load doesn't happen (on client)

asvae:
fix #2488

### DIFF
--- a/packages/ui/src/components/va-image/VaImage.vue
+++ b/packages/ui/src/components/va-image/VaImage.vue
@@ -49,7 +49,7 @@ export default defineComponent({
     draggable: { type: Boolean, default: true },
   },
   setup (props, { emit }) {
-    const loading = ref(true)
+    const loading = ref(false)
     const loadingError = ref(false)
 
     const imageStyles = computed(() => ({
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-image/VaImage.vue
```diff
@@ -49,7 +49,7 @@ export default defineComponent({
     draggable: { type: Boolean, default: true },
   },
   setup (props, { emit }) {
-    const loading = ref(true)
+    const loading = ref(false)
```
m0ksem, line 52: Since issues related to SSR only we can stick to isServer

```suggestion
    const loading = ref(isClient())
```

m0ksem (commented):
I think you should handle loading state in onMounted hook, because loading state is hydrated from server.  Image can not be in loading state anyway until component is **mounted**.

Something like this make sense for me: 
```
onMounted(() => { loading.value = true })
```

I haven't run this PR, but issues also can be caused by `loadingError`.

packages/ui/src/components/va-image/VaImage.vue
asvae: We either get warns on hydration or image not being shown.

PR #2551, opened 2022-10-20
feat(Modal): add beforeClose prop

### LINKED ISSUE #2550: [Modal] before-close need prevent close
need some way to prevent modal close  if form validate fail 

the api like [element-plus](https://element-plus.org/zh-CN/component/dialog.html#%E5%B1%9E%E6%80%A7)
```
const handleClose = (done: () => void) => {
  ElMessageBox.confirm('Are you sure to close this dialog?')
    .then(() => {
      done()
    })
    .catch(() => {
      // catch error
    })
}
```

The existence of `beforeclose` is to solve this demand, and if you don't add this feature, `beforeclose` is useless, this is really too basic

and 

many feature your can learn from element-plus , it is full strong and beautiful api design

example , for Select component , value-by has not default value, every development need to design it ,it is too  cumbersome,
and named as `label` and `value` is very simple  to remember , I hope vuestic component‘s  prop name can same as `element-ui` or `antd`, personal opinion, no discussion

Valar103769:
resolve #2550

### DIFF
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -2137,6 +2137,10 @@
       "title": "Modal functional API",
       "text": "To make it easier to create modal we have a shorthands. For options API it is `$vaModal` that provides the VaModalPlugin (automatically installed with createVuestic). For composition Api we have `useModal` hook. Hook and global object have similar API."
     },
+    "beforeClose": {
+      "title": "Before close",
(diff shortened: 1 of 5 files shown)


m0ksem (approved):
Thanks!

Valar103769:
> Thanks! Feel silly about we haven't implemented this from beginning.

So humble, da lao

PR #2555, opened 2022-10-20
fix: broken icons on landing

asvae:
in-depth problem was that font was applied globally on typography.css include

### DIFF
--- a/packages/docs/src/assets/main.scss
+++ b/packages/docs/src/assets/main.scss
@@ -1,5 +1,5 @@
 @import "~vuestic-ui/src/styles/global/reset.scss";
-@import "~vuestic-ui/src/styles/index.scss";
+@import "~vuestic-ui/src/styles/resources";
 @import "./smart-grid.scss";
 @import "./colors.scss";
 @import "./fonts.scss";


m0ksem:
This issue cause because typography is imported in some component with scoped styles. 

I would better modify `src/assets/main.scss`

```scss
@import "~vuestic-ui/src/styles/global/reset.scss";
@import "~vuestic-ui/src/styles/resources";
@import "./smart-grid.scss";
@import "./colors.scss";
@import "./fonts.scss";
```

image

packages/ui/src/styles/typography/typography.scss
```diff
@@ -4,10 +4,6 @@
 // TODO: Move it to GLOBAL-CSS service, so it can be turned off and scoped.
 @include va-typography;
 
-* {
```
m0ksem, line 7: We need this in global typography styles, so it applies to all texts in p, etc. Maybe we don't need this, but it is useful when you just installed vuestic-ui in your project.

m0ksem (changes requested):

PR #2562, opened 2022-10-24
feat: update docs for the icons setup

### LINKED ISSUE #2387: Icon font import doesn't work
**Vuestic-ui version:** 1.4.10

**Steps to reproduce**
The docs say you can use <link or @import to use the material icon font.  The <link version works, but the @import does not:
```
    @import url('https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;1,700&display=swap');
    @import url('https://fonts.googleapis.com/icon?family=Material+Icons');

```
icons show up

no icons, just placeholder text like expand_more

**Other information**
Another part of the docs says:
```
    By default Vuestic UI uses material icons, so make sure to install the package:
    npm install material-design-icons-iconfont -D

```
This would seem to conflict with the other method, and this also does not get them working.

rustem-nasyrov:
Close #2387

## Description
- Added configuration example.
- Added translations.
- Updated docs.

## Types of changes


asvae (approved):
Thanks :).

PR #2583, opened 2022-10-27
fix: [VaHover] add flex:1

### LINKED ISSUE #2582: [VaHover] need flex:1
need add 
```css
.va-hover{    
    flex: 1;
    display: flex;
}
```

example: va-hover in 
```html
.va-dropdown.flex
 .va-hover.block
    .va-input.flex-1
```
if  .va-hover without  `flex:1;display:flex`,  can let `va-input`‘s `flex:1` invalid

```vue
<va-dropdown>
 <template #anchor>
      <va-hover stateful>
        <template #default="{ hover }">
          <va-input>
            <template #appendInner>
              <va-icon
                v-show="hover"
                name="clear"
                @click.stop="onClear()"
              ></va-icon>
            </template>
          </va-input>
        </template>
      </va-hover>
    </template>
</va-dropdown>
```

image

Valar103769:
to resolve #2582

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-hover/VaHover.vue
+++ b/packages/ui/src/components/va-hover/VaHover.vue
@@ -40,3 +40,10 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss">
+.va-hover {
+  dispaly: flex;
+  flex: 1;
+}
+</style>


packages/ui/src/components/va-hover/VaHover.vue
```diff
@@ -40,3 +40,10 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss">
+.va-hover {
+  dispaly: flex;
```
m0ksem, line 46: Typo here.

```suggestion
  display: flex;
```

m0ksem: I think you need to use `flex-grow` or `width: 100%` class in your example, if you want VaHover to fit your parent component.

For the most common example hover must flow its content:
```vue
<va-hover>
   <va-button>text</va-button>
</va-hover>
```

In current implementation, there is a bug, so `va-hover` is not flow `va-button` and hover is triggered outside of content.
firefox_T46ksVG3NC

m0ksem: So I would prefer fixing current:
image

to:
image

Which makes more sense.

Also, setting width: 100% to hover container child must fill VaHover
image

So it behaves more like headless component.

m0ksem: So my suggestion will be
```suggestion
  display: inline-flex;
  & > * {
     flex: 1;
  }
```

m0ksem (changes requested):

packages/ui/src/components/va-hover/VaHover.vue
Valar103769: VaHover, because there is an additional layer of dom structure, I think It's a doomed road 
 whether it is `dispaly:inline-flex`, or `dispaly:flex;flex:1`,
Neither is a good way, if you have to choose between the two, 
choose `dispaly:inline-flex`

Finally, I think there should be a non-component way to use hover state, like
1. Instructions
2. useElementHover

Considering implementing a v-directive? 
 you can  look at https://vueuse.org/core/useelementhover/

m0ksem: @Valar103769, agree with that.
I would like to see a headless component instead of current implementation. Until then, I would recommend using vueuse composable.

asvae:
Is this still relevant or unsolvable (can close)?

Issue #2565, opened 2022-10-25
Services refactoring


m0ksem:
For now, there is a mess in services folder. Need to clean it up and make it more readable.

- [ ] Make one structure for service
- [ ] Refactor vuestic plugins (Make it service?)
- [ ] Refactor utils
- [ ] Refactor defineVuesticPlugin, good idea with poor implementation

PR #2584, opened 2022-10-27
Refactoring/services

m0ksem:
I made +644 −897

closes https://github.com/epicmaxco/vuestic-ui/issues/2565


README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://vuestic.dev" target="_blank">
-      <img alt="Vuestic UI Logo" width="400" src="packages/ui/src/assets/vuestic-ui-logo.png">
+      <img alt="Vuestic UI Logo" width="400" src=".github/assets/vuestic-ui-logo.png">
```
asvae, line 3: Why? Not big deal, but not exclusive to github (used in docs as well). So we might want to put in docs.

m0ksem: We shouldn't store docs assets in UI package anyway. Since it is related to all packages, we can store it it .github, just because we have this folder, so we do not create new one.

m0ksem:
I'm gonna merge this to prevent suicide after new merge conflict.

PR #2590, opened 2022-10-28
Fix/element color from css variable

m0ksem:
Now useElementBackground uses not current background, but background set in css. Before, it was using *current* background, which in case with transition was changing slowly, so it was incorrect after transition animation is finished.


packages/ui/src/composables/useElementBackground/useElementBackground.ts
```diff
@@ -1,35 +1,35 @@
...
+    if (!element) { return '#fff' } // Likely doesn't have a color, so let's just return white
+    if (element.nodeType !== Node.ELEMENT_NODE) { return recursiveGetBackground(element.parentElement) }
+
+    const bg = getElementBackgroundFromCSS(element)
+
+    if (!bg) { return recursiveGetBackground(element.parentElement) }
+
+    if (isTransparent(bg)) {
+      const parentBg = recursiveGetBackground(element.parentElement)
+      return parentBg
```
rustem-nasyrov, line 27: Redundant variable.

rustem-nasyrov (changes requested):
Great job! But I'm facing errors in the console, when I navigate to `/ui-elements/alert`. 
<details>
<summary>Screenshot</summary>
image

</details>

m0ksem:
@rustem-nasyrov, take a look again, please. Noticed that getElementBackgroundFromCSS is extremely slow, so I added caching here as much as possible.

Issue #2115, opened 2022-07-19
Isolated components support


asvae:
Right now - if you want to use just a couple of vuestic components - you have to pull the whole thing.
Wondering if we can ship our components by themselves.

Check this article for Bit.
https://blog.bitsrc.io/6-ways-to-share-and-reuse-react-components-6d80e2fd16cd

- [ ] Research and play around to see if we can have something useful out of it.

Derranion:
**Update**:
We can use Vuestic [createVuesticEssential](https://vuestic.dev/en/getting-started/tree-shaking#base-layout) to get only needed components \ plugins.

We want to make it possible to install components as a separate package in `npm`, for example `npm install`:

@vuestic/select,
@vuestic/input,
@vuestic/slider

More info:
This is needed because it is really useful to get only one or a few components instead of the entire framework.

m0ksem:
Need to double check, likely web-components build fixes it.

PR #2591, opened 2022-10-29
feat/web components build

m0ksem:
Using Vuestic UI components in React project, create with vite.
firefox_7hkVFpVdS2

TODO: 
- [x] GlobalConfig support between components in shadow DOM without Vuestic Plugin (Might be also helpful for bit components.)
- [x] Docs with examples, limitations etc. (see https://github.com/epicmaxco/vuestic-ui/pull/2862)
- [x] Make similar API as `createVuestic` with `{ config, components, css }` options.
- [x] Test all components.
- [x] Would be nice to have build time global variables to exclude `createVuesticWebComponents` outside web-components build.
 > Decided not to do this, because we're not able to make different TS types for vuestic-ui and vuestic-ui/web-components.
- [x] If build time global variables will be made, we might have `createVuesticWebComponents` and `createVuesticWebComponentsEssential` for tree-shaking and non-tree-shaking build. (not needed, will be tree-shaked by default)
- [x] Stateful doesn't work
- [x] Provide/inject doesn't work
- [x] VaConfig doesn't work (this.$slots are empty) (see https://github.com/vuejs/core/issues/6142)
Might worth it to use sfc, with div.


packages/ui/build/plugins/web-components-nested-styles.ts
```diff
@@ -0,0 +1,67 @@
+import { Plugin } from 'vite'
+import { existsSync, readdirSync, readFileSync, writeFileSync, lstatSync } from 'fs'
+
+const getComponentsList = (text: string) => {
+  const declaration = text.match(/components: {(\w|\s|\n|,)*}/gm)
```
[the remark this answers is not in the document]
m0ksem: This is actually cool, but I don't really understand how to deal with it.

m0ksem:
Blocked by https://github.com/epicmaxco/vuestic-ui/pull/2584. I don't want to make changes in Global Config service until refactoring PR is merged.

packages/ui/build/plugins/web-components-nested-styles.ts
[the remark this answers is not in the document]
asvae: maybe we can ignore this line?

m0ksem:
For tests visit packages/sandbox/web-components - clean new vanila vite project which uses web-components. Maybe we need to make extended demo.

Also, in docs I'll mark web-components as beta, because we haven't tested them super good for now. I'm gonna let it for users or we can come back to it later. For now, this PR just makes cool fact that VuesticUI can be used as Web Components.

aluarius:
Seems for me to be already tested in https://github.com/epicmaxco/vuestic-ui/pull/2862

aluarius:
Is there any way to use Toast as web component, for example?

aluarius:
Smthng strange for VaCollapse.

image

aluarius:
Also for VaDateInput:

image

Seems to be smthng wrong with VaDropdown in web components.

aluarius:
VaRating not rendering at all.

image

aluarius:
Weird slider style.

image

aluarius:
All in all, we need a demo in sandbox with all components presented.

m0ksem:
@aluarius, I made demos for the most used components (skipped demos that were hard to implement). There is a small implementation of Vue Script Setup syntax, but for native rendering (but without reacitivy support). Shouldn't be hard to figure out how it works.

- Bugs from above were fixed by tweaking our compiler.
- Fixed bugs caused by missing reset.css (might be scope of https://github.com/epicmaxco/vuestic-ui/issues/2115)

The last weird issue is with Virtual Scroller, which I suggest removing from Web Components usage (until  https://github.com/epicmaxco/vuestic-ui/issues/2866)

packages/ui/src/components/va-config/hooks/useLocalConfig.ts
```diff
@@ -0,0 +1,17 @@
+import { inject, provide, computed, ComputedRef } from 'vue'
+import { ComponentConfig } from '../../../services/component-config'
+
+/**
+ * We need another key to provide injected value.
+ */
+export const LocalConfigKey = 'VaLocalConfig'
```
rustem-nasyrov, line 7: Why not `Symbol`?

packages/ui/src/services/web-components/register-vuestic-web-components.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export const registerVuesticWebComponents = (options: {
+  /**
+   * In case you need pass some css to shadow-dom of Custom Element
+   *
+   * @example
+   * For Material Icons support inside Custom Element we need to pass `.material-icons` class.
+   *
+   * @default
+   * We do this by default, so you don't need to copy past it. In case you want to use different icon
```
rustem-nasyrov, line 34:
```suggestion
   * We do this by default, so you don't need to copy paste it. In case you want to use different icon
```

packages/ui/src/components/va-config/hooks/useLocalConfig.ts
m0ksem: Web compoent's provide doesn't support Symbol.

Issue #2469, opened 2022-10-07
Make hide-arrows prop in va-tabs


m0ksem:
@m0ksem, @asvae while I was working on this issue, and used the va-tabs component, I had thoughts about to add show/hide pagination property. Do we need it?

_Originally posted by @rustem-nasyrov in https://github.com/epicmaxco/vuestic-ui/issues/2466#issuecomment-1270197961_

m0ksem:
![image](https://user-images.githubusercontent.com/23530004/194617120-ac6e1d87-74a8-425c-97f6-a28c550d3f2b.png)

m0ksem:
How about making this buttons flat? 
![image](https://user-images.githubusercontent.com/23530004/195718126-8ecc9ef7-a261-4ab1-8cb5-6045e0e2f2d0.png)

xiongmao86:
Does merging #2605 meaning this should be closed?

rustem-nasyrov:
@xiongmao86 yes, it should. Thank you that you mentioned it. :)

PR #2600, opened 2022-11-01
feat(va-tabs): support hidden pagination

Valar103769:
closes https://github.com/epicmaxco/vuestic-ui/issues/2469

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-tabs/VaTabs.vue
+++ b/packages/ui/src/components/va-tabs/VaTabs.vue
@@ -28,6 +28,7 @@
         <div
           ref="tabs"
           class="va-tabs__tabs"
+          :class="paginationControlledClass"
           :style="paginationControlledStyles"
         >
           <div
@@ -106,6 +107,7 @@ export default defineComponent({
     right: { type: Boolean, default: false },
     center: { type: Boolean, default: false },
     grow: { type: Boolean, default: false },
+    withPagination: { type: Boolean, default: true },
     disabled: { type: Boolean, default: false },
     hideSlider: { type: Boolean, default: false },
     vertical: { type: Boolean, default: false },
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-tabs/VaTabs.vue
```diff
@@ -106,6 +107,7 @@ export default defineComponent({
     right: { type: Boolean, default: false },
     center: { type: Boolean, default: false },
     grow: { type: Boolean, default: false },
+    withPagination: { type: Boolean, default: true },
```
m0ksem, line 110: Better rename it to `hide-pagination` prop, so it is `false` by default.

```vue
<va-tabs hide-pagination />
```
vs

```vue
<va-tabs :with-pagination="false" />
```

packages/ui/src/components/va-tabs/VaTabs.vue
```diff
@@ -472,6 +481,10 @@ export default defineComponent({
     .va-tabs__tabs {
       position: absolute;
       height: 100%;
+
```
m0ksem, line 484: Might be easier to use `v-if="showPagination && !$props.hidePagination"` on buttons in component template.

m0ksem (changes requested):
Thanks. Looks good, but it can be a bit simpler.

packages/ui/src/components/va-tabs/VaTabs.vue
Valar103769: new pr: https://github.com/epicmaxco/vuestic-ui/pull/2605
 ```css
 .va-tabs__tabs {
      position: absolute;
}
```
`position:absolute` need to be remove, let width grow by content

PR #2610, opened 2022-11-02
Fix/nuxt 1.5.0

m0ksem:
- Fix dev:prepare build
 - Added missing components
 - Added missing composables

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,5 +40,7 @@
   "gitHooks": {
     "pre-commit": "lerna run --concurrency 1 --stream precommit --since HEAD"
   },
-  "dependencies": {}
+  "resolutions": {
+    "vue": "3.2.37"
+  }
 }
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -40,5 +40,7 @@
   "gitHooks": {
     "pre-commit": "lerna run --concurrency 1 --stream precommit --since HEAD"
   },
-  "dependencies": {}
+  "resolutions": {
```
m0ksem, line 43: This makes vue to be 3.2.37 in every package we have.

asvae, line 43: That's pretty weird, why are we bound to specific version?

m0ksem: Because there is types issues between two diferent versions. We're not able to use createVuesticPlugin from packages/ui (where is vue resolved as 3.2.37) in project like packages/sandbox or pacakges/nuxt where is vue resolved as 3.2.38 - it will throw type mismatch error.

m0ksem (commented):
We 100% need all of this stuff, at least for nuxt docs.

PR #2643, opened 2022-11-07
fix(button): word-break:keep-all

Valar103769:
## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-button/VaButton.vue
+++ b/packages/ui/src/components/va-button/VaButton.vue
@@ -216,6 +216,7 @@ export default defineComponent({
   transition: var(--va-button-transition);
   box-sizing: border-box;
   cursor: var(--va-button-cursor);
+  word-break: keep-all;
 
   &::after,
   &::before {


Valar103769:
maybe white-space:nowrap

asvae:
What is this trying to solve? It's pretty common for buttons to have multi-line text :).

Valar103769:
[antd](https://ant.design/components/button-cn/)
image


[element-plus](https://element-plus.org/zh-CN/component/button.html#%E5%9F%BA%E7%A1%80%E7%94%A8%E6%B3%95)

image


in my case , my text  is only two chinese word , but it is  vertical shown, i need to let it horizontion  by manual add `white-space:nowrap` 
It doesn't matter if you don't add it.

asvae:
Aha, interesting. I checked vuetify and quasar, and they have wrap enabled.

I think let's keep wrap, as disabling it would introduce a breaking change, plus, both are mostly matters of taste. :)

PR #2654, opened 2022-11-09
VaDataTable fixes

### LINKED ISSUE #2653: [va-data-table] can't use component at all
**Vuestic-ui version:** 1.5.0

**Steps to reproduce**

1. create a fresh vueJs project (`npm init vue@latest`)
2. install vuestic-ui (`vue add vuestic-ui`)
3. in `App.vue` add `import {VaDataTable} from 'vuestic-ui'` and then `<VaDataTable></VaDataTable>` component in the template

See a working data table

Chrome Inspector returns 2 errors:
* `useIntersectionObserver.ts:30 Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`
* `VaDataTable.vue:524 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'selectedColor')`

rustem-nasyrov:
Close #2653

## Description
- Fixed `v-bind()` usage in styles.
- Fixed intersection observer for the table and its initialization.
- Other little fixes.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
+++ b/packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
@@ -1,4 +1,4 @@
-import { getCurrentInstance, computed, onMounted, ExtractPropTypes } from 'vue'
+import { getCurrentInstance, computed, ExtractPropTypes } from 'vue'
 
 import { useIntersectionObserver, useElementRef } from '../../../composables'
 
@@ -51,9 +51,7 @@ export default function useTableScroll (
     rootMargin: `${props.scrollTopMargin ?? 0}px 0px ${props.scrollBottomMargin ?? 0}px 0px`,
   }))
 
-  onMounted(() => {
-    useIntersectionObserver(intersectionHandler, options, targets)
-  })
+  useIntersectionObserver(intersectionHandler, options, targets)
 
   return {
     scrollContainer,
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/va-data-table/hooks/useTableScroll.ts
```diff
@@ -51,7 +51,9 @@ export default function useTableScroll (
     rootMargin: `${props.scrollTopMargin ?? 0}px 0px ${props.scrollBottomMargin ?? 0}px 0px`,
   }))
 
-  useIntersectionObserver(intersectionHandler, options, targets)
+  onMounted(() => {
+    useIntersectionObserver(intersectionHandler, options, targets)
```
m0ksem, line 55: This is wrong. All composables must be used in scope of setup function, not mounted.

You will cause making things on each mount, which potentially can cause memory leaks. Especially with intersection observer, which must call disconnect method like destruxtor

packages/ui/src/composables/useEvent.ts
```diff
@@ -28,12 +28,12 @@ export const useEvent = <N extends UseEventEventName, E extends Event>(
 
   watch(source, (newValue, oldValue) => {
     if (!Array.isArray(event)) {
-      unref(newValue)?.addEventListener(event, listener as any, capture)
-      unref(oldValue)?.removeEventListener(event, listener as any, capture)
+      unref(newValue)?.addEventListener?.(event, listener as any, capture)
```
m0ksem, line 31: How is that possible for html element to not have addEventListener?

If this is vnode, you might need useHTMLElement to do vnode.$el

m0ksem (changes requested):

packages/ui/src/composables/useEvent.ts
rustem-nasyrov: Because of this: 
image

PR #2658, opened 2022-11-10
Update VaSwitch.vue

Valar103769:
## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes


rustem-nasyrov:
Thank you for your contribution, but this is not how it can be fixed. I'll suggest you to use slots, and inside of this slots use your custom styles. For example:

```html
<va-switch>
  <template #innerLabel>
    <span style="white-space: nowrap;">
      Inner Label slot
    </span>
  </template>
</va-switch>
```

PR #2672, opened 2022-11-11
Fix sorting in the VaDataTable

### LINKED ISSUE #2497: VaDataTable doesn't work on SSR
**Vuestic-ui version:** 1.5.0

image

rustem-nasyrov:
Close #2497

## Description
- Refactored and fixed with it sorting issue on nuxt when `ssr` is truthy.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-data-table/hooks/useSortable.ts
+++ b/packages/ui/src/components/va-data-table/hooks/useSortable.ts
@@ -71,23 +71,24 @@ export default function useSortable (
     }
 
     const column = columns.value.find(column => column.name === sortBySync.value)
+    const columnIndex = column
+      ? columns.value.findIndex(({ name, sortable }) => sortBySync.value === name && sortable)
+      : -1
 
-    if (!column || !column.sortable) {
+    if (!column || !column.sortable || columnIndex === -1) {
       return filteredRows.value
     }
 
-    const columnIndex = columns.value.indexOf(column)
+    const sortingOrderRatio = sortingOrderSync.value === 'desc' ? -1 : 1
 
     return [...filteredRows.value].sort((a, b) => {
-      const firstValue = a.cells[columnIndex].value
-      const secondValue = b.cells[columnIndex].value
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -71,23 +71,24 @@ export default function useSortable (
...
+      ? columns.value.findIndex(({ name, sortable }) => sortBySync.value === name && sortable)
+      : -1
 
-    if (!column || !column.sortable) {
+    if (!column || !column.sortable || columnIndex === -1) {
       return filteredRows.value
     }
 
-    const columnIndex = columns.value.indexOf(column)
+    const sortingOrderRatio = sortingOrderSync.value === 'desc' ? -1 : 1
```
m0ksem, line 82: Is this move `sortingOrderRatio ` is important too? 🤔

packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -71,23 +71,24 @@ export default function useSortable (
     }
 
     const column = columns.value.find(column => column.name === sortBySync.value)
+    const columnIndex = column
+      ? columns.value.findIndex(({ name, sortable }) => sortBySync.value === name && sortable)
+      : -1
```
m0ksem, line 76:
```suggestion
      : -1
    const column = columns.value[columnIndex]
```

packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -71,23 +71,24 @@ export default function useSortable (
...
 
-    if (!column || !column.sortable) {
+    if (!column || !column.sortable || columnIndex === -1) {
```
m0ksem, line 78: Double check for column.sortable?

m0ksem (commented):
This looks very the same from what it was before in my opinion? Can you explain what exactly is changed here?

PR #2699, opened 2022-11-21
Refactor/color indicator

### LINKED ISSUE #2695: VaColorIndicator redesign
- [ ] add 1px elementBorder outline only for colours with contrast lower than 1, in the newest example it will be white only. other elements shouldn’t have any borders.
- [x] On hover indicator should stay at the same size.
- [x] Add `focus` outline.

image

rustem-nasyrov:
Close #2695

image

## Description
- [ ] add 1px elementBorder outline only for colours with contrast lower than 1, in the newest example it will be white only. other elements shouldn’t have any borders.
- [x] On hover indicator should stay at the same size.
- [x] Add focus outline.

## Types of changes


Azgiliat (approved):

packages/ui/src/components/va-color-indicator/VaColorIndicator.vue
```diff
@@ -79,17 +79,14 @@ export default defineComponent({
...
   &__core {
-    transition: transform 0.1s linear;
-    border-radius: 50%;
-    width: 1rem;
+    border-radius: v-bind(borderRadiusComputed);
     height: 1rem;
+    width: 1rem;
+  }
+
+  &:focus {
```
aluarius, line 88: Not `keyboard-only-focus` mixin? Not sure if we want outline for mouse interactions.

aluarius (approved):

Issue #1287, opened 2021-11-30
Strange grid logic


m0ksem:
We have two classes:

d-flex
flex

md6 or sm12 classes don't work without flex class. 

flex class makes nothing by it self. 

It can be confused and also break some frameworks like tailwind.

asvae:
Tailwind shouldn't be an issue, as if you use tailwind helpers you won'd use vuestic helpers.

m0ksem:
> Tailwind shouldn't be an issue, as if you use tailwind helpers you won'd use vuestic helpers.

Yes, since we have tree-shakable CSS user can remove our grid from bundle. Anyway, these two classes can make smby confused.

asvae:
Moved to another issue - we plan to rework CSS helpers

asvae:
Let's handle this after investigation is done in scope of #1222

Issue #731, opened 2021-05-12
CSS helpers docs are not too nice


asvae:
CSS helpers docs are not too nice. We want to improve that.

Issue #2184, opened 2022-07-29
Tailwind transition overthinking


asvae:
We decided to have 1st class tailwind support, so we'd scrape some of our helpers, and just link tailwind solutions instead. That's not an easy process so let's think it over.

* Typography - there is a typography plugin for tailwind (https://play.tailwindcss.com/uj1vGACRJA?layout=preview), also typography doesn't go out of the box for tailwind. So we probably can keep our solution and it's overall a good fit
* [Colors](https://tailwindcss.com/docs/customizing-colors) - not exactly clear on implementation, but we're probably looking to integrate tailwind config with vuestic config.
* [Breakpoints](https://tailwindcss.com/docs/screens) - we're looking for some kind of integration with tailwind solution. As we can't rely completely on tailwind here.
* Grid - We should separate smart (vuestic) grid classes and tailwind. Both could be used at once though.
* [Dark theme](https://tailwindcss.com/docs/dark-mode) - again, would be nice to have a way to combine both, but not sure if that's possible.
* [Reset](https://tailwindcss.com/docs/preflight) could also become a bit of an issue. (maybe not - if we don't use it for vuestic components at all)

----

* Might make sense to introduce **vuestic-tailwind-plugin** to keep all tailwind integration in one place. Or maybe as an optional service?

### Relevant issues
These should ideally be handled in context of this issue.
* #665
* #731
* #764

asvae:
We decided to use tailwind and not to use tailwind for vuestic component.

![image](https://user-images.githubusercontent.com/5394573/209569471-f94edf05-3a3c-4369-ab0e-8c0e30c81b9f.png)

asvae:
![image](https://user-images.githubusercontent.com/5394573/209572522-7f0cb603-be36-4307-b669-7e0996e11a2a.png)

m0ksem:
@asvae, @aluarius done?

aluarius:
Imo yes, can't see anything else to do here.

PR #2745, opened 2022-12-01
Tailwind CSS integration package

aluarius:
Related: #2184
Closes: #1287
Closes: #731

## Description
- [x] New `@vuestic/tailwind` package for syncing Tailwind CSS -> Vuestic UI colors and breakpoint configs.
- [x] Deleted page about `reset.css` from docs.
- [x] `grid` docs page was totally rewritten .

While testing i've found out, that our color config does't understand colors with separators (red-500 is broken, but red500 is ok for example). Will investigate and fix it within another issue.

## Types of changes


packages/tailwind/modules/breakpoint.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @returns vuestic ui thresholds settings
+ */
+export const convertThresholds = (thresholds: Record<string, string>): VuesticThresholdsList => {
+  return Object.fromEntries(
+    Object.entries(thresholds)
+      .map(([key, value]) => {
+        const vuesticValue = Number(value.substring(0, value.length - 2))
+        return [key, vuesticValue]
+      }),
+  )
```
rustem-nasyrov, line 14: Also is there any chance to get `NaN`?
```suggestion
  return Object.entries(thresholds).reduce((acc, [key, value]) => {
    acc[key] =  Number(value.substring(0, value.length - 2)) // what if NaN?
    return acc
  }, {} as VuesticThresholdsList)
```

packages/tailwind/modules/breakpoint.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @param tailwindConfig users or default tailwind config
+ * @returns vuestic breakpoint config
+ */
+export const proceedBreakpoint = (tailwindConfig?: Record<string, any>): VuesticThresholdsList => {
+  let tailwindThresholds
+
+  if (tailwindConfig) {
+    const tailwindThresholdsConfig = tailwindConfig.theme?.screens || tailwindConfig.theme?.extend?.screens
```
rustem-nasyrov, line 25: Is it possible to have in tailwind config both of this?

packages/tailwind/modules/breakpoint.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (tailwindThresholdsConfig) { tailwindThresholds = convertThresholds(tailwindThresholdsConfig) }
+  }
+
+  const isTailwindScreensConfigOverwritten = !!Object.keys(tailwindConfig?.theme?.screens || {}).length
+  const defaultTailwindConfig = require('tailwindcss/defaultTheme')
+  const tailwindDefaultScreensConverted = convertThresholds(defaultTailwindConfig.screens)
```
rustem-nasyrov, line 31: What if we can't access to the default config? Is it possible?

packages/tailwind/modules/colors.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param colors tailwind css colors options
+ * @returns vuestic ui colors options
+ */
+export const convertColors = (colors: Record<string, string | Record<string, string>>): VuesticColorsList => {
+  // system and deprecated by tailwind
+  const ignoredColors = ['inherit', 'current', 'transparent', 'coolGray', 'blueGray', 'trueGray', 'warmGray', 'lightBlue']
+  const result = {} as VuesticColorsList
+
+  for (const prop in colors) {
+    if (!ignoredColors.includes(prop)) {
```
rustem-nasyrov, line 13: Can be this filtered before the `for (const prop in colors) {}` loop?

rustem-nasyrov (changes requested):

packages/tailwind/helpers/helpers.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @param value any value for converting it to raw string
+ * @param nestingLevel current nesting level (is needed for correct indents)
+ * @returns raw string value
+ */
+export const convertValueToString = (value: any, nestingLevel = 1): string => {
```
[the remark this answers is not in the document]
aluarius: Yeap, i've made it universal (thinking about further reusing).

In case of Tailwind integration, there will be values: object, string, number.

packages/tailwind/modules/breakpoint.ts
aluarius: > Also is there any chance to get NaN?

 Only in case of typo. But will add a check just in case.

packages/tailwind/modules/breakpoint.ts
aluarius: No, it's taken from node_modules -> Tailwind CSS (which is a dependency of the package).

packages/tailwind/modules/breakpoint.ts
aluarius: Yes, but this makes no sense - they're interchangeable (overwriting or extending default config -> if it's overwritten, we couldn't just extend).

packages/tailwind/modules/colors.ts
aluarius: I've tried:

image

But in this case (probably because of another way of accessing values) we gain Tailwind deprecated warnings:

image

We still can implement it this way:

image

But it's much more complicated than initial variant. So, let's it leave untouched :)

rustem-nasyrov (approved):

packages/docs/src/locales/en/en.json
```diff
@@ -2981,86 +2980,20 @@
     }
   },
   "grid": {
-    "title": "Grid System",
```
asvae, line 2984: Maybe we can move this to separate page and add a tag in navigation menu and deprecation warning?

That's a breaking change and it's ideal if we keep backwards compatibility even within docs.

packages/docs/src/locales/en/en.json
```diff
@@ -2981,86 +2980,20 @@
...
-      },
-      "justifySelf": {
-        "title": "`justify-self`:",
-        "variants": [
-          "`.justify-self-start` - The item is packed flush to each other toward the start edge of the alignment container in the appropriate axis.",
-          "`.justify-self-end` - The item is packed flush to each other toward the end edge of the alignment container in the appropriate axis.",
-          "`.justify-self-center` - For items that are not children of a flex container, this value is treated as `.justify-self-start`.",
-          "`.justify-self-baseline` - The items are aligned according to baselines"
+    "title": "Grid Layout",
+    "summaryText": "We decided to abandon the concept of creating our own CSS helpers (with the exception of [smart helpers](/styles/smart-helpers)[[target=_blank]]). Therefore, now we recommend our users to use [Tailwind CSS](https://tailwindcss.com/)[[target=_blank]] as the most convenient and modern solution to the issue.",
```
asvae, line 2984: That's historic data and not very relevant to new users.

Let's keep deprecated page with old styles (which we'll delete for 1.7.0) and page with tailwind integration separate.

Tailwind page - should only talk about tailwind and which role it fits in current state of things.

packages/docs/src/locales/en/en.json
```diff
@@ -3207,16 +3140,6 @@
     },
     "variables": "[GitHub Variables Page](https://github.com/epicmaxco/vuestic-ui/tree/develop/packages/ui/src/components/va-backtop/_variables.scss)[[target=_blank]]"
   },
-  "reset": {
```
asvae, line 3210: Same with reset, let's keep it, but mark as deprecated.

packages/tailwind/helpers/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
```
asvae, line 6: Just curious, but why not `JSON.stringify`? :)

packages/tailwind/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "vuesticui",
+    "vuestic",
+    "tailwindcss",
+    "tailwind",
+    "uikit",
+    "uiframework"
+  ],
+  "version": "1.0.0",
+  "license": "MIT",
+  "main": "./dist/sync.cjs",
```
asvae, line 14: Small thing but if you use `tsx` to run script - it makes sense to skip build step and run `.ts` file directly.
If you compile file - you can use `npx`.

packages/tailwind/helpers/helpers.ts
aluarius: I've tried it first of all, but it makes not exactly what we need (or we will be forced to ask our users to make JSON.parse and that's weird).

packages/tailwind/package.json
aluarius: Using `npx` can be problematic here:

image

 I'd prefer to build it, because can't see any benefits of the other option.

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -91,6 +91,13 @@ export const navigationRoutes: NavigationRoute[] = [
     },
```
asvae, line 91: This command didn't work for me: `node ./node_modules/@vuestic/tailwind/dist/sync.cjs`.

Also in docs we use tsx, which users might not have.

image

packages/docs/src/locales/en/en.json
```diff
@@ -3065,6 +3067,25 @@
       }
     }
   },
+  "tailwind": {
```
asvae, line 3070: We need to make docs a bit cleaner, it's very hard to understand right now.

What I would suggest:

Have 2 installation flows explained:
1) You install VuesticUI on top of Tailwind
2) You install Tailwind on top of VuesticUI

Then separately talk about syncing.

Right now it's not clear what steps have to be taken (is tailwind dependency included in plugin, is vuestic included in plugin etc).

packages/tailwind/package.json
asvae: using node directly doesn't work either? why? o.o

packages/docs/src/components/sidebar/navigationRoutes.ts
```diff
@@ -91,6 +91,13 @@ export const navigationRoutes: NavigationRoute[] = [
...
     disabled: true,
```
asvae, line 92: Let's test performance impact of syncing all css variables from tailwind.

If it's a problem - let's add a solution for users to fix that.

packages/docs/src/components/sidebar/navigationRoutes.ts
aluarius: Fixed. Wasn't proceeded situation if no tailwind.config file was found. 

Script reworked to work via `node`.

packages/docs/src/components/sidebar/navigationRoutes.ts
aluarius: I haven't faced any performance issues, except lots of css variables in the code.

packages/docs/src/locales/en/en.json
aluarius: Docs updates.

packages/tailwind/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "vite build"
+  },
+  "peerDependencies": {
+    "tailwindcss": "^3.2.4"
```
asvae, line 22: We can probably add vuestic here as well.

packages/docs/src/page-configs/styles/tailwind/code-examples/configuring.ts
```diff
@@ -0,0 +1,61 @@
+export const configuringVuesticFull = `
+// main.*
+
+import { createVuestic } from 'vuestic-ui'
+import { config } from '../vuestic.config.ts'
```
asvae, line 5: JS file has been generated for me.

packages/tailwind/watch-tailwind.ts
```diff
@@ -0,0 +1,13 @@
+import { tailwindConfigFilename } from "./helpers/helpers";
```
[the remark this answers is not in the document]
aluarius: If we add `vuestic-ui` to the peer dependencies, no more error will be here.

image

packages/tailwind/package.json
aluarius: This makes vite to use `cjs`, better to use monorepo vuestic-ui in build.

packages/docs/src/locales/en/en.json
```diff
@@ -3152,6 +3154,35 @@
...
+      "integration": {
+        "title": "Syncing configs",
+        "text": "To create a tight bind between Vuestic with Tailwind we have a `@vuestic/tailwind` package: it syncs [breakpoint](/services/breakpoints)[[target=_blank]] and [color](/services/colors)[[target=_blank]] settings. You can sync from Vuestic UI to Tailwind or from Tailwind to Vuestic UI!",
+        "instructionIntro": "Here's what you have to do to make it work:",
+        "instruction": [
+          "1.&nbsp;Install this package:",
+          "2.&nbsp;After installation, three commands become available to you:",
+          "`npx sync-tailwind` - formats and transfers the color and breakpoints settings from the `tailwind.config.*` file (and in its absence, it uses the default Tailwind CSS configuration) to the root file `vuestic.config.js` (also added by this command);",
+          "`npx watch-tailwind` - watches `tailwind.config.*` in background and synchronizes the Vuestic UI configuration on the fly;",
+          "`npx sync-vuestic` - transfers color and breakpoints settings from default Vuestic UI config to the `tailwind.config.cjs`.",
```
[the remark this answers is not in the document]
aluarius: It's `cjs` by default.

packages/docs/src/locales/en/en.json
```diff
@@ -3152,6 +3154,35 @@
...
+          "3A.&nbsp;`[Tailwind CSS -> Vuestic UI]` import the `config` configuration in your `main.*` and specify it when initializing the application:",
+          "Also, you may want to synchronize the configuration partially (in the example - only Tailwind CSS colors will be taken):",
+          "3B.&nbsp;`[Vuestic UI -> Tailwind CSS]` Nothing more is needed.",
+          "4A.&nbsp;`[Tailwind CSS -> Vuestic UI]` As a result you will be able to use Tailwind colors and breakpoints setting in Vuestic. For example, the result of synchronization default Tailwind settings will be a Vuestic settings file:",
+          "4B.&nbsp;`[Vuestic UI -> Tailwind CSS]` Vuestic UI setting will be available in the Tailwind CSS scope. Output format follows the requirements, located [here](https://tailwindcss.com/docs/screens#adding-larger-breakpoints)[[target=_blank]] and [here](https://tailwindcss.com/docs/screens#adding-larger-breakpoints)[[target=_blank]].",
```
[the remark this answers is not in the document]
aluarius: It's not steps, but options. Suggested scheme isn't suitable here.

4. summary
4.1. then
4.2. then

4. common
4A. or
4B. or

m0ksem (approved):
I don't like bin commands name. `sync-tailwind` what does it mean? Let's imagine I'm new developer on project, how should I know what this command does? I would rename it to `sync-tailwind-with-vuestic`. or "vuestic sync-with-tailwind" "vuestic sync-to-tailwind". No winner yet. You can keep it as is. I hope there will be a plugin later.

aluarius:
> Let's imagine I'm new developer on project, how should I know what this command does?

Open docs? :)

But i see your point, will rename them.

Issue #2741, opened 2022-11-29
nuxt shows warning message


asvae:
**Vuestic-ui version:** 1.5.1

> From email

I have an issue with vuestic where when I add it with nuxt, it shows a warning message. Here is the message.
 
[Vue warn]: onMounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement.

rain-hsiang:
+1  in nuxt3

dongke404:
problem+1 in nuxt3

PR #2757, opened 2022-12-02
[#2741] Nuxt 3 hook warning fix

aluarius:
Closes: #2741 

## Description
- [x] found out the place of the problem (that was really exhausting -__-),
- [x] `onMounted` hook was changed to watcher,
- [x] minor related refactoring. 

Tested and seems to work as needed:

image

image

But better to take a look twice because it's a pretty fundamental thing.

## Types of changes


rustem-nasyrov (approved):

m0ksem (approved):

asvae:
Tested, seems to work with empty setup, but nuxt is pretty messed up right now.

We'd need to fix that after test setup from anton.

image

PR #2764, opened 2022-12-05
VaNavbar - default slot, properlyalign slots, Hooks - useDeprecated

### LINKED ISSUE #2759: Make navbar default slot
Also, right slot isn't at right if there is not left slot.

image

And why do we need va-navbar__content?

Middle isn't aligned at middle, actually.

rustem-nasyrov:
Close #2759

## Description
- Added default slot the center. Left two slots. **Feedback needed**.
- Fixed alignment for the slots.
- Refactored and replaced `useDeprecatedProps` with `useDeprecated` hook.
- Added slots description and translations.

image

## Types of changes


packages/ui/src/components/va-navbar/VaNavbar.vue
```diff
@@ -173,9 +165,15 @@ export default defineComponent({
   }
 
   @include media-breakpoint-down(sm) {
-    height: top-mobile-nav-height;
+    flex-direction: column;
+    align-items: center;
+    height: $top-mobile-nav-height;
```
aluarius, line 170: We've decided not to use global variables, because we're planning to get rid of them over time.

aluarius (approved):

packages/ui/src/components/va-navbar/VaNavbar.demo.vue
```diff
@@ -19,7 +19,35 @@
       </va-navbar>
     </VbCard>
 
-    <VbCard style="width: 100%;" title="Custom color">
+    <VbCard title="Slots" style="width: 100%;">
+      <va-navbar color="primary">
+        <template #left>
```
asvae, line 24: Would be nice to have a demo with combination of different slots (rl, rc, lc, rcl)

packages/ui/src/components/va-navbar/VaNavbar.vue
```diff
@@ -4,19 +4,19 @@
...
-      </div>
-
-      <div v-if="$slots.right" class="va-navbar__right">
-        <slot name="right" />
-      </div>
+    <div v-if="$slots.left" class="va-navbar__left">
+      <slot name="left" />
     </div>
+
+    <div v-if="$slots.center || $slots.default" class="va-navbar__center">
```
asvae, line 11: Do we need both center and default to prevent breaking change? If yes - we should follow with deprecation.

packages/ui/src/components/va-navbar/VaNavbar.vue
```diff
@@ -74,38 +74,46 @@ export default defineComponent({
 @import "variables";
 
 .va-navbar {
+  display: grid;
+  grid-template: "left center right" / 1fr 1fr 1fr;
```
asvae, line 78: We should test that with different content sizes. Seems too rigid to me.

packages/ui/src/components/va-navbar/VaNavbar.demo.vue
rustem-nasyrov: Only for the demo or  for the docs too?

packages/ui/src/components/va-navbar/VaNavbar.vue
rustem-nasyrov: I've checked all browsers and common resolutions, looks fine.

aluarius (approved):

Issue #2747, opened 2022-12-01
`VaCollaplse` rounded borders bug


aluarius:
If bg color is set for collapse content block, its design becomes broken.

<img width="876" alt="image" src="https://user-images.githubusercontent.com/64714442/205050322-0779f304-8fff-4457-ba93-4bcca1624be4.png">

Issue #2576, opened 2022-10-26
More flexible VaCollapse


asvae:
Here's the case:
![image](https://user-images.githubusercontent.com/5394573/198048391-8e31e3c8-adb5-48bf-9e9d-fe7a6edb4cfd.png)
<img width="585" alt="image" src="https://user-images.githubusercontent.com/5394573/198048445-23abbcf6-2949-45d0-9714-9560bade29a5.png">
<img width="436" alt="image" src="https://user-images.githubusercontent.com/5394573/198051225-99da4a98-f0a0-4c16-bb6c-6f5728e4b9a9.png">

I want to have bold color, but if I pass slot - it destroys collapse styling.
- [ ] We probably want a more flexible slot structure for collapse to only slot in header text instead of whole header (`header-text`)
- [ ] Additionally we'd want maybe to have bold by default, normal text is somewhat not too visible.

asvae:
@xx13 do you think bold is a good idea here?

PR #1640, opened 2022-04-08
[#1338] Fix `Va-Modal` docs

RVitaly1978:
## Description
close #1338

changes:
- [x] remove `slot` tag from templates in the docs and demo
- [x] tweak locales to avoid **intlify** warnings
- [x] add `inheritAttrs: false` to `ModalElement` component (wrapper)

image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

PR #2774, opened 2022-12-06
`VaCollapse` component minor updates

aluarius:
Closes: #2576
Closes: #2747
Closes: #1640

## Description
Updates to `VaCollapse` related to the issues above: 
- [x] border radius for expanded state fix;
- [x] added header content slot;
- [x] documentation updates;
- [x] added menu example to VaAccordion docs page:

image

## Types of changes


packages/docs/src/page-configs/ui-elements/accordion/examples/Menu.vue
```diff
@@ -0,0 +1,75 @@
+<template>
```
rustem-nasyrov, line 1: I think with this example, we might mislead users with the `VaSidebar` component. Because we are duplicating `VaSidebar` functionality.

rustem-nasyrov (approved):

packages/docs/src/page-configs/ui-elements/accordion/examples/Menu.vue
aluarius: Current docs `VaSidebar` also uses `VaAccordion`.

Issue #2663, opened 2022-11-10
VaAvatar improvements


asvae:
Features we want:
- [x] fallback to text and image
- [x] somehow avatar group
- [ ]  gravatar integration example (shouldn't be part of component api)
- [x] add example with badge

Examples we want:
- [ ] we need to tell that you can globally set a fallback for avatar component (as users might not be able to figure that out on themselves)

### Reference
* https://www.naiveui.com/en-US/os-theme/components/avatar#fallback.vue

rustem-nasyrov:
I would like to discuss about `fallback` property implementation.
- The `fallback` property should have a string type.
- If we prepend `image | icon | text` text to it, then component should render fallback with given type. For example:
```html
<va-avatar src="path/not/exists" fallback="icon:profile" />
<va-avatar src="path/not/exists" fallback="image:path/to/fallback-image" />
<va-avatar src="path/not/exists" fallback="text:Not found" />
```
Or instead of creating as property, we can implement it as custom vue directive:
```html
<va-avatar src="path/not/exists" va-fallback.icon="profile" />
<va-avatar src="path/not/exists" va-fallback.image="path/to/fallback-image" />
<va-avatar src="path/not/exists" va-fallback.text="Not found" />
```

m0ksem:
Let's have a docs example with gravatar.

PR #2793, opened 2022-12-14
Feat/va avatar

### LINKED ISSUE #2396: va-avatar shouldn't use color if src is provided and image exists
**Vuestic-ui version:** 1.4.10
image

Probably best would be to check refs on that. Not sure what's the perfect approach.
image

### LINKED ISSUE #2767: Avatar with slot and loading state
**Vuestic-ui version:** 1.5.2

```vue
  <va-avatar loading>
    A
  </va-avatar>
```

image

image

rustem-nasyrov:
Close #2767
Close #2396
Partially #2663

ToDo:
- [ ] Gravatar
- [ ] Fallback slot (resolves text and custom fallback)

## Description
- Remove background color when image source was provided
- Fixed loading state
- AvatarGroup
- Fallback src
image

## Types of changes


packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -86,6 +125,11 @@ export default defineComponent({
   vertical-align: var(--va-avatar-vertical-align);
   border-radius: var(--va-avatar-border-radius);
   font-family: var(--va-font-family);
+  background-color: v-bind('backgroundColorComputed');
```
aluarius, line 128: We heed quotes here only in case of nested value (ex: backgroundColorComputed.secondary).

packages/ui/src/components/va-avatar/hooks/useAvatarProps.ts
```diff
@@ -0,0 +1,7 @@
+export const useAvatarProps = {
+  color: { type: String, default: 'info' },
+  fallbackSrc: { type: String, default: null },
```
aluarius, line 3: As i remember, we have decided not to use `null` for String default.

rustem-nasyrov: Forget about it. Will fix.

packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -26,48 +27,86 @@
...
+
+      return props.src
+    })
+
+    const onLoadError = (event: Event) => {
+      hasLoadError.value = true
+      emit('error', event)
+    }
+
+    watch(() => props.src, () => {
```
aluarius, line 91: Mb?

```suggestion
    watch(srcComputed, () => {
```

packages/ui/src/components/va-avatar/VaAvatar.vue
rustem-nasyrov: Yeah, I understand it. But in my opinion will be nice to decide how we will format it all over the project for better consistency.

packages/ui/src/components/va-avatar-group/VaAvatarGroup.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <div
+    class="va-avatar-group"
+    :class="classComputed"
+    role="group"
+  >
+    <va-avatar
+      v-for="(option, idx) in maxOptions"
```
aluarius, line 8: If we have `role=group`, here we need an `listitem` or `treeitem` role (i'd prefer `listitem`).

packages/ui/src/components/va-avatar/VaAvatar.vue
rustem-nasyrov: It was on purpose. Here I'm watching for `src` changes, not `src` and `fallback` props together.

PR #2800, opened 2022-12-19
Update _header.scss

Valar103769:
image

## Types of changes


m0ksem:
We have this letter spacing according to our design system, so it is not a bug in my opinion. Feel free to discuss it. You can set `letter-spacing: initial;` to prevent such issue.

Issue #2769, opened 2022-12-05
`VaBadge` positioning is weird


aluarius:
We can set position of the badge via `left` and `bottom` separate boolean props.

<img width="267" alt="image" src="https://user-images.githubusercontent.com/64714442/205650707-1b446ff7-97f6-4a12-958d-669c2d54941c.png">

That's pretty weird and i'd like to have one `position` prop that will give me a flexibility to set any (top-left, bottom-left, right-top, right-bottom).

xiongmao86:
I am really sorry the words didn't register when I read this comment for the first time. I have added the `left-top`, `left-bottom`, `right-top` and `right-bottom` to be Placement alias in #2777.

PR #2808, opened 2022-12-23
Feat/va badge updates

aluarius:
Closes: #2769 

## Description
- [x] new `placement` prop, that allows to set any badge position (old left and bottom props are still working until 1.6.0);
- [x] new `offset` prop, that allows to amend badge position relatively to the main axis;
- [x] related refactoring and minor bugfixing;
- [x] docs updates;
- [x] updates to the Vuestic site (somewhere was used a hack for what now `offset` prop does).   

## Types of changes


rustem-nasyrov (changes requested):
image
image
image

packages/ui/src/components/va-badge/VaBadge.demo.vue
```diff
@@ -23,32 +23,22 @@
...
-              text="Bottom+Left"
-              bottom
-              left
-            >
-              Float at
-            </va-badge><br><br>
-            <va-badge text="icon">
-              <va-icon name="face" />
-            </va-badge><br><br>
+          <VbCard title="Transparent">
```
rustem-nasyrov, line 26: Isn't would be better to control transparency by CSS variables?

rustem-nasyrov (approved):

packages/ui/src/components/va-badge/VaBadge.demo.vue
aluarius: Imo not convenient for dynamic states.

PR #2809, opened 2022-12-23
docs: adding document for `colors` of `useColors`

### LINKED ISSUE #2266: Make colors from useColors as writableComputed
image

Can we just have `colors` as writable computed with get and set instead of two methods?

```
const { colors } = useColors()
```

xiongmao86:
Adding document for `colors` of `useColors`.

## Types of changes

### DIFF
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -4076,7 +4076,7 @@
       "ColorConfig": "An object where the keys are the color names and the values are valid CSS colors.",
       "ColorInput": "A valid CSS color",
       "methods": "Methods",
-      "useColors": "Return colors methods and and variables",
+      "useColors": "Return colors methods and variables",
       "setColors": "Used to merge new colors into config or update existing colors.",
       "getColors": "Returns current color config",
       "getColor": "Return color by name",
--- a/packages/docs/src/locales/en/en.json
+++ b/packages/docs/src/locales/en/en.json
@@ -4062,7 +4062,7 @@
     },
     "reactivity": {
       "subtitle": "Reactivity",
-      "about": "As you can see, methods like `getColor`, `getColors` don’t return `Proxy`. If you want to use color as reactive Proxy, use computed properties like in the example above."
+      "about": "Methods like `getColors` return an computed reactive that can also be accessed by name `colors`. The Advantage is forbitting accidantly change by `colors.value = { ... }`, while allowing the properties to mutate, for example you can change `primary` color by writing `colors.primary = '#ff0'`, multiply poperties changes are also supported, you can write `setColors({ primary: '#00f', secondary: '#0ff' })`"
     },
     "otherServices": {
       "subtitle": "Colors config with other services",
@@ -4076,7 +4076,7 @@
       "ColorConfig": "An object where the keys are the color names and the values are valid CSS colors.",
       "ColorInput": "A valid CSS color",
       "methods": "Methods",
-      "useColor": "Return colors methods",
+      "useColors": "Return colors methods and and variables",
       "setColors": "Used to merge new colors into config or update existing colors.",
       "getColors": "Returns current color config",
       "getColor": "Return color by name",
@@ -4084,7 +4084,9 @@
       "getHoverColor": "Return a color that you can use as a `hover` color.",
       "getFocusColor": "Return a color that you can use as a `focus` color.",
       "getGradientBackground": "Return a gradient-color that you can use as a `background-image` CSS property's value.",
-      "hookMethods": "useColors hook methods"
+      "hookMethods": "useColors hook methods",
+      "hookVariables": "useColors hook variables",
+      "colors": "An computed reactive where the keys are color names and values are valid CSS colors."
     }
   },
   "breakpoints": {
--- a/packages/docs/src/page-configs/services/colors-config/page-config.ts
+++ b/packages/docs/src/page-configs/services/colors-config/page-config.ts
@@ -15,6 +15,12 @@ const columnsApiMethods: TableColumn[] = [
   { title: 'description', type: 'markdown' },
 ]
 
+const columnsApiHookVariables: TableColumn[] = [
+  'variable',
+  { title: 'type', type: 'code' },
+  { title: 'description', type: 'markdown' },
+]
+
 const tableDataApiTypes: TableData = [
   ['ColorConfig', '{ [colorName: string]: string; }', 'colorsConfig.api.ColorConfig'],
   ['ColorInput', 'string', 'colorsConfig.api.ColorInput'],
@@ -24,6 +30,7 @@ const tableDataApiMethods: TableData = [
   [
     'useColors',
     `() => {
+      colors,
       setColors,
       getColors,
       getColor,
@@ -32,7 +39,7 @@ const tableDataApiMethods: TableData = [
       getFocusColor,
       getGradientBackground
     }`,
-    'colorsConfig.api.useColor',
+    'colorsConfig.api.useColors',
   ],
 ]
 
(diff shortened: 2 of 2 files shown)


packages/ui/src/composables/useColors.ts
```diff
@@ -42,18 +42,20 @@ export const useColors = () => {
...
+    },
   })
 
-  const setColors = (colors: Partial<ColorVariables>): void => {
-    globalConfig.value.colors!.variables = {
-      ...globalConfig.value.colors.variables,
-      ...colors,
-    } as ColorVariables
+  const setColors = (colorVars: Partial<ColorVariables>): void => {
+    colors = colorVars
```
xiongmao86, line 58: This bailed a warning: 
`Partail<ColorVarialbes>` cannnot assign to `ColorVariables`.
Does `colors = colorvars` not calling `set`?
What should I do to fix this warning?

aluarius, line 58: You definitely shouldn't try to rewrite the whole `colors` variable. May be this also will help with type problem.

aluarius (changes requested):
Implementation should be reworked.

packages/docs/src/components/header/components/HeaderSelector.vue
```diff
@@ -44,12 +44,12 @@ export default defineComponent({
...
 
   setup (props, { emit }) {
-    const { getColors } = useColors()
+    const { colors } = useColors()
 
     const toggleSidebar = () => emit('toggleSidebar', !props.minimized)
 
     return {
-      colors: computed(getColors),
+      colors,
```
xiongmao86, line 52: @aluarius, thank you very much for your help. I take a new way to solve issue 2266. There are some places to change in similar way, So I have pick one place as an example for discussion. I have changed `colors: computed(getColors)` to `colors`, is this the intended solution?

xiongmao86, line 52: @aluarius, thank you very much for your help. I take a new way to solve issue 2266. There are some places to change in similar way, So I have pick one place as an example for discussion. I have changed `colors: computed(getColors)` to `colors`, is this the intended solution?

xiongmao86, line 52: @aluarius, thank you very much for your help. I take a new way to solve issue 2266. There are some places to change in similar way, So I have pick one place as an example for discussion. I have changed `colors: computed(getColors)` to `colors`, is this the intended solution?

aluarius: No, as was said in the initial issue `computed with get and set instead of two methods`. Now we have getColors and setColors, and we need to unite them. In the last example you have just renamed one of the methods.

Your first implementation was on the right way, unlike this one.

xiongmao86:
The last way I think of is to move `globalConfig.value.colors!.variables` into `getColors`, but this doesn't pass the commit checking. I am sorry that I can't think of another way. Could you explain why you want to make this change. What is the background intention, what makes it better to make this change?

xiongmao86:
Oh, never mind, it's closed automatically. Sorry.

xiongmao86:
The last way I think of is to move `globalConfig.value.colors!.variables` into `getColors`, but this doesn't pass the commit checking. I am sorry that I can't think of another way. Could you explain why you want to make this change. What is the background intention, what makes it better to make this change?

rustem-nasyrov:
Hello, @xiongmao86! You don't need to move `globalConfig.value.colors!.variables` to the `getColors` composable. To do this issue (#2266) I suggest you to use a intermediate variable. For example:
```javascript
let intermediate = null;
const colors = computed({
  get: () => intermediate,
  set: (value) => intermediate = value,
})
```
The purpose of the issue is to write less code and improve developer experience. After resolving this issue we'll be able to set and get value from one variable without using extra ones in components.

xiongmao86:
Hi, @rustem-nasyrov, is introducing an intermediate variable without it I can't make a computed `colors`. If hypothetically I manage to move them into a single computed, will it be better off?

P.S. I have look `useReactiveComputed`, and I am confused of why it is implemented that way, why not just use `computed({ get: ..., set: ... }), can I also ask it here?

rustem-nasyrov:
@xiongmao86, yep, you can implement it without intermediate variable. It's just my suggestion.

rustem-nasyrov:
> P.S. I have look useReactiveComputed, and I am confused of why it is implemented that way, why not just use `computed({ get: ..., set: ... }), can I also ask it here?

@m0ksem, could you please answer this? :)

m0ksem:
> > P.S. I have look useReactiveComputed, and I am confused of why it is implemented that way, why not just use `computed({ get: ..., set: ... }), can I also ask it here?
> 
> @m0ksem, could you please answer this? :)

ReactiveComouted is a fun thing to do. Basically, it is the same computed, but while computed returns ref, reactiveComputed returns reactive. 

In default computed you can change computed value like this:

```ts
colors.value = { what: 'newObject' }
```

That's not what you expect from colors object, cuz colors object should stay always the same object, but with different property values. Code from upon becomes:

```ts
colors.what = 'newObject'
```

In such case:
- colors object always stay the same and can not be changed, only properties are mutable.
- colors still be an object being computed, relay on some other reactive value.
- we get rid from .value here (good)
- it is not clear that colors actually reactive value (bad), but it is problem of reactive in Vue.

@xiongmao86, you can take a look at VueUse docs, there should be similar composable.

Basically, yeah. It just reactive that computes, but not ref.

xiongmao86:
> @xiongmao86, yep, you can implement it without intermediate variable. It's just my suggestion.

Sorry @rustem-nasyrov, there might be an misunderstanding, I am not criticizing your judgement, as a newbie, I don't have enough coding knowledge to judge, I ask to know more.

xiongmao86:
Thanks, @m0ksem, great explanation, I can ponder about it this weekend.

rustem-nasyrov:
> > @xiongmao86, yep, you can implement it without intermediate variable. It's just my suggestion.
> 
> Sorry @rustem-nasyrov, there might be a misunderstanding, I am not criticizing your judgement, as a newbie, I don't have enough coding knowledge to judge, I ask to know more.
> 
> I have changed code 2 days ago; can you take a look and see if it's ok to merge?

No worries. I'll check PR a little bit later. :)

xiongmao86:
Thanks, @rustem-nasyrov.

m0ksem:
Hey, @xiongmao86! I checked this PR and found out that this issue was fixed. We use reactiveComputed instead of writableRef.

I think we can close this PR and issue. We have colors to change any colors like:
```ts
colors.primary = '#f0f'
```

or apply new colors from object using 

```ts
setColors({ primary: '#f0f' })
```

What I would appreciate is updating our docs to make it clear how this feature works. For me, it is obvious, cuz I made it, but if you have any suggestions how to improve our docs I would much appreciate it!

Sorry for making you do extra stuff, hope at least you found useReactveComputed interesting :)

xiongmao86:
Hi, @m0ksem, thanks for clarify the situation that I am not capable to discern. I am not very good at documenting; I'll try my best to update the docs. And thanks for explaining `colors`, it's very interesting, and I like it very much.

xiongmao86:
Here is my draft, please let me know your suggestions. @aluarius, @rustem-nasyrov, @m0ksem. Thanks.

packages/docs/src/locales/en/en.json
```diff
@@ -4076,15 +4076,17 @@
       "ColorConfig": "An object where the keys are the color names and the values are valid CSS colors.",
       "ColorInput": "A valid CSS color",
       "methods": "Methods",
-      "useColor": "Return colors methods",
+      "useColors": "Return colors methods and and variables",
```
m0ksem, line 4079:
```suggestion
      "useColors": "Return colors methods and variables",
```

m0ksem (approved):
Looks good.

xiongmao86:
@aluarius, @rustem-nasyrov, @m0ksem, thank you very much for helping me with this pr.

PR #2815, opened 2022-12-27
fix: add fallback color for the button

rustem-nasyrov:
## Description
- Fix button colors on safari <16 versions.

## Types of changes


aluarius (changes requested):
It brokes `plain` state behaviour (hover style, for example).

PR #2819, opened 2022-12-28
feat: add definable generated styles anchor

LighthouseKeeperYN:
## Description
Add definable anchor for the styles generated at runtime. 

Example of the anchor: `<meta name=vuestic-style>`

## Types of changes

### DIFF
--- a/packages/ui/src/utils/dom.ts
+++ b/packages/ui/src/utils/dom.ts
@@ -20,6 +20,8 @@ export const addOrUpdateStyleElement = (id: string, getStyles: () => string): vo
     stylesElement.setAttribute('type', 'text/css')
     stylesElement.setAttribute('id', id)
     stylesElement.innerHTML = getStyles()
-    document.head.append(stylesElement)
   }
+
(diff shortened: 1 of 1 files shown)


rustem-nasyrov (commented):
In the reference they're insert styles before an anchor meta tag.
As I checked `addOrUpdateStyleElement` method usage, it used only in the breakpoint services. We need to check for other generatable styles.

LighthouseKeeperYN:
> In the reference they're insert styles before an anchor meta tag. As I checked `addOrUpdateStyleElement` method usage, it used only in the breakpoint services. We need to check for other generatable styles.

I have already checked and I think all the other styles are part of the components, I'm not sure if they're in the scope of this task.

aluarius:
> As I checked addOrUpdateStyleElement method usage, it used only in the breakpoint services. We need to check for other generatable styles.

Initially it was used by colors service, not sure if it is for now.

m0ksem:
Our styles included by bundler, not by style_inject, so I don't see any reasons why do we need this feature. I prefer to skip this feature until someone asks for it.

PR #2820, opened 2022-12-28
fix: replace error with warning

### LINKED ISSUE #2805: [VaAlert] [Safari]  The string did not match the expected pattern.
**Vuestic-ui version:**
`"@vuestic/nuxt": "^1.0.9"` uses `"vuestic-ui": "^1.4.10"` under the hood

**Steps to reproduce**
1. Open `Safari version 15`
2. Open https://vuestic.dev/en/ui-elements/alert
3. Open Safari devtools and console
4. Check the errors: 
Снимок экрана 2022-12-21 в 15 10 36

No console errors

Error which breaks other components in the project.

**Other information**
The error also happening on `iOS 16` and it basically just breaks all interactions on the page.
As far as I understand it breaks inside `composables/useElementBackground/utils.ts` in `getElementBackgroundFromCSS` function in the `matches` method:

Screenshot 2022-12-21 в 15 13 58

For some reason safari cannot handle the selector (or element?) and explodes. Here's other reference to the issue: https://www.reddit.com/r/firefox/comments/5nbmqi/on_handling_invalid_selector_strings/

rustem-nasyrov:
Close #2805

image

## Description
- Replaced selector error with warning message.

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useElementBackground/utils.ts
+++ b/packages/ui/src/composables/useElementBackground/utils.ts
@@ -1,5 +1,6 @@
 import { getWindow } from '../../utils/ssr'
 import { isCSSVariable } from '../useColors'
+import { __DEV__ } from '../../utils/env'
 
 export type ColorArray = [number, number, number, number]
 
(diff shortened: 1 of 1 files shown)


aluarius (approved):
As i understand, alert won't break the page anymore, but will be still unfunctional in old Safari?

rustem-nasyrov:
> As i understand, alert won't break the page anymore, but will be still unfunctional in old Safari?

It's also a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=460399) in the chromium, but they prefer to not throw an error.

PR #2826, opened 2022-12-30
fix: button state colors for safari browsers

rustem-nasyrov:
## Description
- Fix button colors on safari <16 versions.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-button/hooks/useButtonTextColor.ts
+++ b/packages/ui/src/components/va-button/hooks/useButtonTextColor.ts
@@ -1,6 +1,7 @@
 import { computed, Ref, ComputedRef, getCurrentInstance } from 'vue'
 
-import { getOpacity, useColors } from '../../../composables'
+import { useColors } from '../../../composables'
+import { isServer } from '../../../utils/ssr'
 
 import { ButtonPropsTypes } from '../types'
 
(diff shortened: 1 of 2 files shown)


rustem-nasyrov:
image
Oopsie. :D

aluarius (changes requested):
Now plain states are broken in new (16.2) Safari :c

I guess we need to specify version at `getOpacity`.

packages/ui/src/components/va-button/hooks/useButtonTextColor.ts
```diff
@@ -1,21 +1,45 @@
...
+* @returns {number} Inverted opacity value
+* */
+const getOpacity = (opacity: number): number => {
+  if (isServer()) { return opacity }
+
+  if (/^((?!chrome|android).)*safari/i.test(window?.navigator?.userAgent)) {
+    const isLatestSafari = /(version.)15|16/i.test(window?.navigator?.userAgent)
+
+    if (!isLatestSafari) {
+      return opacity < 1 ? 1 - opacity : opacity
```
m0ksem, line 37: wtf?

Looks like it is impossible to have opacity 0 here. Is that ok?

Should be more documented, cuz for now it is magic.

m0ksem (approved):

packages/ui/src/components/va-button/hooks/useButtonTextColor.ts
rustem-nasyrov: Oopsie, forgot about it. Will fix it. :)

aluarius (approved):
Gj! Now everything works fine everywhere!

Issue #2204, opened 2022-08-04
create-vuestic package


m0ksem:
The idea is to make users able to install vuestic with one command.

```bash
npx create-vuestic
```

This must ask user if he wants to use vuestic-admin or clean project.

- Use Vuestic Admin template? yes\no
- Use Nuxt3? yes/no
- Use tree-shaking?
  - Select CSS modules:
    - typography, grid, reset
  - Select plugins
    - breakpoints, css variables ???
    
see #1991

asvae:
We have notion tasks for this. Maybe we can incorporate it here?

PR #2828, opened 2023-01-02
feat: create-vuestic

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2204

CLI tool to create new app with vuestic-ui. Should replace vue-cli-plugin-vuestic. Must be recommended way to create Vuestic Admin app too.

```
npm create vuestic@latest
```
image

Three templates:
image

Prompts:
- TreeShaking (for vite & nuxt)
- AgGrid
- * Might add epicspinners
- * Might be tailwind


aluarius:
Just in case (for nuxt):

image

aluarius:
Think we should provide here more info (for dummies).

- Vue 3 + Vite
- Nuxt 3

image

aluarius:
May be we should add test Vuestic UI components to the default App.vue? Other UI frameworks do. Instant demonstation.

aluarius:
Does `treeshaking` option affects result for nuxt?

aluarius:
image

aluarius:
I'd prefer not to make preselected options, because it's not obvious - i've thought that green circle is a currently hovered option.

image

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const addVuesticToVue3App = async () => {
+  const { projectName } = await useUserAnswers()
+
+  // Install vuestic-ui
+  const { addDependency } = await usePackageJson()
+  addDependency('vuestic-ui', 'latest')
+
+  // Add plugin
+  const mainPath = resolvePath(process.cwd(), projectName, 'src/main.js') || resolvePath(process.cwd(), projectName, 'src/main.ts')
```
aluarius, line 19: Why not via resolve-path.ts?

packages/create-vuestic/src/steps/2.addVuestic/insert-nuxt-module.ts
```diff
@@ -0,0 +1,42 @@
...
+  vuestic: {
+    ${vuesticConfig.split('\n').join('\n    ')}
+  },
+`
+}
+
+export const insertNuxtModule = (source: string, css: string[]) => {
+  const vuesticConfig = createVuesticConfig(css)
+
+  if (/vuestic: \{/.test(source) || /modules: \[((.|\n)*)\]/gm.test(source)) {
```
aluarius, line 34: Does this situation possible? 0_o

aluarius (approved):

packages/create-vuestic/src/steps/2.addVuestic/insert-nuxt-module.ts
m0ksem: Not yet. But would be nice if create-vuestic can append vuestic-ui to already existing project in future.

packages/create-vuestic/src/generators/create-vue.ts
```diff
@@ -0,0 +1,13 @@
+import { execp } from './../utils/exacp';
+
+type CreateVueFeature = 'ts' | 'jsx' | 'router' | 'pinia' | 'tests' | 'vitest' | 'cypress' | 'playwright' | 'eslint'
```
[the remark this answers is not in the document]
m0ksem: I was thinking if we should copy create-vue behavior, but missing playwright makes this CLI easier, so I think to skip it for now. This type represents create-vue features, but we will ignore playwright (propose cypress instead). And I just forgot about JSX.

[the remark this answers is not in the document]
m0ksem: Having both playwright and cypress requires to have select between this two, but we have multipleselect of all create-vue features in one window.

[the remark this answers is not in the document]
m0ksem: I added playwright and jsx here.

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
m0ksem: I use here resolvePath, no?

rustem-nasyrov (approved):

PR #2829, opened 2023-01-02
fix(#2827): remove va-card__inner div from VaCard

### LINKED ISSUE #2827: va-card click bound to va-card__inner for some reason, not on parent
image

This makes impossible to handle click event on parent in native way (need for a directive), because click is handled in component, stopPropagation and emits vue event, not native.

m0ksem:
closes #2827

Prev
image

Now
image

No more extra va-card__inner div.


rustem-nasyrov (approved):
Great job!

Issue #2824, opened 2022-12-29
It is unclear what is colors config variables 


m0ksem:
**Vuestic-ui version:** 1.5.3

**Steps to reproduce**

**What is the expected behavior?**

**What is the current behavior?**

**Other information**

We have 2 presets and current variables. If we use darktheme, that means that user should change presets in config. If not, user should change current variables. 

I see two ways to deal with it:
- write about this behaviour in docs and hope people read them
- make variables actually a proxy to current preset values, so in default configuration variables and light preset actually the same object.

PR #2830, opened 2023-01-02
Fix#2821/nuxt color presets fix

### LINKED ISSUE #2821: Dark Theme on Nuxt 3
**Vuestic-ui version:** 1.5.3

**Steps to reproduce**
I just wrote the code for setting the theme to "dark" in the app.vue. It does make some of the components dark but not all, the base background color and va-input components are not affected. However, if I write the code for setting the theme to dark in some component in the /pages directory, the dark theme works fine and as expected in Hot Module Reload, but if I refresh the page, the base background color and va-input components aren't affected with the dark theme again.

All components in the app will turn "dark"

**Other information**
Nuxt Version 3 (Stable)

m0ksem:
closes #2821
close https://github.com/epicmaxco/vuestic-ui/issues/2824

Need to double-check if issue is gone with this PR.

### DIFF
--- a/packages/nuxt/src/runtime/plugin.ts
+++ b/packages/nuxt/src/runtime/plugin.ts
@@ -5,7 +5,7 @@ import {
   VaModalPlugin,
   BreakpointConfigPlugin,
 } from 'vuestic-ui'
-import { ref } from 'vue'
+import { ref, watchEffect } from 'vue'
 
 import type { VuesticOptions } from '../types'
 import { defineNuxtPlugin } from '#app'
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/runtime/plugin.ts
```diff
@@ -42,18 +42,19 @@ export default defineNuxtPlugin((nuxtApp) => {
...
-        htmlAttrs: {
-          style: renderCSSVariables()
-        }
-      }))
-    }
+    watchEffect(() => {
+      const colorConfig = getGlobalProperty(app, '$vaColorConfig')
+
+      if (colorConfig) {
+        const renderCSSVariables = colorConfig.renderCSSVariables
```
aluarius, line 49: May be will be better to handle & re-render only changed-new colors from color config? In performance purpose.

aluarius (approved):

aluarius:
`write about this behaviour in docs and hope people read them` - this isn't proceeded via this pr.

packages/nuxt/src/runtime/plugin.ts
m0ksem: I think it is a minor thing

Issue #2785, opened 2022-12-10
VaDataTable timeout


asvae:
Implement timeout for table actions.

Initial case:
Big table freezes on filtration

Potential solution:
* Implement a `timeout` prop, that would apply to all table updates, so that table would debounce for a bit for expensive changes (rendering mostly), allowing user to do multiple actions without freeeze.
* That should be applied to filter, sorting, pagination update, data changes.
* Solutions should be relatively easy to support - maybe we can move it to composable, to then reuse in some other components, like select.

PR #2832, opened 2023-01-03
[#2785] Throttling composable, integration to VaSelect and VaDataTable

aluarius:
Closes: #2785 

## Description
- [x] Added `useThrottleValue` and `useThrottleFunction` composables. First one accepts reactive variable and returns its actual value once per delay period (in the end, if no other changes happens, it returns value, that is actual for the last call, even it was throttled). The second one just returns throttled function.
- [x] Integration to the VaSelect (options search) and VaDataTable (sorting, filtering, pagination and any other data changes). 
- [x] Demos and docs update. 

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-data-table/hooks/useFilterable.ts
+++ b/packages/ui/src/components/va-data-table/hooks/useFilterable.ts
@@ -1,10 +1,11 @@
 import { Ref, watch, computed, PropType, ExtractPropTypes } from 'vue'
 
-import { useThrottleValue } from '../../../composables'
+import { useThrottleValue, useThrottleProps } from '../../../composables'
 
 import type { DataTableRow, DataTableFilterMethod, DataTableItem } from '../types'
 
 export const useFilterableProps = {
+  ...useThrottleProps,
   filter: { type: String, default: '' },
   filterMethod: { type: Function as PropType<DataTableFilterMethod | undefined> },
 }
--- a/packages/ui/src/components/va-data-table/hooks/usePaginatedRows.ts
+++ b/packages/ui/src/components/va-data-table/hooks/usePaginatedRows.ts
@@ -2,11 +2,12 @@ import { Ref, computed, PropType, ExtractPropTypes } from 'vue'
 
 import { useCurrentPageProp } from './useCommonProps'
 
-import { useThrottleValue } from '../../../composables'
+import { useThrottleValue, useThrottleProps } from '../../../composables'
 
 import type { DataTableRow } from '../types'
 
 export const usePaginatedRowsProps = {
+  ...useThrottleProps,
   ...useCurrentPageProp,
   perPage: { type: Number as PropType<number | undefined> },
 }
--- a/packages/ui/src/components/va-data-table/hooks/useSortable.ts
+++ b/packages/ui/src/components/va-data-table/hooks/useSortable.ts
@@ -1,6 +1,6 @@
 import { PropType, computed, ref, Ref, watch, ExtractPropTypes, ComputedRef } from 'vue'
 
-import { useThrottleFunction } from '../../../composables'
+import { useThrottleFunction, useThrottleProps } from '../../../composables'
 
 import type {
   DataTableColumnInternal,
@@ -11,6 +11,7 @@ import type {
 } from '../types'
 
 export const useSortableProps = {
+  ...useThrottleProps,
   sortBy: { type: String as PropType<string | undefined> },
   sortingOrder: { type: String as PropType<DataTableSortingOrder | undefined> },
 }


packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type UseThrottleProps = ExtractPropTypes<typeof useThrottleProps>
+
+/**
+ * @param fn passed function to throttle
+ * @param props { delay } call delay in ms
+ */
+export function useThrottleFunction<Output> (fn: ThrottledFunction<Output>, props: UseThrottleProps) {
+  const delay = toRef(props, 'delay') ?? 0
```
rustem-nasyrov, line 35: If the delay is `0`, then we can return value without going down to the rest of the function?

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  return function (this: ComponentInternalInstance, ...args: ThrottledFunctionArgs) {
+    const invoke = () => fn.apply(this, args)
+
+    if (!unref(delay)) { return invoke() }
+
+    if (isThrottled.value) {
+      isThrottled.value = false
+
+      setTimeout(() => (isThrottled.value = true), unref(delay))
```
rustem-nasyrov, line 54: We should clear timers.

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    return lastCallResult.value
+  }
+}
+
+/**
+ * @param value passed reactive value to throttle
+ * @param props { delay } call delay in ms
+ */
+export function useThrottleValue<T> (value: Ref<T>, props: UseThrottleProps): Ref<T> {
```
rustem-nasyrov, line 67: If we don't need `this` here, maybe better to use the function expression?

packages/ui/src/composables/useThrottle.ts
aluarius: See 49 line.

packages/ui/src/composables/useThrottle.ts
aluarius: We should clear intervals, but why timeouts?

packages/ui/src/composables/useThrottle.ts
aluarius: Are there any reasons for this? I thought about it, but suggested that two pretty similar functions in the one file should be written in the same way.

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+}
+
+type UseThrottleProps = ExtractPropTypes<typeof useThrottleProps>
+
+/**
+ * @param fn passed function to throttle
+ * @param props { delay } call delay in ms
+ */
+export function useThrottleFunction<Output> (fn: ThrottledFunction<Output>, props: UseThrottleProps) {
```
m0ksem, line 34: Because it is a composable, it would be correctly to pass `Ref<number>` as dealy instead of props.

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+  const delay = toRef(props, 'delay') ?? 0
+
+  const isThrottled = ref(true)
+
+  /**
+   * No way this will be returned without reassignment, so we don't want typescript
+   * to always keep undefined as possible return type. If function returns undefined itself
+   * it will be still presented by typescript as undefined (expected behaviour).
+   */
+  const lastCallResult = ref<Output>(undefined as any as Output) as Ref<Output>
```
m0ksem, line 44: Why is it ref?

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  return function (this: ComponentInternalInstance, ...args: ThrottledFunctionArgs) {
```
m0ksem, line 46: Can't see any reason why here `this` is a `ComponentInternalInstance`. For example, in `useSortable` there is no this in function. BTW, there likely always no this, cuz it used in setup function.

If there will be a chance to have this, you can always use bind().

packages/ui/src/composables/useThrottle.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const isThrottled = ref(true)
+
+  const previousCallValue = ref<T>() as Ref<T>
+  const previousReturnedValue = ref<T>() as Ref<T>
+  const currentCallValue = ref<T>() as Ref<T>
+
+  watch(value, () => {
+    // we save and return last call value at the end if no more calls were provided
+    previousCallValue.value = value.value
```
m0ksem, line 80: Can't we use useThrottleFunction here?

m0ksem (changes requested):
I like the idea, but have some quesitions.

Also, what the advantage of having setTimeout instead of `prevDate - new Date() > delay`?

aluarius:
> what the advantage of having setTimeout instead of prevDate - new Date() > delay

It may be reimplemented in this way:

image

But, first of all,  setTimeout seems to be essencial for this situation. Secondly, the another place of use can not be reimplemented via suggested scheme, because we need to return the value without next function call.

image

So, i just don't want to have different implementations of the same things in one function.

packages/ui/src/composables/useThrottle.ts
aluarius: And you're suggesting not to reuse useThrottleProps?

packages/ui/src/composables/useThrottle.ts
aluarius: I'll change the type to any, but prefer to leave it:
- i'm not absolutely sure that there're no way to have no `this` - current state serves both versions;
- we should pass `this` anyway, even it's an empty object.

packages/ui/src/composables/useThrottle.ts
aluarius: No, they're pretty different. May be i'll be able to reuse it here somehow, but it won't code shorter, but more complicated.

rustem-nasyrov (approved):

m0ksem (approved):

m0ksem:
Resolve conflicts and need qa after two weeks :)

rustem-nasyrov:
@aluarius please, take a look at this issue: #2837. Is it possible to add given feature to this PR?

m0ksem:
> @aluarius please, take a look at this issue: #2837. Is it possible to add given feature to this PR?

How about different PR?

PR #2840, opened 2023-01-04
Feat/tree view improvements

### LINKED ISSUE #2545: Tree view suggestions
The `TreeNode` component must be not available to end users.
- [x] In the docs we need to unite slots from TreeNode to the TreeView component.
image
- [ ] Make public exportable TreeView and TreeNode types and private ones.
- [ ] Create TreeNodeType as generic.
- [x] Pass to the bindable slots whole TreeNode type (props).
- [x] We need to expose only Treeview here:
image

- [x] Separate colors and select (checkbox) demos:
image

### LINKED ISSUE #2638: [va-tree-view] on-node-click event
image

### LINKED ISSUE #2639: [va-tree-view] open tree-view by clicking on icon only
image

Is it possible to not open node if clicked on text, but on arrow icon only?

I managed it by stoping propagation from label, so click on label does what I need

```vue
<va-tree-view :nodes="objectNodes(o)">
  <template #content="node">
    <div @click.stop="selectedObject = o" class="w-full">
      {{ node.label }}
    </div>
  </template>
</va-tree-view>
```

### LINKED ISSUE #2661: VaTreeView keyboard navigation doesn't work
**Vuestic-ui version:** 1.5.0

- [ ] Should work

rustem-nasyrov:
Close #2639
Close #2638
Close #2545
Close #2661

## Description
- Keyboard navigation
- Refactoring and improvements
- VueBook: added demo for testing keyboard navigation

## Types of changes


m0ksem (approved):
Looks good for me. Have you found a guide for keyboard navigation somewhere? I would like to select node with space, because for now I need to select checkbox with tab and then go back.

If you used some guide, for example, w3c, how about mentioning it in docs?

rustem-nasyrov:
> Looks good for me. Have you found a guide for keyboard navigation somewhere? I would like to select node with space, because for now I need to select checkbox with tab and then go back.

What you think about this scenario:
- Pressing on `space` key will emit `selected` event while tree view isn't in `selectable` mode.
- Pressing on `space` key will toggle checkbox (`selectable` mode).
> If you used some guide, for example, w3c, how about mentioning it in docs?

I've used this [guide](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role#keyboard_interactions) from MDN.

m0ksem:
> > Looks good for me. Have you found a guide for keyboard navigation somewhere? I would like to select node with space, because for now I need to select checkbox with tab and then go back.
> 
> What you think about this scenario:
> 
>     * Pressing on `space` key will emit `selected` event while tree view isn't in `selectable` mode.
> 
>     * Pressing on `space` key will toggle checkbox (`selectable` mode).
> 
> 
> > If you used some guide, for example, w3c, how about mentioning it in docs?
> 
> I've used this [guide](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role#keyboard_interactions) from MDN.

So let's use space as described here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role#multi-select_keyboard_interactions

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
```diff
@@ -0,0 +1,180 @@
...
+  emit: TreeViewEmitsFunc,
+}
+
+const useTreeKeyboardNavigation = (
+  props: ExtractPropTypes<typeof useTreeViewProps>,
+  methods: UseTreeKeyboardNavigationMethods,
+) => {
+  const { emit, toggleNode, toggleCheckbox } = methods
+  const isElementExpanded = (currentElement: TreeNodeElement): boolean =>
+    currentElement?.getAttribute('aria-expanded') === 'true' || false
```
aluarius, line 17:
```suggestion
    !!currentElement?.getAttribute('aria-expanded') === 'true'
```

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (!currentElement) {
+      return null
+    }
+
+    let previousElement = getPreviousElement(currentElement)
+    let lastChildElement = isElementExpanded(previousElement) && getLastChildElement(previousElement)
+
+    if (lastChildElement) {
+      do {
+        if (isElementExpanded(lastChildElement)) {
```
aluarius, line 35: Imo it's very deep - can remove 1-2 levels via early break.

aluarius (approved):
Great job, i've done with code. Next i'll leave a few suggestions related to fucntionality.

aluarius:
Ignore if disagree. Just thought by me, who newer tryed treeview keyboard navigation before.

1. Better to handle here also arrows left-right to move up-down (it's universal prev-next).
image

aluarius:
2. I've tried space and enter to expand it and just eventually realized, that here used right-left. Definetly should be updated.
image

aluarius:
I give up :) All specifications say that enter-space should select node. But for me it looks unobvious.

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  const getLastChildElement = (currentElement: TreeNodeElement): TreeNodeElement => {
+    return (currentElement?.querySelector('.va-tree-node-children')?.lastElementChild || null) as TreeNodeElement
+  }
+
+  const onHorizontalMove = (currentElement: TreeNodeElement, dir: 'left' | 'right') => {
+    if (dir === 'left') {
+      getParentElement(currentElement)?.focus()
+    } else if (dir === 'right') {
+      if (isElementExpanded(currentElement)) {
```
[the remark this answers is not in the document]
aluarius: Ah, i see. Newermind.

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
rustem-nasyrov: Get attribute returns string value.

LighthouseKeeperYN: You can still remove "|| false"

aluarius: > Get attribute returns string value.

Yea, just inattentive look. Here just `currentElement?.getAttribute('aria-expanded') === 'true'` will be enough.

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
LighthouseKeeperYN: I recommend avoiding while loop in general and limit usage of for loop to the minimum. Everything that can be done by those can by done in more readable declarative way using array methods.

LighthouseKeeperYN: Though I'm not sure there are other ways if it's linked list type of thing you're dealing with.

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
```diff
@@ -97,24 +98,28 @@ export default defineComponent({
...
       toggleNode,
-      getNodeProperty,
       toggleCheckbox,
+      getNodeProperty,
+      handleKeyboardNavigation,
     } = useStrictInject(TreeViewKey, INJECTION_ERROR_MESSAGE)
 
     const labelComputed = computed(() => getText(props.node) || '')
-    const isExpandedComputed = computed(() => !!props.node.expanded)
+    const isExpandedComputed = computed(() => props.node.hasChildren ? !!props.node.expanded : undefined)
```
LighthouseKeeperYN, line 113: const isExpandedComputed = computed(() => props.node.hasChildren && !!props.node.expanded)

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
```diff
@@ -123,23 +128,28 @@ export default defineComponent({
 
     const tabIndexComputed = computed(() => props.node.disabled ? -1 : 0)
 
-    const handleToggleNode = (event: Event, node: TreeNode) => {
-      if (node.expanded) {
-        (event.target as HTMLElement)?.blur()
-      } else {
-        toggleNode(node)
+    const onNodeClick = (type: typeof expandNodeBy) => {
+      type = expandNodeBy === 'node' && type === 'leaf' ? 'node' : type
```
LighthouseKeeperYN, line 132: I recommend creating new variable instead of reassigning arguments. The name of the variable will hint to what what code does making it more readable.

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
LighthouseKeeperYN, line 106:
```suggestion
    } else if (isElementExpanded(currentElement)) {
```

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  const onVerticalMove = (currentElement: HTMLElement, dir: 'up' | 'down') => {
+    if (dir === 'up') {
+      findPreviousElement(currentElement)?.focus()
+    }
+
+    if (dir === 'down') {
+      findNextElement(currentElement)?.focus()
+    }
```
LighthouseKeeperYN, line 119: if (dir === 'up') {
      findPreviousElement(currentElement)?.focus()
    } else {
      findNextElement(currentElement)?.focus()
   }

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  const handleKeyboardNavigation = (event: KeyboardEvent, node: TreeNode) => {
+    const currentElement = event.target as HTMLElement
+
+    switch (event.code) {
```
LighthouseKeeperYN, line 125: Brackets are redundant for switch case. You can write it like this:
`      case 'ArrowUp':
        onVerticalMove(currentElement, 'up')
        break
`

packages/ui/src/components/va-tree-view/hooks/useTreeKeyboardNavigation.ts
rustem-nasyrov: Yep, it's kinda a linked list.

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
rustem-nasyrov: I need here `undefined` for the purpose of removing the `aria-*` attribute if is node expandable/not expandable.

Issue #2839, opened 2023-01-04
All CSS always import in Nuxt 3


azoom-nguyen-van-nam:
**Vuestic-ui version:** 1.5.3 and **@vuestic/nuxt:** 1.0.9

I used Vuestic with `Nuxt 3` and config as below:

![image](https://user-images.githubusercontent.com/101168054/210527692-53b0cae8-b603-465b-bf10-4765297f3f5a.png)
![image](https://user-images.githubusercontent.com/101168054/210527141-30a98f38-0066-4060-85e9-585888eee5fb.png)

But in the head tag of HTML document, even though I only use 1 button on each page, I always see all the CSS being imported.
<img width="1314" alt="image" src="https://user-images.githubusercontent.com/101168054/210528297-1b10f575-6344-4bce-b786-7b96c824b61b.png">

and the CSS maybe duplicated: ⬇️ 
<img width="1317" alt="image" src="https://user-images.githubusercontent.com/101168054/210530874-e48dc3ef-3292-4f40-8f58-e41312155386.png">

If it was built for Production ENV, would it still be that way?
**_Please help me explain or point out the error?_**

m0ksem:
Hey, @azoom-nguyen-van-nam. Please use `@vuestic/nuxt` following guide below:

https://vuestic.dev/en/getting-started/nuxt#installation

`nuxtApp.vueApp.use(createVuestic())` is an incorrect way to register vuestic plugin, because nuxt will not tree-shake anything.

With  `@vuestic/nuxt` you don't need to import `vuestic-ui` at all.

If you still have some questions, feel free to ask for help in [Discord](https://discord.gg/u7fQdqQt8c).

azoom-nguyen-van-nam:
@m0ksem oh sorry. It did't work!
![image](https://user-images.githubusercontent.com/101168054/210681910-b84ce532-9323-4a78-aa48-34398acd6659.png)

<img width="1435" alt="image" src="https://user-images.githubusercontent.com/101168054/210682421-740695f0-8f4c-402f-81f2-a03f72161e9b.png">

Although it appears to have lost one import, all of the CSS is still imported into the head tag.

m0ksem:
Thanks, @azoom-nguyen-van-nam. Can confirm this issue is related to `@vuestic/nuxt` package when all CSS set to true.

You can prevent all CSS to be registered for now by manually defining which CSS modules you want to use:
```ts
  vuestic: {
    // ...
[...]

PR #2842, opened 2023-01-05
fix(#2839): do not register vuestic-ui.css in nuxt which contains all…

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2839

It looks like we were register vuestic-ui.css which contains all css for IIFE and CJS builds. I'm not sure if we want to import vuestic/css which should return different css for es and cjs build. So instead I choose to register vuestic/styles/index.css manually.

Please, need some sort of QA.


aluarius:
Have you faced problem like this?

image

Issue #2852, opened 2023-01-06
Check lighthouse issues


asvae:
Go through the docs with lighthouse and see where we have issues. 
Either fix these issues or document them into different issues then link to this one.

Issue #743, opened 2021-05-13
Docs refresh


asvae:
Updating docs is a bit problematic. As you can have update happen when user has page loaded.

If they try to navigate to some page - they might have an issue, as module might be not available anymore.

So we want to introduce some solution to make user interaction seamless.

Couple of ideas:
* keep multiple versions of bundle on server (let's say up to 24h), each in its own folder.
* sometimes send a request (to some json, where version is defined?) and if anything is changed (flags or version) - present user with update dialog similarly to what other docs use. This is not ideal as it requires action from user.

Issue #650, opened 2021-04-22
Move documentation to nuxt when it's ready for vue3


asvae:
We implemented nuxt docs before (see 87de7304be35def5cd60e96c5711dfa6ff75fa7b - `/packages/vue-cli-docs/old/nuxt-docs` ), but had to switch to vue-cli because it wasn't ready.

Ideally we still want to use nuxt for seo/speed reasons. It's also not that hard to implement.

- [ ] Implement documentation on nuxt. Don't leave vue-cli docs.
- [ ] Ensure things work: CI, navigation etc.
- [ ] Ensure there is no bugs (also check the leftovers in #1577)

See also https://github.com/epicmaxco/vuestic-ui/issues/1642

Nuxt docs were worked on, but removed in [this PR](https://github.com/epicmaxco/vuestic-ui/pull/2103), you can use these changes when you work on implementation.

The idea is that we want to only have one docs project in repo to avoid confusion.

We can also check https://github.com/epicmaxco/vuestic-ui/pull/1142 in context of this PR

m0ksem:
We probably need to wait on nuxt release or move to pnpm from yarn.

PR #2854, opened 2023-01-06
[nuxt-docs] define-page-config plugin

m0ksem:
Highest level of overengineering that human can ever imagine

Basically, here is a plugin-compiler for page-configs. It handles imports from name. I thought it is a good idea, to skip ugly
```
const block = new PageGenerationHelper(__dirname)
```

Then it also skips imports of code examples. Also, code example now is file, so we have syntax highlight in code examples now.

It is easier to write page config now, but much harder to add new block. At first, I thought we can compile page-configs to skip markdown step, but because of locale we need to compile them in runtime(in ssr, but runtime).

I'm thinking to implement runtime blocks, like we had previously. 

I'm thinking about making runtime block generator and make it as auto-import. Then, if it is block.component - compiler will resolve full path to component. This way it will be still easy to add new blocks, but also possible to make some additional pre-compile if needed.

Thanks for the review, @aluarius, now I know where to dig.

closes https://github.com/epicmaxco/vuestic-ui/issues/743 - with nuxt we wouldn't have this stupid caching (it's not working offline btw)
closes https://github.com/epicmaxco/vuestic-ui/pull/2861
closes https://github.com/epicmaxco/vuestic-ui/issues/2852
closes https://github.com/epicmaxco/vuestic-ui/issues/650

TODOS:
- [x] Include fontawesome and iconify fonts
- [ ] VaButtonToggle i18n keys not found
- [ ] Refactor compile to use transform functions inside blocks.
- [ ] Move newer added docs to nuxt docs (like web-components, tailwind and create-vuestic).
- [ ] Use tailwind from CDN (I hope my browser will cache it for outage coding :))) )


m0ksem:
old docs link https://github.com/epicmaxco/vuestic-ui/tree/10043bd249f38116b9dda0ec789b94cb1643c978

m0ksem:
Partially implements https://github.com/epicmaxco/vuestic-ui/issues/2881

m0ksem:
Added playground component:
image

Should be easy to make and edit them,

aluarius:
1. Some (lots) translations are broken (especially titles). Examples:
- /en/introduction/accessibility-guide#accessibility-guide-examples-keyboard-interactions-title
- /en/getting-started/configuration-guide#configuration-guide-examples-page-config-button-default-title

2. Hash links are totally broken. Example: /en/getting-started/installation#installation-quick-start-title

3. Don't forget about bundle size script.

image

4. Strange code formatter:

image

image

5. Smthng wrong (/en/getting-started/ssr-guide):

image

6. /en/getting-started/kitchensink broken.

7. /en/styles/grid - all example a little bit drunk.

8. In all components no way to return default preset:

image

9. Go here - /en/ui-elements/data-table - and then navigate another page. Content won't change.

10. If page doesn't have meta (title, for example), the last page's title will be shown - it's weird.

11. We should be able to set meta title, meta description and meta robots. I'll be happy to have ability to set `OpenGraph` rich protocol (for beautiful snippets in search results), but it's up to you.

m0ksem:
@aluarius What we do with removing grid? And what's with SmartHelpers? I completely removed Vuestic Grid from Nuxt, so it is in pure tailwind now. I'll need some help here.

aluarius:
> And what's with SmartHelpers?

We won't touch them. They're not conflicting with Tailwind.

> What we do with removing grid?

Ideally,  we should keep them until all grid helpers (except smart ones) will be moved to Tailwind.

aluarius:
No favicon:

image

aluarius:
Minor - incorrect helpers style in search input:

image

aluarius:
Minor, but major for me:

image

My eyes are really bleeding because of this red - can't read without pain.

aluarius:
Minor - wrong gap:

image

aluarius:
/en/introduction/accessibility-guide - missing texts (see https://vuestic.dev/en/introduction/accessibility-guide)

rustem-nasyrov:
Big paddings in mobile version. Code font-size is big. 
image

rustem-nasyrov:
Slider
image

Issue #2855, opened 2023-01-09
Some Aria attributes are wrongly rendered without hyphen


asbermudez:
**Vuestic-ui version:** 1.5.3

**Steps to reproduce**

Use va-input
```html
<VaInput />
```

**What is the expected behavior?**

All the aria attributes rendered in the input should be use hyphen, like:

```html
<input aria-readonly="false" aria-required="false" aria-disabled="false" [...] />
```

**What is the current behavior?**

Some of the attributes are rendered without hyphen, resulting in wrong html properties, like:

```html 
<input
  ariadisabled="false"
  arialabel="Some label"
  ariareadonly="false"
  ariarequired="false"
/>
```

**Other information**

This might be related to the issue:
https://github.com/vuejs/core/issues/5477

PR #2856, opened 2023-01-09
[#2855] Aria attributes hydration fix

aluarius:
Closes: #2855 

## Description
Minor updates and checks of aria attributes keys hydration. Let's use them in kebab case to prevent possible problems in future.

## Types of changes


m0ksem (approved):
We'll forget about it in few days. How about to make some useAriaComposable which accepts only kebab-case attributes. This can be done with TS, so you don't need to take here lodash. Or we can have lodash.kebabCase here for each attribute.

rustem-nasyrov (approved):
LGTM.

aluarius:
What this composable should do except `useAriaAttributes composable which accepts only kebab-case attributes`? Just checking key names?

m0ksem:
> What this composable should do except `useAriaAttributes composable which accepts only kebab-case attributes`? Just checking key names? Just can't see the way for it to be more useful and stay universal.

Yeah. Something like:
```ts
const useAriaAttributes = <T>(o: ComputedRef<KebabCaseKeys<T>>) => o
```

aluarius:
> Something like

Honestly, can't see any value here, especially because they're pretty static - won't change more then once in a few months. So, changing all correct aria attributes to the composable that is doing the same is pretty weird, but can discuss.

m0ksem:
> > Something like
> 
> Honestly, can't see any value here, especially because they're pretty static - won't change more then once in a few months. So, changing all correct aria attributes to the composable that is doing the same is pretty weird, but can discuss.

This might be helpful for future components, that made by someone who doesn't aware of this issue.

Issue #2796, opened 2022-12-16
[big bug]: if fonts.googleapis.com fail, the page is breakdown


Valar103769:
when 
```js
@import url('https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;1,700&display=swap');
@import url('https://fonts.googleapis.com/icon?family=Material+Icons');

```

fetch error ,
the page's font has problem,
<img width="1509" alt="image" src="https://user-images.githubusercontent.com/36926073/208030257-c0b0467f-c31a-415e-a28b-edb753ac717a.png">

1. the `fonts.googleapis.com/` not just used to icon ?  way can cause font error? 
 
2. in china, can't use  `google` and `googleapis.com`,  There are not many ICONS, can you build in some? Or use the npm package, not the cdn,

Valar103769:
@asvae There's a production problem, one or two more and I'm out of here 
<img width="164" alt="image" src="https://user-images.githubusercontent.com/36926073/208031088-2f43d6eb-7b46-4d1a-87bc-1b48f9e3ddbb.png">

Valar103769:
this is font-family problem? 

https://www.bilibili.com/video/BV1b54y1Z7pu/?spm_id_from=333.999.0.0

<img width="1098" alt="image" src="https://user-images.githubusercontent.com/36926073/208031641-da8de065-80d3-4d23-85ff-687aeacf657c.png">

```js
@font-face {
  font-family: Emoji;
  src: local('Apple Color Emojiji'), local('Segoe UI Emoji'), local('Segoe UI Symbol'), local('Noto Color Emoji');
  unicode-range: U+1F000-1F644, U+203C-3299;
}
.font-serif {
  font-family: Georgia, Cambria, 'Times New Roman', Times, serif;
}
.font-mono {
  font-family: Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
}
```
The author of the video is one of the strongest css figures in our country

rustem-nasyrov:
Hi, @Valar103769! If it's possible could you please download fonts directly into the project and use it? [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro?query=source+sans+pro) and replace icons with this [icon library](https://www.npmjs.com/package/material-design-icons-iconfont).

rustem-nasyrov:
[...]

PR #2861, opened 2023-01-11
Docs/create vuestic

m0ksem:
Docs for https://github.com/epicmaxco/vuestic-ui/pull/2828 (must be merged after #2828)

Just docs here for review. Find something unclear.


packages/create-vuestic/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const answers = await useUserAnswers()
+
+  console.log('')
+
+  await scaffoldProject(answers)
+  await addVuestic(answers)
+  await addAgGrid(answers)
+  await initGit()
+  await installDeps()
```
LighthouseKeeperYN, line 27: Do they have to be sequential? If not you can use Promise.all to resolve them in parallel.

LighthouseKeeperYN (approved):
lgtm

packages/create-vuestic/src/index.ts
m0ksem: Most of the yes. We need to scaffold project before adding vuestic, also deps should be installed after all complete. For me, it is just clear to have it all sequential.

aluarius (approved):

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
```diff
@@ -0,0 +1,46 @@
...
+  ])
+  await writeFile(mainPath, mainSource)
+
+  // Add fonts
+
+  const htmlPath = resolvePath(process.cwd(), projectName, 'index.html')!
+  let htmlSource = await readFile(htmlPath, 'utf-8')
+  htmlSource = insertHead(htmlSource, [
+    `<link href="https://fonts.googleapis.com/css2?family=Source+Code+Pro&family=Source+Sans+Pro:wght@400;700&display=swap" rel="stylesheet">`,
+    `<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">`
```
rustem-nasyrov, line 43: What if users don't have access to these CDNs?

packages/create-vuestic/src/utils/package-manager.ts
```diff
@@ -0,0 +1,25 @@
...
+    return manager
+  }
+  // This environment variable is set by npm and yarn but pnpm seems less consistent
+  const userAgent = process.env.npm_config_user_agent
+
+  if (userAgent) {
+    if (userAgent.startsWith("yarn")) {
+      manager = "yarn"
+    } else if (userAgent.startsWith("pnpm")) {
+      manager = "pnpm"
```
rustem-nasyrov, line 15: I saw that nuxt recommends to use `--shamefully-hoist` flag on installing. Maybe when is project is generating will be nice to add `.npmrc` file? Where we can put this flag.

rustem-nasyrov (commented):
Great work!

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
m0ksem: Then he will need to pay for internet. I can't see any possibility of not having access to CDN. 
Btw, it is possible to change CDN links and install fonts manually. But I would recommend to use CDN for fonts and icons.

packages/create-vuestic/src/utils/package-manager.ts
m0ksem: Nudt already does it, so we don't need to.

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
rustem-nasyrov: @m0ksem https://github.com/epicmaxco/vuestic-ui/issues/2796

PR #2867, opened 2023-01-17
Feat/fallback component

rustem-nasyrov:
Partially #2663

## Description
- Reworked fallback behavior and created separate component for it.

## Types of changes


rustem-nasyrov:
@aluarius, @m0ksem, please take a look at the preset and vuestic configuration behavior.

packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -73,16 +84,6 @@ export default defineComponent({
 
     const hasLoadError = ref(false)
 
-    const srcComputed = computed(() => {
-      if (props.src && props.fallbackSrc && hasLoadError.value) {
-        emit('fallback')
```
m0ksem, line 78: Might be helpful to emit fallback event on image, so user can react if image wasn't loaded.

packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -13,11 +13,12 @@
...
     <slot v-bind="avatarOptions" v-else>
       <img
-        v-if="srcComputed"
-        :src="srcComputed"
+        v-if="$props.src && !hasLoadError"
+        :src="$props.src"
         :alt="$props.alt"
         @error="onLoadError"
       >
+      <va-fallback v-else-if="hasLoadError && $props.src" v-bind="VaFallbackProps" />
```
m0ksem, line 21: Would be nice to have a slot for fallback maybe.

m0ksem (approved):
In my opinion, these props are good for global configuration, but if user wants to configure some specific image to have a fallback it is much easier to provide a slot. But it is not a big deal.

aluarius (approved):
We won't make a documentation page for this component?

rustem-nasyrov:
> We won't make a documentation page for this component?

No, it's not that necessary component.

PR #2873, opened 2023-01-18
fix: component unit tests

### LINKED ISSUE #1145: Fix components unit tests
Component unit tests are disabled.
We need to include tests.

LighthouseKeeperYN:
close #1145

I restored and added missing component unit tests, so far 90% of tests only check if component was mounted without an error. I commented out some tests cases I couldn't get to work, it's probably easier to write new ones.

## Types of changes


rustem-nasyrov (approved):

m0ksem (approved):
Looks like a good start :) We gonna need to fix components typing in tests and make a better coverage.

PR #2876, opened 2023-01-18
fix: build vuestic from sandbox if it doesn't exist

aluarius:
## Description
If `ui/dist` doesn't exist, `yarn build:analysis` command at the `sandbox` workspace fails. Let's handle this situation.

## Types of changes

### DIFF
--- a/packages/sandbox/scripts/bundle-analysis.ts
+++ b/packages/sandbox/scripts/bundle-analysis.ts
@@ -24,6 +24,10 @@ import * as process from 'process'
   }
 
   if (forceRebuild || !existsSync(bundleSizesCachePath)) {
+    if (!existsSync('../ui/dist')) {
+      await $('yarn workspace vuestic-ui build', {})
+    }
+
     await Promise.all([
       $('vite build --config ./configs/vite/vite.empty.ts', { successMessage: 'Empty config was built.' }),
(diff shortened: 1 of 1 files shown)


m0ksem (commented):
I'm thinking about to generate "Run build analisys" md, instead of making vuestic-ui build and build analisys table. Cuz, for now we force all contributors to generate this table, but no one really need it.

aluarius:
Now we have to ways to build analysis md:

1. Vuestic UI -> Bundle Analysis -> Md for docs
2. Cache -> Md for docs

And both of them need to build Vuestic and test bundles at least one time.

You don't want to keep cache in git, so, i can't see any other options.

m0ksem:
> Now we have two ways to build analysis md:
> 
>     1. Vuestic UI -> Bundle Analysis -> Md for docs
> 
>     2. Cache -> Md for docs
> 
> 
> And both of them need to build Vuestic and test bundles at least one time.
> 
> You don't want to keep cache in git, so, i can't see any other options.

How about: if there is no cache and it dev - make an empty file that contains "In dev there is no data about bundle size, please run: bla-bla-bla". Then we should have this in docs instead of table. 99% of time working with docs in dev no one will really care about this bundle sizes. Same warning can be added on top of table in dev "This sizes used from cache, please run to get new value and restart docs: bla-bla-bla". Make our life easier.

m0ksem (approved):

Issue #2228, opened 2022-08-12
`VaAutocomplete` component


aluarius:
Ref: https://element-plus.org/en-US/component/autocomplete.html

Do some research on which of the implementation ways we want to take
* inline autocomplete (hints you text in input) vs dropdown suggestion (probably dropdown suggestion)
* separate component vs implementing on top of select (we though using select could be a better idea)

## Potential complications
* Search select has a separate search field.
<img width="292" alt="image" src="https://user-images.githubusercontent.com/5394573/189095242-e98ecdb3-b54d-4277-9d2c-85476cd1c0e7.png">

PR #2909, opened 2023-02-01
[#2228] autocomplete functional for va-select

aluarius:
Closes: #2228
Closes: #2913

## Description
- [x] Autocomplete functionality for `va-select`.
- [x] A few new `va-select` features, that are suitable not only for `autocomplete` mode. 
- [x] Restored loading state for `va-select`. 
- [x] Docs and demos updates. 

## Types of changes


aluarius:
Deep QA is really what i need.

rustem-nasyrov:
I suggest to use primary color as highlight color by default, because current version is barely visible.

image

packages/ui/src/components/va-select/components/VaSelectContent/VaSelectContent.vue
```diff
@@ -74,16 +118,84 @@ export default defineComponent({
...
+        } else if (newAutocompleteValue !== oldAutocompleteValue && autocompleteInputValue.value !== newValueStringConverted) {
+          // autocomplete input handling
+          emit('autocomplete-input', newAutocompleteValue)
+        }
+
+        if (!newDropdownShownState && oldDropdownShownState && newValueStringConverted !== autocompleteInputValue.value) {
+          // handling input wrapper blur, not for multiple state
+          autocompleteInputValue.value = props.multiple ? '' : newValueStringConverted
+        }
+      })
```
m0ksem, line 180:
```suggestion
    watch(value, (newVal, oldVal) => {
      if (props.multiple) {
        autocompleteInputValue.value = ''
        return
      }

      if (typeof newVal === 'undefined') {
        autocompleteInputValue.value = ''
        return
      }

      autocompleteInputValue.value = convertToString(newVal as string)
    })

    watch(autocompleteInputValue, (newVal, oldVal) => {
      if (props.autocomplete) {
        emit('autocomplete-input', newVal)
      }
    })
```

This looks good for me. Am I missing some behavior here?

I changed value prop be actually value, but not visible selectOption. Idk why it was wrong.

VaSelect.vue
```ts
    const selectContentPropsComputed = computed(() => ({
      ...pick(props, ['placeholder', 'autocomplete', 'multiple', 'disabled']),
      tabindex: tabIndexComputed.value,
      value: valueComputed.value, // <--
      valueString: valueComputedString.value,
      hiddenSelectedOptionsAmount: hiddenSelectedOptionsAmount.value,
      isAllOptionsShown: isAllOptionsShown.value,
      dropdownContentShown: showDropdownContent.value,
      focused: isInputFocused.value,
      getText,
    }))
```
firefox_KZXCMLrqlQ

The last part to make is ask Select to clear autoCompleteInput text. With code above you can remove `dropdownContentShown` and `valueString` and pass instead `searchText`. Which will be cleared in VaSelect component after option is selected. 

I would like to see components where most of state is handled in that component and children components just used to show this state.

m0ksem: Also, highlight color sucks. Should be primary with some opacity I guess.

m0ksem (changes requested):

packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -254,9 +257,14 @@ export default defineComponent({
     }
 
     watch(() => props.hoveredOption, (newOption: SelectOption | null) => {
-      (!lastInteractionSource.value || lastInteractionSource.value === 'keyboard') && (isValueExists(newOption)) && scrollToOption(newOption!)
+      (!lastInteractionSource.value || lastInteractionSource.value === 'keyboard') &&
+      (isValueExists(newOption)) && scrollToOption(newOption!)
```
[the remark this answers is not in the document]
aluarius: Very useful! Think should be included to the tech talks.

aluarius:
> Missing translation?

No, it's inner Vuestic UI i118 variable and it doesn't supported by docs -> shown `as is`. The same with other cases (for example, `va-file-upload`).

aluarius:
> I've ended typing and selected the option. The remaining input text confused me, maybe we somehow visualize it?

I've decided to clear it after select.

m0ksem (changes requested):
Demo doesn't work for me...

packages/ui/src/components/va-select/hooks/useAutocomplete.ts
```diff
@@ -0,0 +1,57 @@
...
+type UseAutocompleteProps = ExtractPropTypes<typeof useAutocompleteProps> & { multiple: boolean }
+
+export const useAutocomplete = (
+  props: UseAutocompleteProps,
+  value: Ref<SelectOption[]>,
+  valueString: Ref<string>,
+  dropdownShown: Ref<boolean>,
+  getText: (option: SelectOption) => string,
+) => {
+  const getLastOptionText = (v: SelectOption[] | string | undefined) => {
```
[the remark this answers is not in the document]
aluarius: Yea, you're right. It's a redundant part since this function was used for valueString as well.

packages/ui/src/components/va-select/hooks/useAutocomplete.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (newValue && newValue !== getLastOptionText(value.value)) {
+      dropdownShown.value = true
+    }
+  })
+
+  watch(dropdownShown, (newValue, oldValue) => {
+    if (!props.autocomplete || newValue || !oldValue) { return }
+
+    autocompleteValue.value = props.multiple ? '' : String(valueString.value)
+  })
```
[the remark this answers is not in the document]
aluarius: > Instead of valueString?

No, if we don't want to duplicate transformation login for `value -> valueString`. Yes if it's acceptable. Up to you.

packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -652,8 +687,8 @@ export default defineComponent({
       focusPreviousOption,
       focusNextOption,
       showDropdownContentComputed,
-      showDropdown,
-      hideDropdown,
+      handleDropdownOpen,
+      handleDropdownClose,
```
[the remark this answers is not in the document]
aluarius: :D

Actually, i tried to handle `open/close-dropdown` events from `VaSelectContent` and names of these two methods were conflicting. After i rejected this idea, i've decided to keep new names, because they're more informative for me - not just open, but proceeding additional related logic.

packages/ui/src/components/va-select/hooks/useAutocomplete.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const setAutocompleteValue = (v: string) => (autocompleteValue.value = v)
```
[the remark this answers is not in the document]
aluarius: Ok, i'll move it to `VaSelect`, but i can't agree, that it's not connected to the composable - lot's of our composables expose methods like this - `useHover`, for example. It exposes methods `onMouseEnter`/`onMouseLeave` that serve as handlers for parent and changing internal component's state. Same as here.

packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -271,17 +284,7 @@ export default defineComponent({
...
-
-      if (typeof valueComputed.value === 'string' || typeof valueComputed.value === 'number') { return valueComputed.value }
-
-      if (Array.isArray(valueComputed.value)) {
-        return visibleSelectedOptions.value.map((value) => getText(value)).join(props.separator) || props.clearValue
-      }
-
-      return getText(valueComputed.value)
-    })
+    const valueString = useStringValue(props, valueComputed, visibleSelectedOptions, getText)
```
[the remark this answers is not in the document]
aluarius: I've reimplemented this part in another way.

packages/ui/src/components/va-select/hooks/useAutocomplete.ts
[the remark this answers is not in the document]
m0ksem: These functions are native event handlers, so they are essential for compostable to work.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
+        :options="defaultMultiSelect.options"
+        :max-visible-options="2"
+        multiple
+        autocomplete
```
asvae, line 54: image

I'm expecting text cursor to move left and right when I press respective buttons.
Instead list cursor moves.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
+        :options="defaultMultiSelect.options"
+        :max-visible-options="2"
+        multiple
```
asvae, line 53: Not related to autocomplete, but can't we make "Items not found" formatting a bit better (should occupy the same space as list item)

image

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
+        :options="defaultMultiSelect.options"
+        :max-visible-options="2"
```
asvae, line 52: image

In autocomplete I'm typing "two", then press "space" button. I expect autocomplete to add space. Item is selected instead.

(valid case for autocomplete are words with space in them)

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
+        :options="defaultMultiSelect.options"
```
asvae, line 51: image

I'm expecting select to do nothing on "Space". Right now it behaves the same as on "Enter".

Up to you to investigate deeper, but I checked 3 frameworks and they had the same behavior. I also don't see, how we can make all functionality work without relying on "space"

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      <p>Value: {{ defaultMultiSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
```
asvae, line 50: image

For autocomplete - if value is already selected, and I click on autocomplete - I expect search text to be selected like this:

image

This would allow me to search for another value.

Instead - only cursor is being placed.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      />
+      <p>Value: {{ defaultMultiSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
```
asvae, line 49: image

When you type - text jumps around a bit due to selection padding. Not very critical.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+        highlight-matched-text
+      />
+      <p>Value: {{ defaultMultiSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
```
asvae, line 48: image

When I click on chevron - select is getting open for autocomplete.
When I click on other parts - it's not getting open.

I believe that behaviour should be consistent (same as with normal select). Chevron is meant to make select apart from input, not as a functional element.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+        autocomplete
+        highlight-matched-text
+      />
+      <p>Value: {{ defaultMultiSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
+      <va-select
```
asvae, line 47: image

When I click on autocomplete - it's not getting open.
I expect it to open.

I believe this could be managed with a prop, but by default it probably should get open on click.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+        multiple
+        autocomplete
+        highlight-matched-text
+      />
+      <p>Value: {{ defaultMultiSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete with max visible options"
+      style="width: 400px;"
+    >
```
asvae, line 46: image

When I type `tw`, then click outside of select - it still retains focus:

image

I expect to lose focus.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      <p>Value: {{ defaultSingleSelect.value }}</p>
+    </VbCard>
+    <VbCard
+      title="Multiple autocomplete"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
```
asvae, line 35: Let's remove these placeholders from demos. You can focus placeholder testing on few demos you want. Otherwise they just add confusion.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,107 @@
...
+      title="Autocomplete"
+      style="width: 400px;"
+    >
+      <va-select
+        v-model="defaultSingleSelect.value"
+        class="mb-6"
+        placeholder="Start to write..."
+        :options="defaultSingleSelect.options"
+        autocomplete
+        highlight-matched-text
```
asvae, line 24: highlight-matched-text could probably be true by default

packages/docs/src/page-configs/ui-elements/select/page-config.ts
```diff
@@ -80,6 +80,11 @@ const config: ApiDocsBlock[] = [
     'select.examples.selectedTopShown.text',
     'SelectedTopShown',
   ),
+  ...block.exampleBlock(
+    'select.examples.autocomplete.title',
+    'select.examples.autocomplete.text',
+    'Autocomplete',
```
asvae, line 86: I suggest we add an example with item create. That's a very common case.

Also a demo with relevant event.

asvae (changes requested):
Thanks!

I found a couple of issues with current select, which should be enough for one round of fixes.

Probably one more back and forth would be needed.

packages/ui/src/components/va-select/VaSelect.demo.vue
aluarius: It's base VaSelect behaviour: when dropdown closes, input gets the focus. I'll make an exception for autocomplete mode, but still have some doubts about this.

asvae: > It's base VaSelect behaviour

It was an invalid behaviour then. I checked refs and select always loses focus on click outside.

aluarius: Okey, so i'll make it default for the whole VaSelect, not only for autocomplete.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,109 @@
...
+    >
+      <va-select
+        v-model="defaultMultiSelect.value"
+        class="mb-6"
+        :options="defaultMultiSelect.options"
+        :max-visible-options="2"
+        multiple
+        autocomplete
+      />
+      <p>Value: {{ defaultMultiSelect.value }}</p>
```
asvae, line 51: image

When I press enter with this state - "one" is being selected.

That's definitely not what should happen.

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,109 @@
...
+    </VbCard>
+    <VbCard
```
asvae, line 53: In autocomplete with new option - I can press Enter 2 times - then 2 items with same name are being created

image

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,109 @@
...
+      title="Autocomplete with create new option"
+      style="width: 400px;"
```
asvae, line 55: Can't we have ajax autocomplete example? (can be moved to another issue and out of this release if hard to do)

packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -11,6 +11,109 @@
...
+        v-model="defaultSingleSelect.value"
+        class="mb-6"
+        :options="defaultSingleSelect.options"
+        autocomplete
+        auto-select-first-option
+      />
+      <p>Value: {{ defaultSingleSelect.value }}</p>
+    </VbCard>
     <VbCard
       title="Selected options at the top"
```
asvae, line 118: image

I this case I would keep cursor at the same position it was and not jump with selected item. But please investigate, I might be mistaken. :)

packages/ui/src/components/va-select/VaSelect.demo.vue
aluarius: This problem was already solved by optional `autoSelectFirstOption` prop. Probably,  i'll make it default for autocomplete mode.

packages/ui/src/components/va-select/VaSelect.demo.vue
aluarius: It's base VaSelect behaviour for non-unique `allow-create` mode. But i'll handle it for autocomplete.

packages/ui/src/components/va-select/VaSelect.demo.vue
aluarius: I'd prefer to move it to the separate issue, because it's not related to autocomplete and affects the whole component's logic.

packages/ui/src/components/va-select/VaSelect.demo.vue
aluarius: Also moved to separate issue.

PR #2918, opened 2023-02-03
Feat/skeleton

m0ksem:
image

Docs later in nuxt docs. Review component API (not sure how to make a good skeleton)


rustem-nasyrov:
image
Missing translations or we planning to move at the next week to the nuxt docs?

packages/ui/src/components/va-skeleton/VaSkeleton.vue
```diff
@@ -0,0 +1,210 @@
...
+
+  emits: [/* Define emits here */],
+
+  setup (props, { attrs }) {
+    const doShow = ref(false)
+
+    onMounted(() => {
+      setTimeout(() => {
+        doShow.value = true
+      }, props.delay)
```
rustem-nasyrov, line 43: Shouldn't we clear timeouts on unmounted hook?

packages/ui/src/components/va-skeleton/VaSkeleton.vue
```diff
@@ -0,0 +1,210 @@
...
+      pulse: props.animation === 'pulse',
+      wave: props.animation === 'wave',
+    }))
+
+    const borderRadius = computed(() => {
+      if (props.variant === 'circle') { return '50%' }
+      if (props.variant === 'rounded') { return `calc(${heightComputed.value} / 5)` }
+
+      return '0px'
+    })
```
rustem-nasyrov, line 85: I'll suggest to use css variables here.

rustem-nasyrov (commented):
Looks good!

packages/ui/src/components/va-skeleton/VaSkeleton.vue
m0ksem: There are sometimes when you want ot have rounded button and squared image for example. It will be a pain to deal with CSS variables for each individual case.

packages/ui/src/components/va-skeleton/VaSkeleton.vue
```diff
@@ -0,0 +1,210 @@
...
+    lines: { type: Number, default: 1 },
+    height: { type: [Number, String], default: '1em' },
+    width: { type: [String], default: '100%' },
+    lineGap: { type: String, default: '8px' },
+    textWidth: { type: [String], default: '75%' },
+
+    variant: { type: String as PropType<'text' | 'circle' | 'rounded' | 'squared'>, default: 'squared' },
+  },
+
+  emits: [/* Define emits here */],
```
aluarius, line 35: Forget to delete?

packages/ui/src/components/va-skeleton/components/VaSkeletonGroup.vue
```diff
@@ -0,0 +1,67 @@
...
+
+<script lang="ts">
+import { defineComponent, PropType, computed, ref, onMounted } from 'vue'
+import { useBem } from '../../../composables'
+import { VaConfig } from '../../va-config'
+
+export default defineComponent({
+  name: 'VaSkeletonGroup',
+
+  inheritAttrs: true,
```
aluarius, line 17: It's `true` by default, no?

packages/ui/src/components/va-skeleton/VaSkeleton.vue
asvae: Probably not, will hold component garbage collection for a bit, but given you don't set delay too high - that shouldn't matter at all.

rustem-nasyrov (changes requested):
We still don't have translations on the docs page.

packages/ui/src/components/va-skeleton/VaSkeleton.demo.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <VaContent>
+    <VbDemo>
+      <VbCard title="Headline" style="min-width: 200px;">
+        <VaSkeleton tag="h1" variant="text" class="va-h1" />
```
asvae, line 5: Can we focus on props in prop demos?

What I'd hope to see is:

* no props (default) select
* demos for variants with minimal amount of additional props/classes
* more rich examples can go to docs, or you can leave one here
* skeleton group could be probably done in different file, given it's a different component
* skeleton group should ideally display component functionality, not idea, so using card component is not needed - it makes it harder to understand what component is doing.

packages/ui/src/components/va-skeleton/VaSkeleton.vue
```diff
@@ -0,0 +1,210 @@
...
+      doShow,
+      heightComputed,
+      widthComputed,
+      borderRadius,
+    }
+  },
+})
+</script>
+
+<style lang="scss" scoped>
```
asvae, line 104: is "scoped" intended here?

m0ksem:
Updated demos and docs.
image

m0ksem:
@aluarius, please check if accessibility is okay :)

rustem-nasyrov (approved):
Great job!

packages/ui/src/components/va-skeleton/VaSkeleton.vue
```diff
@@ -0,0 +1,215 @@
+<template>
+  <component
+    class="va-skeleton"
+    role="status"
+    aria-busy="true"
```
aluarius, line 5: If VaSkeleton is show via `v-if`, `aria-busy` and `aria-atomic` are redundant imo.

aluarius (commented):
- No `slot` description in docs.
- Will be fine to have real example of loading content using `VaSkeleton` (via `setTimeout` for example).

packages/ui/src/components/va-skeleton/VaSkeleton.vue
m0ksem: It shows with some delay, so if you just refresh page it doesn't blink and shows only if something loading for a very long time (>300ms), but still yet, the component itself represent loading and if page is loading 700ms, skeleton will be shown with busy state for a 400ms.

m0ksem: I think it might be better to render this element, so it is busy anytime even before delay. We can visually hide it so it is no blinking, but still present in DOM as loading indicator. That's makes sense for me.

m0ksem:
Slots are not used here in VaSekeleton, so they're not in docs.

PR #2923, opened 2023-02-06
Low elements amount bugfix and auto wrapper size mode for `VaVirtualScroller`

aluarius:
Closes: #2905

## Description
- [x] fixes bug with low amount of items in the list;
- [x] added `auto` mode for `wrapper-size` prop;
- [x] demos updates.  

## Types of changes


m0ksem (changes requested):
Kapture 2023-02-11 at 00 10 23

Everything is broken for me. 
Set a big array. But shows only 100, 600 (bench 300) for some reason.
image

aluarius:
It handles only less than 700 000 items lists. More - arithmetical issues (mb first time in my life i should use bigInt). But imo it makes no sense to try to handle such huge lists.

m0ksem (approved):
Make sense.

PR #2928, opened 2023-02-07
Re-implement nuxt docs

asvae:
There are some issues with current nuxt-docs implementation which make me very scared to merge it into `develop` codebase.

So far I moved docs work to `nuxt-docs` branch in `upstream` and I hope we can polish it there until it's all right to bring back to develop.

What I've noticed so far:
* **We have 2 folders**: `docs` and `nuxt-docs`. That alone is already significant enough to postpone merge. Having 2 projects that have about the same functionality is one truck of technical debt. We're guaranteed to have issues with that many months ahead. It's also impossible to just replace `docs` with `nuxt-docs` as there are many things missing from `nuxt-docs`, so this task requires some commitment.
* **With auto import navigation doesn't work in Webstorm** (I'm not sure about VS Code). That kills workflow for me. So we either have to move back to full imports or a way for Webstorm and VS Code to work with current approach.

With both issues the problem is that it's way harder to deal with consequences than to prevent them from happening in a first place. That makes me believe that keeping this out of `develop` before fix is an optimal approach.

Smaller issues:
* `yarn build` + `node .output/server/index.mjs`
  * scrolling down one one page, then switching to another briefly shows page at scroll location then goes to beginning of the page -> should do that immediately


rustem-nasyrov:
Issues that I found:

- [ ] Long loading between docs pages
- [ ] Counter: examples alignments. Different sizes of buttons and input in the component itself.
- [ ] Rating: solid backgrounded numbers colors are too dark
- [ ] Slider: vertical example doesn’t fit in the example card
- [ ] Switch: align examples
- [ ] Time input: validation icons too far
- [ ] Card: align examples
- [ ] Color classes: fix default example
- [ ] Mobile menu: sidebar is open be default, state doesn’t work
- [ ] Kitchensink: Examples doesn’t fit on the mobile emulator
- [ ] Mobile colors page: color card examples doesn’t fit
- [ ] Component sandbox: mobile design

m0ksem:
In this PR also a lot of fixes relative ui lib in nuxt environment or usage without our css styles (modular components). Would be nice to have these changes in v1.6.

m0ksem:
> With auto import navigation doesn't work in Webstorm (I'm not sure about VS Code). That kills workflow for me. So we either have to move back to full imports or a way for Webstorm and VS Code to work with current approach.

Sadly, there is no way for now in ts to set autocomplete behavior. Without autoimport in example block we gonna need to import both raw and component imports which is a mess. So I would wait for an opportunity to create autocomplete options. Let's stay with the current solution and fix it later.

>  Long loading between docs pages

All pages loaded dynamically, which improves first load. 

---

I would stick to limitation from above, then losing SSR support and improvements in lighthouse. We also able to test VuesticUI in Nuxt environment and as you can see there are lot's of issues because of SSR.

m0ksem:
Missing page: 
- Aspect Ration, 
- Tailwind
- missing Alerts about 1.6.0
- SmartHelpers relay on grid.scss
- CSS variables lost \n in code.
- Slider lost first circle in SSR.

PR #2950, opened 2023-02-10
fix: add CleaveOptions type to mask prop to match real use

asbermudez:
## Description
The mask property from the useCleave() hook had the missing CleaveOptions type which is a valid input type for it

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/VaInput.demo.vue
+++ b/packages/ui/src/components/va-input/VaInput.demo.vue
@@ -558,6 +558,18 @@
         mask="date"
       />
       Value: {{ maskReturnFormattedValue }}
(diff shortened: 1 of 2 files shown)


asbermudez:
@m0ksem sorry, I accidentally removed it by removing my fork...

m0ksem:
Hi, @asbermudez. I changed a bit this PR. Can you check if it solves your particular issue using VaInput.demo.vue? I tried with phone and it's OK. Want to double-check this.

m0ksem (approved):

asbermudez:
@m0ksem Works like a charm as far as I can see :)

PR #2973, opened 2023-02-15
Feat/ag grid

### LINKED ISSUE #1200: Bring back ag-grid demos
Ag grid theme demos were deleted by mistake, we need to bring them back.

Here's the demo:
https://github.com/epicmaxco/vuestic-ui/commit/11fec1b5dd47d8129e20ac473c8b3bdc0882f99c#diff-e601272cac9c6c619d39367631d9bba3fab55b07e65c1a284ef9628206d6dacc

### LINKED ISSUE #2300: Ag-grid theme extension should use css instead of scss
Some projects might not use scss at all, so we ideally do not want to force loader on users.

Also see [AG Grid deprecation of sass](https://www.ag-grid.com/angular-data-grid/global-style-upgrading-to-v28-sass/).

We might also want to check if we can use some of their CSS variables.

Another reason is that vite doesn't support `~` alias, so `@import "~@ag-grid-community/core/dist/styles/ag-theme-base/sass/ag-theme-base-default-params"` gives an error unless you add the following to vite config:

```typescript
resolve: {
  alias: {
    '~@ag-grid-community': require('path').resolve(__dirname, 'node_modules/@ag-grid-community')
  }
}
```

With precompiled CSS - that's not a problem.

### LINKED ISSUE #2992: Missing AG Grid page

rustem-nasyrov:
Close #2992
Close #2300 
Close #1200

image

## Description
- Added latest ag-grid version support.
- Added css build for non-scss users.
- Added demo file.

## Types of changes


m0ksem (approved):
I updated docs and package.json exports.

PR #3016, opened 2023-02-17
Docs/remove old new badge

m0ksem:
Please double-check if I removed some important badge...


rustem-nasyrov (commented):
We updated `va-image` component.

aluarius (approved):
`smart-helpers` already were in the previous release.

PR #3019, opened 2023-02-18
Fix/file loading

m0ksem:
Fix minified file render in production

Added CSS Variables table to block.api

image

### DIFF
--- a/packages/docs/modules/page-config/blocks/api/index.vue
+++ b/packages/docs/modules/page-config/blocks/api/index.vue
@@ -4,17 +4,24 @@ import { parseComponent } from './component-parser'
 import merge from 'lodash/merge'
 import camelCase from 'lodash/camelCase'
 import ApiTable from './components/api-table.vue';
-import { ManualApiOptions } from './types';
+import { CssVariables, ManualApiOptions } from './types';
 
 const props = defineProps({
   componentName: {
     type: String,
+    required: true,
   },
   component: {
-    type: Object as PropType<DefineComponent>
+    type: Object as PropType<DefineComponent>,
+    required: true,
   },
   manual: {
-    type: Object as PropType<ManualApiOptions>
+    type: Object as PropType<ManualApiOptions>,
+    default: () => {},
+  },
+  cssVariables: {
+    type: Array as PropType<CssVariables>,
+    required: true,
   }
 })
 
@@ -75,23 +82,28 @@ const eventsOptions = Object
   .entries(withManual.events || {})
   .filter(([key, prop]) => !prop.hidden)
   .map(([key, prop]) => ({
-    name: { name: key, ...prop },
+    name: key,
     description: t(getTranslation('events', key)),
   }))
 
 const slotsOptions = Object
   .entries(withManual.slots || {})
   .map(([key, prop]) => ({
-    name: { name: key, ...prop },
+    name: key,
     description: t(getTranslation('slots', key)),
   }))
 
 const methodsOptions = Object
   .entries(withManual.methods || {})
   .map(([key, prop]) => ({
-    name: { name: key, ...prop },
+    name: key,
     description: t(getTranslation('methods', key)),
   }))
+
+const cssVariablesOptions = props.cssVariables.map(([name, value, comment]) => ({
+  name, 'Default Value': '`' + value + '`', /* comment */ // TODO: Enable comment when everywhere is used correct comments
+  // TODO: Or add tanslations after i18n splitted
+}))
 </script>
 
 <template>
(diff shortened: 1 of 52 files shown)


m0ksem:
For VaInputWrapper need to move variables in VaInput variables or create a way to show only css varialbes in block.api.

For example,

```ts
block.api('VaInputWrapper', { props: false, methods: false, events: false })
```

packages/docs/modules/page-config/blocks/api/index.vue
```diff
@@ -75,23 +82,28 @@ const eventsOptions = Object
   .entries(withManual.events || {})
   .filter(([key, prop]) => !prop.hidden)
   .map(([key, prop]) => ({
-    name: { name: key, ...prop },
+    name: key,
     description: t(getTranslation('events', key)),
   }))
```
rustem-nasyrov, line 87: Why not to use `Array.prototype.reduce()` though? Instead of iterating over array two times, we do loop once (here and above).

packages/docs/modules/page-config/blocks/api/transform.ts
```diff
@@ -1,4 +1,32 @@
...
+  lines.forEach((line, index, lines) => {
+    const variable = (line.match(/(--va-.*)?:(.*)?;/) || []).slice(1)
+
+    if (variable.length) {
+      const comment = parseCssComment(lines[index - 1] || '')
+      variables.push([variable[0], variable[1], comment])
+    }
+  })
+  
+  return variables
```
rustem-nasyrov, line 23: Also here can be used `reduce()`. 😄

packages/docs/page-config/ui-elements/image/index.ts
```diff
@@ -22,7 +22,6 @@ export default definePageConfig({
     block.subtitle("all.api"),
     block.api("VaImage", apiOptions),
 
-    block.subtitle("all.cssVariables"),
-    block.file("vuestic-ui/src/components/va-image/_variables.scss"),
+    block.subtitle("all.cssVariables"),
```
rustem-nasyrov, line 25: Duplication.

packages/docs/page-config/ui-elements/input/index.ts
```diff
@@ -26,9 +26,8 @@ export default definePageConfig({
     block.subtitle("all.api"),
     block.api("VaInput", apiOptions),
 
-    block.subtitle("all.cssVariables"),
-    block.paragraph("VaInput:"),
-    block.file("vuestic-ui/src/components/va-input/_variables.scss"),
+        block.paragraph("VaInput:"),
```
rustem-nasyrov, line 29: image

packages/docs/page-config/ui-elements/modal/index.ts
```diff
@@ -35,7 +35,6 @@ export default definePageConfig({
     block.api("VaModal", apiOptions),
 
     block.subtitle("all.cssVariables"),
```
rustem-nasyrov, line 37: Duplicates.

rustem-nasyrov (changes requested):
Missing css rules at the inner loading component docs.
<details>
image
</details>

packages/docs/modules/page-config/blocks/api/index.vue
m0ksem: I would prefer to have readable code instead of reduce. In my opinion reduce must be used only for calculation in array, not to create new one.

rustem-nasyrov:
> Missing css rules at the inner loading component docs.

Still empty name column. 
image

m0ksem:
@rustem-nasyrov, double layout category fixed in https://github.com/epicmaxco/vuestic-ui/commit/82cbd44a415da9edf712fe2482ce4bd94cc07377 develop

rustem-nasyrov (approved):

m0ksem:
VaInnerLoading had wrong css variables name, so I changed them. Looks like breaking change, but I believe no one cares about it...

packages/docs/page-config/ui-elements/toast/toast-api.ts
```diff
@@ -50,7 +47,7 @@ const optionsTableData: [keyof ToastOptions, string][] = [
   ["color", "string"],
 ];
 
-export const optionsApi = (block: PageGenerationHelper) =>
+export const optionsApi = (block: any) =>
```
aluarius, line 50: No way to type this?

packages/docs/modules/page-config/blocks/file/transform.ts
```diff
@@ -1,13 +1,15 @@
...
 import { extname } from 'path'
+import { readFile } from 'fs/promises'
 
 export default defineBlockTransform(async function (block) {
   if (block.type !== 'file') { return }
 
   const importName = block.args[0].slice(1, -1)
   const importPath = (await this.importer.resolveAbsolutePath(`${importName}`))!
-  const importComponent = this.importer.importDefault('file', `${importPath}?raw`)
+  const content = (await readFile(importPath, 'utf-8')).toString()
```
aluarius, line 10: Add at least `?` before `toString` or any typo will lead to page down.

aluarius (approved):

packages/docs/modules/page-config/blocks/file/transform.ts
m0ksem: I'll throw human-readable error. It's okay to have Error in transform functions, so vite can handle files correctly.

Issue #2894, opened 2023-01-25
VaInput doesn't expose focus method


m0ksem:
Looks like it is just returned in setup state. We must use `expose()` method. The same might be with Select and other components.

Otherwise, for now, it is impossible to call 

```ts
const input = ref<typeof VaInput>()
input.focus()
// only input.setupState.focus() works.
```

PR #3021, opened 2023-02-18
fix(#2894): dispatch focus event when focus

m0ksem:
- [ ] Need to double check other components


aluarius (changes requested):
image

Found while testing VaSelect.

PR #3034, opened 2023-02-21
ui: tmp fix buttons

asvae:
Before: 
image

After:

image

Needs a follow up on buttons. And decision if we want to have margins set.


m0ksem:
I would completely remove margin (with css variable). Makes no sense for me.

asvae:
Yeah, we also don't use it in other components.

m0ksem (approved):

Issue #3046, opened 2023-02-22
SSR does not add component variables


asbermudez:
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
Use VaAvatar component with SSR with Nuxt, although it happens with several components.

**What is the expected behavior?**
Should render the necessary visual variables by default:

![image](https://user-images.githubusercontent.com/46224115/220625711-3415a90d-57ab-4b25-b27f-76da9be46b35.png)

![image](https://user-images.githubusercontent.com/46224115/220625366-5af82802-eb97-4a82-9920-f5a955eaa168.png)

**What is the current behavior?**
It does not render the variables and makes it look broken while it hydrates

![image](https://user-images.githubusercontent.com/46224115/220625868-1a193c15-6663-4d62-bf1c-95c9b01c1148.png)

![image](https://user-images.githubusercontent.com/46224115/220625514-6984bad3-866a-4522-8584-32e9f1382e7d.png)

**Other information**
This was not happening in version 1.5.x, for what I can see then we were not using CSS variables for some values (like size) which seems to be the source of the issue as they don't get added until hydration.

asbermudez:
Want to point out, that this was happening with few components that used the same technique for handling things like colors in version `1.5.x`, like the **VaButtonDropdown** using color "danger".

![image](https://user-images.githubusercontent.com/46224115/220628703-5e45ec2e-8eaa-4239-bbf5-365f4d93e6e3.png)

m0ksem:
Scope: This is related to Vue's `v-bind` in CSS. For some reason, some CSS variables are not rendered in Nuxt SSR. In VaButton for example, `backgroundColor` renders, but not `backgroundColorOpacity`. We have this issue on docs landing page.

Possible solutions:
- Use `:style="{}"` instead of `v-bind`.

m0ksem:
@asbermudez, what nuxt modules do you use?

asbermudez:
@m0ksem this is the output of `nuxi info`

```bash
Nuxt project info:                                                          09:49:59

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v18.14.0`
- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.2`
- Package Manager:  `npm@9.3.1`
- Builder:          `vite`
- User Config:      `app`, `modules`, `postcss`, `vite`, `typescript`, `runtimeConfig`, `imports`, `umami`, `auth`, `nitro`
[...]

PR #3051, opened 2023-02-23
fix(#3046): remove vuestic from excluded optimizeDeps

m0ksem:
image

Hell yeah, found it!

Maybe it is a solution for https://github.com/epicmaxco/vuestic-ui/issues/3046, but it is not directly fixes it. I'll double-check it in new nuxt project.


aluarius (approved):

m0ksem:
This doesn't fix #3046

PR #3052, opened 2023-02-23
fix: minor va-image fix

aluarius:
Have found a few problems while checking Vuestic Admin.

## Description
- CSS variables renamed, added missing `:root`.
- Content width minor fix.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-image/_variables.scss
+++ b/packages/ui/src/components/va-image/_variables.scss
@@ -1,5 +1,7 @@
 :root,
 :host {
+  // the relative position is used to render va-image inside va-viewer
+  // can also be suitable if you want to leave size and position control to external classes
   --va-image-content-position: absolute;
   --va-image-content-img-object-position: 50% 50%;
 }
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/va-image/_variables.scss
```diff
@@ -1,4 +1,6 @@
+:root,
 :host {
-  --va-image-position: absolute;
-  --va-image-object-position: 50% 50%;
+  --va-image-content-position: absolute;
```
m0ksem, line 3: Any reason for this to be a CSS variable? I can't see any cases when user is going to change it to `relative`. I would remove all css variables from here.

m0ksem (changes requested):

packages/ui/src/components/va-image/_variables.scss
aluarius: image

`content-position` is also a real case.

But i can agree that `content-width` is redundant.

m0ksem: So there are cases when user want content position to be relative? Can you add a comment on top of css variable? We will show this later in docs.

m0ksem (changes requested):

m0ksem (approved):

PR #3054, opened 2023-02-23
Feature/expandable row

### LINKED ISSUE #2837: VaDataTable expandable row
Is there a way to have additional information per row in a data table? I don't see a special slot for that in the Vuestic data table.

Something similar to expandable rows in Vuetify: https://vuetifyjs.com/en/components/data-tables/#expandable-rows
Or row details in BootstrapVue: https://bootstrap-vue.org/docs/components/table#row-details-support

Is there a manual workaround for appending a special `tr` to a `va-data-table` to accomplish that, or it requires a slot?

LighthouseKeeperYN:
close #2837

Adds expandable row for data table.

Breaking changes: cell slot bindings were changed from `cell` to `{ cell, row }`. `row `  property is necessary to access expandable row trigger method.

## Types of changes

image


m0ksem (changes requested):
Looks cool. 

Found few issues: 
- No collapse animation (expand animation looks good)
- Missing --expanded classes. It is a normal practice to add some background to an expanded row, people must have classes for it. For example:
image

aluarius (changes requested):
Smthng happens for `VaDataTable.new.demo.vue`. It worked before, so, something goes wrong.

image

m0ksem (approved):
I changed demo a bit, so it looks a bit interesting:
image

Also, I moved back rowData to slot binds and added it to expandableRow.

packages/ui/src/components/va-data-table/VaDataTable.vue
```diff
@@ -41,104 +39,118 @@
...
+
+                    <slot v-else name="cell" v-bind="{ cell, row }">
+                      <span v-if="$props.grid" class="va-data-table__grid-column-header">{{ columnsComputed[cellIndex].label }}</span>
+                      {{ cell.value }}
+                    </slot>
+                  </td>
+                </tr>
+                <tr
+                  v-show="row.isExpandableRowVisible"
+                  class="va-data-table__table-tr--expanded va-data-table__table-expanded-content"
```
m0ksem, line 141: Also, classes here are important to have, because likely user want table to be customizable.

packages/ui/src/components/va-data-table/hooks/useRows.ts
```diff
@@ -54,15 +54,28 @@ const buildTableRow = (
     itemKey,
     source,
     cells: columns.map(column => buildTableCell(initialIndex, itemKey, source, column)),
+    rowData: source,
```
m0ksem, line 57: Alias to rowData, so it has the same API as DataTableCell

aluarius (approved):

PR #3055, opened 2023-02-23
Fix/ssr css variables

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3046

Added a build plugin which converts `v-bind(variable)` in <style> to `:style=""`.

Basically, creates a css variable like `--va-[index]-[kebab-name]` and pass it to style using v-bind and replaces vue v-bind in css with generated css variable. This way we ignore vue `useCSSVars` which is not available in cjs build. We also ensure that passed value to css variable is going to be String, which makes it work in Nuxt.

### DIFF
--- a/packages/ui/build/plugins/component-v-bind-fix.ts
+++ b/packages/ui/build/plugins/component-v-bind-fix.ts
@@ -24,7 +24,14 @@ const renderObjectGuard = (styleContent: string, binds: string[]) => {
   const cssVariablesObject = binds.map((vBind, index) => {
     return `'--va-${index}-${kebabCase(vBind)}': String(${vBind})`
   }, '').join(',')
-  return `typeof ${styleContent} === 'object' ? { ...${styleContent}, ${cssVariablesObject} } : ${styleContent} + \`;${cssVariablesString}\``
+
+  // Merge existing style with css variables
+  const arrayStyle = `[...${styleContent}, \`${cssVariablesString}\`]`
+  const objectStyle = `{ ...${styleContent}, ${cssVariablesObject} }`
+  const stringStyle = `${styleContent} + \`;${cssVariablesString}\``
+
+  // Handle if style is an object, array or string
+  return `typeof ${styleContent} === 'object' ? (Array.isArray(${styleContent}) ? ${arrayStyle} : ${objectStyle}) : ${stringStyle}`
 }
 
 export const transformVueComponent = (code: string) => {
--- a/packages/ui/build/plugins/component-v-bind-fix.ts
+++ b/packages/ui/build/plugins/component-v-bind-fix.ts
@@ -1,9 +1,10 @@
 import { Plugin } from 'vite'
+import kebabCase from 'lodash/kebabCase'
 
 const parseVBinds = (style: string) => {
   return style
     .match(/v-bind\((.*)\)/g)
-    ?.map((line) => line.match(/v-bind\((.*)\)/)![1]) ?? []
+    ?.map((line) => line.match(/v-bind\(([^)]*)\)/)![1]) ?? []
 }
 
 const getRootNodeCode = (code: string) => {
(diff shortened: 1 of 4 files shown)


packages/ui/build/plugins/component-v-bind-fix.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    return code.replace(/style="([^"]*)"/, `:style="\`$1;${cssVariablesString}\`"`)
+  }
+
+  if (/\/>$/.test(rootNode)) {
+    const newRootNode = rootNode.replace('/>', ` :style="\`${cssVariablesString}\`"/>`)
+    return code.replace(rootNode, newRootNode)
+  }
+
+  const newRootNode = rootNode.replace('>', ` :style="\`${cssVariablesString}\`">`)
```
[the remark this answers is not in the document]
m0ksem: That's the idea. To replace first occurred `>`. Am I doing something wrong here?

packages/ui/src/composables/useCSSVariables.ts
```diff
@@ -2,5 +2,8 @@ import { computed } from 'vue'
 import kebab from 'lodash/kebabCase.js'
 
 export const useCSSVariables = (prefix: string, cb: () => Record<string, string>) => {
-  return computed(() => Object.entries(cb()).map(([key, value]) => ({ [`--${prefix}-${kebab(key)}`]: value })))
+  return computed(() => Object.entries(cb()).reduce((acc, [key, value]) => {
+    acc[`--${prefix}-${kebab(key)}`] = value
```
aluarius, line 6: Would be nice to have here return value typing as in `useBem`. But up to you.

packages/ui/build/plugins/component-v-bind-fix.ts
```diff
@@ -0,0 +1,92 @@
+import { Plugin } from 'vite'
+import kebabCase from 'lodash/kebabCase'
+
+const parseVBinds = (style: string) => {
+  return style
+    .match(/v-bind\((.*)\)/g)
```
aluarius, line 6: Comment these two lines, unreadable :c

packages/ui/build/plugins/component-v-bind-fix.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const renderCssVariablesString = (vBinds: string[]) => {
+  return vBinds.map((vBind, index) => {
+    return `--va-${index}-${kebabCase(vBind)}: \${String(${vBind})}`
+  }, '').join(';')
+}
+
+const renderObjectGuard = (styleContent: string, binds: string[]) => {
+  const cssVariablesString = renderCssVariablesString(binds)
+  const cssVariablesObject = binds.map((vBind, index) => {
```
aluarius, line 24: Result will be string, no?

packages/ui/build/plugins/component-v-bind-fix.ts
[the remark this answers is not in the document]
aluarius: Mb you should test it here as at 71 line?

aluarius (approved):
Logic is pretty hard for understanding, please comment it as much as possible.

packages/ui/build/plugins/component-v-bind-fix.ts
m0ksem: It is as string that contains the rendered Object. Renamed.

packages/ui/src/composables/useCSSVariables.ts
m0ksem: To much effort for it...

packages/ui/build/plugins/component-v-bind-fix.ts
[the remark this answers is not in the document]
m0ksem: I think it is false positive here, because rootNode is `<tagName attrs>`.

aluarius (approved):
Not sure if it related to this pr, but if you're using light theme of the nuxt-docs and then go to http://localhost:3000/en/styles/colors, it automatically switches to the dark one.

That pretty weird.

PR #3066, opened 2023-02-24
[#2980] VaTreeView expandable node cursor style

### LINKED ISSUE #2980: [VaTreeView] pointer or cursor option when hovering over Expandable Node

I think it's not clear enough that you can click on the nodes themselves to expand them (instead of clicking on the arrow).

I think it would be more clear if the cursor style would be `pointer` when hovering over the node (except for the text).

**Describe alternatives you've considered**

--

**Additional context**

Clicking on the nodes also does not work in the docs.

thunfisch987:
## Description

(closes #2980 )
When expandNodeBy is "node" you can click on the node itself to expand it.
This is not visible for the end user because the mouse is still the same.
This PR changes the cursor style to `pointer` when expandNodeBy==="node" and it has children (=when it's expandable).
Used a scss variable called --va-tree-node-cursor-type (so the developer could change it)

## Markup:

<details>

`components/va-tree-view/components/VaTreeNode/VaTreeNode.vue`  
```vue
<div class="va-tree-node-content__body" :class="cursorClassComputed">
  <slot name="content" v-bind="$props.node">{{ labelComputed }}</slot>
</div>
```
```ts
const cursorClassComputed = useBem('va-tree-node-content', () => ({
  cursor: (props.node.hasChildren === true && expandNodeBy === 'node'),
}))
```
`.../VaTreeNode/_variables.scss`  
```scss
.va-tree-node {
  ...
  &--cursor {
    cursor: var(--va-tree-node-cursor-type);
  }
  ...
}
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
+++ b/packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
@@ -127,7 +127,7 @@ export default defineComponent({
     }))
 
     const cursorClassComputed = useBem('va-tree-node-content', () => ({
-      expandable: (props.node.hasChildren === true && expandNodeBy === 'node'),
+      clickable: (props.node.hasChildren === true && expandNodeBy === 'node'),
     }))
 
     const tabIndexComputed = computed(() => props.node.disabled ? -1 : 0)
@@ -220,6 +220,10 @@ export default defineComponent({
     &--indent {
       margin-left: var(--va-tree-node-indent);
     }
+
+    &--clickable {
+      cursor: pointer;
+    }
   }
 
   &-children {
@@ -244,10 +248,6 @@ export default defineComponent({
       cursor: pointer;
       pointer-events: all;
     }
-  }
-
-  &--expandable {
-    cursor: pointer;
   }
 
   &:focus-visible > .va-tree-node-root {
--- a/packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
+++ b/packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
@@ -127,7 +127,7 @@ export default defineComponent({
     }))
 
     const cursorClassComputed = useBem('va-tree-node-content', () => ({
-      cursor: (props.node.hasChildren === true && expandNodeBy === 'node'),
+      expandable: (props.node.hasChildren === true && expandNodeBy === 'node'),
     }))
 
     const tabIndexComputed = computed(() => props.node.disabled ? -1 : 0)
@@ -246,8 +246,8 @@ export default defineComponent({
     }
   }
 
-  &--cursor {
-    cursor: var(--va-tree-node-cursor-type);
+  &--expandable {
+    cursor: pointer;
   }
 
   &:focus-visible > .va-tree-node-root {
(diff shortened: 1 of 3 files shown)


thunfisch987:
I hope i formatted everything right

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
```diff
@@ -126,6 +126,10 @@ export default defineComponent({
       indent: props.node.hasChildren === false,
     }))
 
+    const cursorClassComputed = useBem('va-tree-node-content', () => ({
+      cursor: (props.node.hasChildren === true && expandNodeBy === 'node'),
```
m0ksem, line 130: Basically, following BEM convention, you need to specify modifier. Let's have a specific cursor for a specific (expandable) content node.

```suggestion
      expandable: (props.node.hasChildren === true && expandNodeBy === 'node'),
```

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
```diff
@@ -241,6 +246,10 @@ export default defineComponent({
     }
   }
 
+  &--cursor {
+    cursor: var(--va-tree-node-cursor-type);
```
m0ksem, line 250:
```suggestion
    cursor: pointer;
```

I find strange user for typing code like:
```css
:root {
  --va-tree-node-cursor-type: loading;
}
```
99% user want to have cursor pointer over interactive (expandable) element. It is also unclear from CSS variable that this variable is used for only nodes which can be expanded.

m0ksem (changes requested):
Thanks! Appreciate effort dealing with useBem and css variables. Usually do not expect new contributors for digging deep in project. But I think for this particular issue there is no need in CSS variables. We use them only for properties that we expect user might want to change, like backgrounds, colors, sizes or alignment, but not cursor over interactive elements. Also, `cursor` class name doesn't represent muddier of va-tree-node.

thunfisch987:
> ... represent _muddier_ of ...

What does muddier mean in that context?

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
thunfisch987: I changed it in the code, but i personally think that just `expandable` is a bit misleading because it only gets applied when `expandNodeBy` is also `"node"`, so maybe `expandablebynode` / `expandableByNode` (however this seems kind of repeating)

m0ksem:
> 

I meant “modifier”, “cursor class name doesn't represent modifier of va-tree-node."

packages/ui/src/components/va-tree-view/components/VaTreeNode/VaTreeNode.vue
thunfisch987: Any Ideas?

rustem-nasyrov: `on-expandable`.

m0ksem: Make sense. Then it can be renamed to `clickable` or `interactive`.

m0ksem (approved):

m0ksem:
I updated PR, there was a small issue with expandable selector. Overall looks good. Lets simple call it clickable :)

Issue #2970, opened 2023-02-14
Split translation across pages


asvae:
Idea is that 
* vue-18n-unused would just compare other translations to en, then developer would compare to examples etc.
* tree shaking would work properly (right now we're using full translation file everywhere)
* more convenient to work with (all files for page are in one folder)
* do not forget to add "all" translations

m0ksem:
Might worth it to add TS support.

export default definePageTranslation({
  title: string,
  description?: string
  examples?: Record<string, {
     title: string,
     description?: string
  }>
})

Issue #2254, opened 2022-08-17
Add "how to test" page


m0ksem:
Let's make a small guide for user how to test vuestic app with jest, vitest.

asvae:
Also added to this issue - https://github.com/epicmaxco/vuestic-ui/issues/1991. I see it as a part of bigger problem.

m0ksem:
Code snippet for vitest.

```
export default {
  plugins: [vue()],
  test: {
    deps: {
      inline: [
        "vuestic-ui"
      ]
    },
    globals: true,
    environment: 'jsdom',
  },
  resolve: {
    alias: {
      "@": fileURLToPath(new URL("./src", import.meta.url)),
      'vuestic-ui': 'vuestic-ui/dist/esm/main.js'
    },
  },
};
```

gabriellatavares:
How can I test with jest?

PR #3073, opened 2023-02-25
Docs/i18n split

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2254
closes https://github.com/epicmaxco/vuestic-ui/issues/2970

Split translations for docs.
image

And testing docs in one PR:
image

### DIFF
--- a/packages/docs/locales/en.json
+++ b/packages/docs/locales/en.json
@@ -1667,7 +1667,8 @@
     "i18n": "I18n",
     "webComponents": "Web Components",
     "colorsClasses": "Colors Classes",
-    "skeleton": "Skeleton"
+    "skeleton": "Skeleton",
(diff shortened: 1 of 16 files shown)


m0ksem:
Maybe gonna add a TS support here and store translations in one file after discussion.

m0ksem:
Decided not to make a TS support and prefer separated files instead.

packages/docs/locales/en.json
```diff
@@ -1667,7 +1667,8 @@
     "i18n": "I18n",
```
asvae, line 1667: Some translations look to be broken:

image

asvae (changes requested):
Local version is somewhat broken.

PR #3085, opened 2023-02-27
feat: remove extra padding

rustem-nasyrov:
## Description
Removed redundant padding.

image

## Types of changes


aluarius (approved):
Seems that there're no more need in this hack for Safari. But better double check from real device after deploy to prod.

rustem-nasyrov:
> Seems that there're no more need in this hack for Safari. But better double check from real device after deploy to prod.

I've checked it in the Browserstack. At first sight it was looking good.

Issue #3080, opened 2023-02-27
Something wrong with images


kushich:
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
go to image or carousel page

**What is the expected behavior?**
Correct image sizes

**What is the current behavior?**
<img width="1072" alt="Screenshot 2023-02-27 at 09 35 02" src="https://user-images.githubusercontent.com/45711229/221491936-88f579b7-3aed-492e-b97f-7f2914706d1d.png">
<img width="1099" alt="Screenshot 2023-02-27 at 09 35 12" src="https://user-images.githubusercontent.com/45711229/221491940-136d533a-fb30-48af-b4f1-4a9bfa95d845.png">
![Uploading Screenshot 2023-02-27 at 09.37.34.png…]
<img width="1025" alt="Screenshot 2023-02-27 at 09 37 46" src="https://user-images.githubusercontent.com/45711229/221492121-e89e5f1c-daf5-4369-be7d-03d84c0aa34e.png">
<img width="1096" alt="Screenshot 2023-02-27 at 09 37 54" src="https://user-images.githubusercontent.com/45711229/221492128-b973f8e2-e1b4-439a-86d8-2ef14dc83686.png">

aluarius:
Seems to be just a problems with docs examples, but i'll check.

PR #3090, opened 2023-02-27
fix: va-image docs examples fixes

aluarius:
Closes: #3080 

## Description
- [x] fixed examples in docs;
- [x] carousel image width fix. 

## Types of changes

### DIFF
--- a/packages/docs/page-config/ui-elements/image/examples/Slots.vue
+++ b/packages/docs/page-config/ui-elements/image/examples/Slots.vue
@@ -1,7 +1,6 @@
 <template>
-  <div class="mb-4 w-1/3 flex">
+  <div class="mb-4 flex gap-2">
     <va-button
-      class="mr-2"
       color="success"
       @click="newImage"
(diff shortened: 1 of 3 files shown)


rustem-nasyrov (approved):

m0ksem (changes requested):
image

Ration and object fit demo are incorrect. I think we need to remove flex from here.

Also, why we do not use css `aspect-ratio` here? It will work with flex stretch. Maybe in addition to padding hack.

aluarius:
Why `fit` example is incorrect? Contain-cover-fill seems to be visually correct.

Ratio didn't notice, will investigate and fix.

aluarius:
There were some reasons not to use `aspect-ratio` but can't remember them. Will create separate research issue.

aluarius:
> Ratio didn't notice, will investigate and fix.

Fixed here https://github.com/epicmaxco/vuestic-ui/pull/3052

Issue #2446, opened 2022-09-27
`useSelectableList` composable refactoring


aluarius:
Current implementation isn't obvious and pretty complicated. Will be nice to refactor it, simplify and leave comments about current logic, because if we will need to rely on it in new components or will be forced to add new features for it, that will cause lot's of bugs and persistent headache.

aluarius:
Will deal with it during https://github.com/epicmaxco/vuestic-ui/issues/2093

Issue #2093, opened 2022-07-14
Track-by, value-by, text-by for VaButtonToggle


m0ksem:
useSelectable

Issue #1074, opened 2021-09-06
Investigate textBy, valueBy, trackBy usage


Derranion:
Investigate **textBy**, **valueBy**, **trackBy** props inside " **useSelectableList** " to make sure that they are used correctly  - there could be more bugs like this one:

-  [if text-by points to number - searchable select breaks](https://github.com/epicmaxco/vuestic-ui/issues/1057)

Some dependencies and names are pretty questionable as well.

Update documentation if something will be changed.

asvae:
Would add that we also want to restore tests, as these are fully covered, if I remember correctly.

PR #3092, opened 2023-02-27
Refactor: `useSelectable` refactor and integration to `VaButtonToggle`, `VaSelect` bugfix

aluarius:
Closes: #1074 
Closes: #2446
Closes: #2093
Closes: #3107 
Closes: #3138

## Description
- [x] `useSelectableList` refactoring;
- [x] `useSelectableList` integrated to `VaButtonToggle`;
- [x] zero values proceeding fix;
- [x] dev helper fix;
- [x] zero value deselection fix; 
- [x] related refactoring and bugfixing. 

## Types of changes


packages/ui/src/utils/env.ts
```diff
@@ -1 +1 @@
-export const __DEV__ = typeof process !== 'undefined' && process.env.NODE_ENV !== 'production'
+export const __DEV__ = typeof process === 'undefined' ?? process.env.NODE_ENV !== 'production'
```
m0ksem, line 1: Check this: https://github.com/unjs/std-env, maybe we could use it instead.

m0ksem: `typeof process === 'undefined' ` will always return Boolean which is not undefined or null, so `??` will return left operand always...

packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -177,7 +176,7 @@ export default defineComponent({
       }, { _noGroup: [] }))
     const optionGroupsThrottled = useThrottleValue(optionGroups, props)
 
-    const isValueExists = (value: SelectOption | null | undefined): value is SelectOption => !!value || value === 0
+    const isValueExists = (value: SelectOption | null | undefined): value is SelectOption => !!value || value === 0 || value === false
```
m0ksem, line 179:
```suggestion
    const isValueExists = (value: SelectOption | null | undefined): value is SelectOption => value !== undefined && value !== null
```

packages/ui/src/components/va-select/components/VaSelectOption/VaSelectOption.vue
```diff
@@ -85,10 +85,9 @@ export default defineComponent({
 
     const isSelected = computed(() => props.getSelectedState(props.option))
     const isFocused = computed(() => {
-      if (!props.currentOption && props.currentOption !== 0) { return false }
+      if (!props.currentOption && props.currentOption !== 0 && props.currentOption !== false) { return false }
```
m0ksem, line 88:
```suggestion
      if (props.currentOption === undefined || props.currentOption === null) { return false }
```

packages/ui/src/components/va-select/VaSelect.vue
```diff
@@ -413,7 +413,7 @@ export default defineComponent({
     const hoveredOption = ref<SelectOption | null>(null)
 
     const selectHoveredOption = () => {
-      if (!hoveredOption.value && hoveredOption.value !== 0) { return }
+      if (!hoveredOption.value && hoveredOption.value !== 0 && hoveredOption.value !== false) { return }
```
m0ksem, line 416: lodash.isNil....

packages/ui/src/composables/useSelectableList.ts
```diff
@@ -16,34 +17,56 @@ export const useSelectableListProps = {
...
+
+      return value
+    }
+
+    return optionByValue
   }
 
-  const getValue = (option: SelectableOption) => getOptionProperty(option, props.valueBy)
+  const getTrackBy = (option: SelectableOption): string | number => {
+    console.log()
```
m0ksem, line 43:
```suggestion
```

packages/ui/src/composables/useSelectableList.ts
```diff
@@ -16,34 +17,56 @@ export const useSelectableListProps = {
 }
 
 export function useSelectableList (props: ExtractPropTypes<typeof useSelectableListProps>) {
-  const isStringOrNumber = (option: SelectableOption): option is (string | number) => {
-    const typeOfOption = typeof option
-    return typeOfOption === 'string' || typeOfOption === 'number'
+  const getOptionProperty = (option: SelectableOption, prop: StringOrFunction) => {
```
m0ksem, line 20: This method can be outside of composable and this way we potentially can save a bit of memory.

packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
m0ksem:
```suggestion
    const isValueExists = (value: SelectOption | null | undefined): value is SelectOption => value !== undefined && value !== null
```

m0ksem (changes requested):
Please, consider lodash.isNil.

Also, I'm not gonna allow you to merge it without tests...

Please, cover with tests:
- [ ] value-by-key
- [ ] useSelectableList
- [ ] VaSelect (yep, just with tests related to trackBy, valueBy, etc.)
- [ ] VaButtonToggle (just  trackBy, valueBy, etc.)
- [ ] VaOptionList (trackBy, valueBy, etc.)

It is much easier to write tests here then testing it is manually for me... Also, use whole power of Copilot here :)

rustem-nasyrov (approved):

packages/ui/src/components/va-select/tests/VaSelect.spec.ts
```diff
@@ -1,14 +1,65 @@
...
+        attrs: {
+          options,
+          textBy: 'label',
+          valueBy: 'value',
+          trackBy: 'id',
+          groupBy: 'group',
+          modelValue: value,
+        },
+      })
+
```
m0ksem, line 56: wrapper.find('input').text().toBe('Three')

packages/ui/src/utils/env.ts
```diff
@@ -1 +1,5 @@
-export const __DEV__ = typeof process !== 'undefined' && process.env.NODE_ENV !== 'production'
+const processShim: typeof process = typeof process !== 'undefined' ? process : ({} as typeof process)
+const envShim = processShim.env || ({} as typeof process.env)
+const nodeEnv = envShim.NODE_ENV || ''
```
m0ksem, line 3:
```suggestion
const nodeEnv = envShim.NODE_ENV || 'prod'
```

Let's make it prod by default if there are not process

m0ksem (approved):

packages/ui/src/utils/env.ts
aluarius: This will break warnings in demos, cause there're no process. Have made a separate issue for this: https://github.com/epicmaxco/vuestic-ui/issues/3141

packages/ui/src/utils/tests/value-by-key.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      props: {
+        option: { prop: 'string' },
+        prop: undefined,
+      },
+      expected: { prop: 'string' },
+    },
+    {
+      props: {
+        option: { prop: 'string' },
+        prop: 1,
```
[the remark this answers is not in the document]
aluarius: It's initial behaviour:
- if something wrong with the type of the key (number, undef etc.) and helper isn't able to proceed it, option is returned itself;
- If the type of the key is correct, but target prop wasn't found in the provided object, it obviously returns undefined.

I was thinking about making both cases to return `undefined`, but it may create mega breaking change for nothing.

PR #3094, opened 2023-02-27
feat: add gravatar usage example

### LINKED ISSUE #2663: VaAvatar improvements

rustem-nasyrov:
Close #2663 

image

## Description
- Gravatar usage examples.
- Default fallback with global config.

## Types of changes

### DIFF
--- a/packages/docs/locales/en.json
+++ b/packages/docs/locales/en.json
@@ -1858,6 +1858,10 @@
       "fallback": {
         "title": "Fallback",
         "text": "You can set fallback image."
+      },
+      "gravatarIntegration": {
+        "title": "Integrate with Gravatar",
+        "text": "You can easily use Gravatar with `VaAvatar` component."
       }
     },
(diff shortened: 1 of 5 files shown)


aluarius (approved):

asvae:
Maybe we can use some email (maybe of evan or posva?).
As that's what example is trying to show, not the fact that VaImage accepts image.

https://en.gravatar.com/site/implement/hash/
If it's too hard - let's drop the idea and use your current implementation.

asvae (changes requested):
commented

asvae (approved):
Cool!

Issue #3062, opened 2023-02-24
Add config into release script if we published all sub-packages


m0ksem:
Need a small reminder to publish @vuestic/nuxt, @vuestic/tailwind, @vuestic/ag-grid etc.

PR #3096, opened 2023-02-28
chore: added todo list after release

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3062


asvae:
It would help to have these commands executed only as part of `minor` and `major` releases (`experimental` and `next` could be spared).

I'm thinking:

```ts
const releaseTodo = {
  minor: [
    '- Update and release other packages (like nuxt, create-vuestic, etc)',
    // ...
  ],
  major: [
    
  ]
}

export type ReleaseConfig = {
  // ...
  releaseTodo?: keyof typeof releaseTodo,
}
```

asvae (changes requested):
commented

asvae (approved):
Perfect!

PR #3101, opened 2023-03-01
fix: remove grid and reset styles from config defaults

### LINKED ISSUE #2762: Using vuestic-ui in new nuxt project breaks layout
**Vuestic-ui version:** 1.5.2

imageimage

rustem-nasyrov:
Close #2762

## Description
- Removed grid and reset stylesheets from pre-defined configuration.

## Types of changes


asvae:
Please also be careful and be sure this does not introduce breaking change.

packages/nuxt/src/module.ts
```diff
@@ -20,7 +20,7 @@ export default defineNuxtModule<VuesticOptions>({
 
   defaults: {
     config: {},
-    css: true,
+    css: ['smart-helpers', 'typography'],
```
m0ksem, line 23: Also, update `@default true` to `@default ['smart-helpers', 'typography']` in types.

m0ksem (approved):
I fixed nuxt build, so you can test it.

packages/docs/modules/vuestic.ts
```diff
@@ -25,8 +25,10 @@ export default defineNuxtModule<VuesticOptions>({
     nuxt.options.alias['@vuestic/ag-grid-theme'] = resolve(__dirname, '../../ag-grid-theme/src/styles/index.scss');
     nuxt.options.alias['vuestic-ui/styles/typography.css'] = resolve(__dirname, '../../ui/src/styles/typography/typography.scss');
     nuxt.options.alias['vuestic-ui/styles/grid'] = resolve(__dirname, '../../ui/src/styles/grid/grid.scss');
+    nuxt.options.alias['vuestic-ui/styles/grid.css'] = resolve(__dirname, '../../ui/src/styles/grid/grid.scss');
```
aluarius, line 28: Can't understand, why we need these two aliases.

aluarius (approved):

packages/docs/modules/vuestic.ts
m0ksem: Same. But it doesn't look harmful.

rustem-nasyrov: Otherwise docs build fails.

m0ksem:
I'm gonna release 1.1.0 after this PR.

PR #3113, opened 2023-03-02
Remove nuxt playground

m0ksem:
Removed playground. Use sandbox:nuxt instead.


rustem-nasyrov:
@m0ksem, isn't playground required for the Nuxt modules? I checked that a lot of modules have a playground.

m0ksem:
> @m0ksem, isn't playground required for the Nuxt modules? I checked that a lot of modules have a playground.

I don't think it is required. It is a good practice to have a playground to check how module works. We have this playground in sandbox, so we don't need to have two playgrounds IMO. It doesn't impact the module in terms of types, build or whatever.

packages/nuxt/package.json
```diff
@@ -24,11 +24,7 @@
   ],
   "scripts": {
     "build": "nuxt-module-build",
-    "prepack": "nuxt-module-build",
-    "dev": "nuxi dev playground",
-    "dev:build": "nuxi build playground",
-    "dev:prepare": "nuxt-module-build --stub && nuxi prepare playground",
-    "dependenciesInstalled": "node ./scripts/stud-nuxt.mjs"
+    "prepack": "nuxt-module-build"
```
rustem-nasyrov, line 27: Aren't they equal?

rustem-nasyrov: I mean `build` and `prepack` are running the same command.

m0ksem: prepack is used by npm and runs before npm publish. Build is used by us to get thing builded.

rustem-nasyrov (approved):

PR #3115, opened 2023-03-02
Feat/colors extensions

### LINKED ISSUE #1425: Color helpers
Should be added as Optional Service
Depends on Colors Service

Usage example:
```html
<span class="text--primary bg--secondary">primary secondary</span>
```

We had this functionality before, but it disappeared mysteriously 🤪.
- [ ] Service should generate CSS dynamically based on colors config. I suggest you look into prior implementation (couple years back should be enough). The idea is that we attach `styles` to `head`, then replace that block on colors change. Old implementation was well implemented.
- [ ] should support only text and background: `text--primary`, `bg--secondary`
- [ ] Should have no configuration
- [ ] CSS helpers should rely on CSS variables (that are part of Colors Service).
- [ ] Document Color helpers as a separate page in Services category. Add installation instructions.
- [ ] There are still some places with `text--primary`, these do nothing and we need to fix these (remove or make do something).

rustem-nasyrov:
Close #1425

image

## Description
- Added UnoCSS integration guide page.
- Translations.
- Setup example.

## Types of changes


packages/docs/page-config/extensions/unocss/code/unocss.config.js
```diff
@@ -0,0 +1,30 @@
...
+  presets: [
+    presetUno({
+      dark: 'media',
+    }),
+    presetIcons(),
+    presetTheme({
+      theme: {
+        dark: {
+          colors: colorsPreset.dark,
+        },
```
m0ksem, line 26:
```suggestion
        dark: {
          colors: colorsPreset.dark,
        },
        light: {
           colors: colorsPreset.light
        }
```

Seems quite logical to me. Not sure if it is correct.

packages/docs/page-config/extensions/unocss/code/unocss.config.js
```diff
@@ -0,0 +1,30 @@
...
+export default defineConfig({
+  theme: {
+    colors: colorsPreset.light,
+    breakpoints: convertNumbersToCssUnits(thresholdsPreset),
+  },
+  presets: [
+    presetUno({
+      dark: 'media',
+    }),
+    presetIcons(),
```
m0ksem, line 21: Is this required?

m0ksem (approved):

packages/docs/page-config/extensions/unocss/code/unocss.config.js
rustem-nasyrov: As I understood, preset was designed like that.

aluarius (approved):

m0ksem:
Can you resolve this conflict? Try to use split translations here.

Issue #3049, opened 2023-02-22
Button dropdown is somewhat broken in docs


asvae:
**Vuestic-ui version:** 1.6.1

> Button dropdown component isn’t very mobile friendly in the docs, all of the buttons to the right shift when you open one of the menus.

Issue #3047, opened 2023-02-22
Some select is probably broken on landing page


asvae:
**Vuestic-ui version:** 1.6.1

> Why does the select width expand to 100% when you open it and go smaller again after selecting an option?

> iOS chrome and safari

https://vuestic.dev

m0ksem:
I think it is related to VaDropdown. I'm not able to reproduce it in Safari 16.3.

asvae:
See https://github.com/epicmaxco/vuestic-ui/issues/3047

aluarius:
As for Maxim, wasn't able to reproduce for Chrome, Safari and Firefox.

Suggest to close it for now and reopen if someone will complain.

m0ksem:
@aluarius, check mobile safari

rustem-nasyrov:
@aluarius, here is reproduction:

<img width="407" alt="image" src="https://user-images.githubusercontent.com/29167241/221797218-a1bed7ec-43cc-4bd3-9646-96008e5abdac.png">
<img width="420" alt="image" src="https://user-images.githubusercontent.com/29167241/221797290-d1d3dc57-fe58-45f8-ad06-7e467662f37c.png">

aluarius:
Yea, thnx, already found it.

aluarius:
All in all, this issue is related to https://github.com/epicmaxco/vuestic-ui/issues/3049 

To solve it in optimal way, me and Maxim decided to rewrite VaDropdown via render function.

aluarius:
Closed via https://github.com/epicmaxco/vuestic-ui/pull/3119

PR #3119, opened 2023-03-03
[#3049] [#3047] va-dropdown rewritten to render reimplementation

aluarius:
Closes: #3049 #3047 

## Description
[x] `va-dropdown` reimplementation to render function.

## Types of changes

### DIFF
--- a/packages/docs/page-config/ui-elements/dropdown/examples/PlacementAndOffset.vue
+++ b/packages/docs/page-config/ui-elements/dropdown/examples/PlacementAndOffset.vue
@@ -22,7 +22,7 @@
             :offset="offset"
           >
             <template #anchor>
-              <Coordinates :placement="placement" />
+              <div><Coordinates :placement="placement" /></div>
             </template>
 
             <va-dropdown-content> Dropdown </va-dropdown-content>
--- a/packages/ui/src/components/va-dropdown/VaDropdown.vue
+++ b/packages/ui/src/components/va-dropdown/VaDropdown.vue
@@ -1,39 +1,21 @@
-<template>
-  <div
-    class="va-dropdown"
-    :class="computedClass"
-    ref="anchorRef"
-    role="button"
-    :aria-label="t('toggleDropdown')"
-    :aria-disabled="$props.disabled"
-    :aria-expanded="!!valueComputed"
-    :aria-controls="idComputed"
-    style="position: relative;"
-  >
-    <slot name="anchor" v-bind="{ value: valueComputed, hide, show }" />
-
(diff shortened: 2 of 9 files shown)


m0ksem:
@aluarius, I fixed VaDropdown render. There are still weird issue with Nuxt. There are hydration mismatch SOMETIMES (for example, in our landing, but not in sandbox). It renders COMMENT node with `v-if` content on server, but on client. I have not clue how to debug it correctly and how we can fix it.

Looks like it is because VaIcon has slot in dynamic tag and there is v-if inside. The last commit here is me fixing hydration missmatch. 

😭

BTW, overall looks good.

packages/docs/page-config/ui-elements/dropdown/examples/PlacementAndOffset.vue
```diff
@@ -22,7 +22,7 @@
             :offset="offset"
           >
             <template #anchor>
-              <Coordinates :placement="placement" />
+              <div><Coordinates :placement="placement" /></div>
```
m0ksem, line 25: Wrappers can be removed now

packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -136,7 +119,11 @@ export default defineComponent({
       if (close && props.trigger !== 'none') { valueComputed.value = false }
     }
 
-    const elRef = useHTMLElement('anchorRef')
+    // const anchorRef = shallowRef<HTMLElement>()
+    const elRef = useHTMLElement('computedAnchorRef')
+    // watch(elRef, (v) => {
```
m0ksem, line 124: This can be removed

m0ksem (approved):

m0ksem:
The main problem here is to find out how to reproduce this bug in vue sfc to make a bug report.

asvae:
> The main problem here is to find out how to reproduce this bug in vue sfc to make a bug report.

@m0ksem it only reproduces for compiled component?

m0ksem:
> > The main problem here is to find out how to reproduce this bug in vue sfc to make a bug report.
> 
> @m0ksem it only reproduces for compiled component?

Complete opposite. I'm not able to reproduce it in sandbox, where we use compiled vuestic. It seems to be similar to Vue hydration mismatch in transition component. For some reason on client comment vnode with "v-if" doesn't render, so it is not critical.

packages/docs/page-config/ui-elements/dropdown/examples/PlacementAndOffset.vue
aluarius: Wrapper is still needed. Without it error is thrown.

PR #3127, opened 2023-03-06
TreeView add children-by prop

### LINKED ISSUE #3077: [VaTreeView] TreeNode type interface defines children property as required
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
Create a VaTreeView and assign an Array of `TreeNode`s with different property to the `:nodes` prop.

Omitting the `children` property should not throw a type error (at least on the first / highest level.
It's very tedious to add `children: []` to ever TreeNode to the "Sub-TreeNodes"

If you don't have a children property on one of the TreeNodes Typescript throws a tantrum.

**Other information**
Is this behaviour wanted?
I personally think, that the `children` property should be required on the top-level, but not on every Sub-TreeNode

rustem-nasyrov:
Close #3077

## Description
- Added `children-by` property support.

## Example
```html
<!-- template -->
<va-tree-view
  :nodes="nodes"
  children-by="items"
/>
```
```js
// script
const nodes = [
  {
    id: 1,
    label: 'Category',
    items: [
      { id: 2, label: 'Item' },
      {
        id: 3,
        label: 'Subcategory',
        items: [
          { id: 4, label: 'Item' },
          { id: 5, label: 'Item' },
        ],
      },
      { id: 6, label: 'Item' },
    ],
  },
  { id: 7, label: 'Item' },
]
```

## Types of changes


packages/ui/src/components/va-tree-view/types.ts
```diff
@@ -3,7 +3,7 @@ export type TreeViewEmitsFunc = (event: string, newValues: unknown) => void
 
 export interface TreeNode {
   id: number | string
-  children: TreeNode[]
+  children?: TreeNode[]
```
m0ksem, line 6: Why didn't you remove children property here? Seems to be useless.

m0ksem (approved):

packages/ui/src/components/va-tree-view/hooks/useTreeHelpers.ts
```diff
@@ -105,6 +109,7 @@ export const useTreeHelpers = (props: ExtractPropTypes<typeof useTreeViewProps>)
   const getDisabled = (node: TreeNode) => getNodeProperty(node, props.disabledBy)
   const getExpanded = (node: TreeNode) => getNodeProperty(node, props.expandedBy)
   const getTrackBy = (node: TreeNode) => getNodeProperty(node, props.trackBy)
+  const getChildren = (node: TreeNode) => getNodeProperty(node, props.childrenBy) || []
```
aluarius, line 112: Minor. Imo better use here ?? to catch unexpected values.

packages/ui/src/components/va-tree-view/hooks/useTreeView.ts
```diff
@@ -87,13 +88,15 @@ const useTreeView: UseTreeViewFunc = (props, emit) => {
         nodes.forEach((node: TreeNode) => {
           if (node.disabled) { return }
 
-          if (node.children?.length) { toggleChildren(node.children) }
+          const children = getChildren(node)
```
aluarius, line 91: Can it be computed not to calc this value every time?

aluarius (approved):

packages/ui/src/components/va-tree-view/hooks/useTreeView.ts
rustem-nasyrov: Nope, I'll think about it later. 😅

Issue #2964, opened 2023-02-14
[VaSelect]: selectedTopShown mode should keep hovered option under cursor after selection


aluarius:
<img width="465" alt="image" src="https://user-images.githubusercontent.com/5394573/218587075-8efb643d-b497-456b-8f54-d42cc3af0644.png">

I this case I would keep cursor at the same position it was and not jump with selected item. But please investigate, I might be mistaken. :)

_Originally posted by @asvae in https://github.com/epicmaxco/vuestic-ui/pull/2909#discussion_r1105075039_

Issue #2864, opened 2023-01-12
Switch height bug


aluarius:
In flex container va-switch with, for example, error label will be rendered in this weird way:

<img width="729" alt="image" src="https://user-images.githubusercontent.com/64714442/212016798-b5c97064-d8cd-4fb0-a9df-76275dcde6fc.png">

No problem for block:

<img width="192" alt="image" src="https://user-images.githubusercontent.com/64714442/212016868-e7f08208-f0af-4887-a44a-5004cb77e24a.png">

User can solve this with align-items: flex-start, but it's not good :)

This behaviour is caused by height: 100% of the va-switch container and @m0ksem says that we need it.

So, we should solve this situation.

m0ksem:
Might be related to this issue https://github.com/epicmaxco/vuestic-ui/issues/1323

PR #3129, opened 2023-03-07
[#2864] `va-select` hovered position fix

aluarius:
Closes: #2964

## Description
- [x] multiple updates of option selection logic;
- [x] hovered option for `selectedTopShown` state after selection loses her hovered state, it goes to the previous option;

Need QA, expecially of keyboard navigation.

## Types of changes


packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
```diff
@@ -210,18 +225,35 @@ export default defineComponent({
...
+    const previousOptionComputed = computed((): SelectOption | undefined => {
+      const previousOptionIndex = currentOptionIndex.value - 1
+      const previousOption = currentOptions.value[previousOptionIndex]
+      const previousOptionCheck = isValueExists(previousOption) && !(previousOptionIndex === 0 && getDisabled(previousOption))
+
+      if (previousOptionCheck) {
+        return findNextActiveOption(currentOptionIndex.value - 1, true)
+      }
+
+      return undefined
```
rustem-nasyrov, line 237: Redundant `undefined`.

rustem-nasyrov (changes requested):
When I selected and deselect options by `enter/return` key multiple times I get this result:
image

packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
aluarius: It's not redundant -> eslint rules forces computed to have obvious return value.

Issue with `0` not related to this pr, will move it separately.

rustem-nasyrov (approved):

packages/ui/src/components/va-select/components/VaSelectOptionList/VaSelectOptionList.vue
m0ksem: How about `null`. It is not undefined, it is empty.

aluarius: It's not empty, it doesn't exist, undefined. Moreover, `findNextActiveOption` may return `undefined` if `find` method will find nothing. I don't think it's a good case to have one computed to return both `null` and `undefined` in same situations.

m0ksem (approved):

Issue #3053, opened 2023-02-23
For vuesic grid make it more visible it's deprecated


asvae:
> I was trying to use the grid recently and it wasn't working as expected so I went to look at issues and it looks like grid is not complete? Is that something that is on the road map for the future?

I would suggest having a bright disclamer + label in menu so that it's absolutely clear what's coming (and also that shouldn't use grids at all and page is here for people who're not ready for update).

https://vuestic.dev/en/styles/grid
https://vuestic.dev/en/styles/reset

PR #3130, opened 2023-03-07
[#3053] grid/reset deprecated alert color changed

aluarius:
Closes: #3053 

## Description
- [x] deprecated color changed to danger.

image

## Types of changes

### DIFF
--- a/packages/docs/page-config/styles/grid/index.ts
+++ b/packages/docs/page-config/styles/grid/index.ts
@@ -2,7 +2,7 @@ export default definePageConfig({
   blocks: [
     block.title("grid.title"),
     block.paragraph("grid.summaryText"),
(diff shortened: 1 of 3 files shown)


m0ksem:
I would make it solid, but blue.

packages/docs/modules/page-config/blocks/alert/index.vue
```diff
@@ -9,6 +9,10 @@ const props = defineProps({
   color: {
     type: String,
     default: 'info'
+  },
+  outline: {
```
m0ksem, line 13: Always prefer boolean props to be false by default. So it must be solid true. But it is minor.

m0ksem (approved):

PR #3135, opened 2023-03-07
Fix/dropdown border overflow

m0ksem:
Added shift that counts border.


aluarius (approved):
Only place where smthng wrong:

image

aluarius (approved):

PR #3139, opened 2023-03-07
Fix/button wrong focus, switching to CSS hover/pressed states

### LINKED ISSUE #3089: Button group has wrong focus styles
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
focus on button group

Visible outline for focused state

Less visible focus styles for secondary preset
Invisible focus styles for plain preset

https://user-images.githubusercontent.com/45711229/221526907-6bf445ec-ab3d-45a8-943d-8f5191da4446.mov

damirJa:
Closes #3089
## Description
- Changed the hover and pressed behavior from JS to CSS.
- Removed the use of additional :after and :before pseudo-elements.
- Created background helper functions that return 3 colors - general, hover, and pressed - and can switch from one to another.
- Managed background opacity through the alpha channel.
- Added a new Gradient background helper that can use alpha opacity.
- Modified the button group outline to work like all the others, without using pseudo-elements.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-button-group/VaButtonGroup.demo.vue
+++ b/packages/ui/src/components/va-button-group/VaButtonGroup.demo.vue
@@ -20,6 +20,69 @@
       </va-button-group>
     </VbCard>
 
+    <VbCard title="Presets & styles">
+      <va-button-group class="mb-6">
+        <va-button>One</va-button>
+        <va-button>Two</va-button>
+        <va-button>Three</va-button>
+      </va-button-group>
+
+      <va-button-group
+        preset="primary"
+        class="mb-6"
+      >
+        <va-button>One</va-button>
+        <va-button>Two</va-button>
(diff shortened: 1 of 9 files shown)


aluarius (changes requested):
Facing several critical issues, deep debugging is needed and then another one round of QA. Also should be compared with design (https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/(wip)-Vuestic-UI-library?node-id=9-34&t=rUHdtqZLCHNk1TPz-0).

- can't see any states (hover/pressed) for default preset (in demos, for some reason works for several buttons in docs, but with wrong appearance):
image

- plain hover appearance incorrect, pressed as well:
image

- pagination hover broken:
image
image


- for some reasons button toggle buttons are not clickable anymore.

- focus outline color should be dicussed with Vasyl (previous one was suggested by him).

PR #3142, opened 2023-03-08
[#3140] autocomplete readonly state fix

aluarius:
Closes: #3140 

## Description
- [x] readonly state fix;
- [x] additional related to disabled and readonly bugfix. 

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.demo.vue
+++ b/packages/ui/src/components/va-select/VaSelect.demo.vue
@@ -35,6 +35,32 @@
       <p>Value: {{ defaultSingleSelect.value }}</p>
     </VbCard>
     <VbCard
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-select/VaSelect.demo.vue
```diff
@@ -34,6 +34,32 @@
       />
       <p>Value: {{ defaultSingleSelect.value }}</p>
     </VbCard>
+    <VbCard
+      title="Autocomplete readonly"
+      style="width: 400px;"
+    >
+      <va-select
```
m0ksem, line 41: This can be replaced with tests :)

m0ksem (changes requested):

PR #3143, opened 2023-03-08
Feat/image usage example

### LINKED ISSUE #2952: va-image doesn't show images from assets in nuxt3
**Vuestic-ui version:** 1.5.3, nuxt module version 1.0.10

**Steps to reproduce**
In a nuxt3 project 
```
<template>
  <va-image src="~/assets/img/overview.png" />
</template>
```
does not show the image while a normal image tag (``) would render fine. 
Image should be rendered.
Image is not rendered.
**Other information**
There must be an issue with the correct path to the image. If I check the dev-tools, the path for the `img`-tag is `/_nuxt/assets/img/overview.png` while it stays unchanged with `va-img`.

rustem-nasyrov:
Close #2952

image

## Description
- Added how to information for nuxt apps.

## Types of changes


packages/docs/page-config/ui-elements/image/index.ts
```diff
@@ -19,6 +19,13 @@ export default definePageConfig({
     block.example("SrcSet"),
     block.example("Lazy"),
 
+    block.headline('image.imagesInNuxt.title'),
```
m0ksem, line 22: It is not really image in nuxt. The same way images must be imported in vite spa. You need to mention that "`va-image` behaves the same as img tag, so if you need to use image not form `/public` directory you will need to import it, so bundler like vite or webpack will create a correct img url."

m0ksem (changes requested):

Issue #2142, opened 2022-07-22
Dropdown must react different on events if it touch events. For example:


m0ksem:
Dropdown must react different on events if it touch events. For example:
- trigger hover must open on touch click, since there is no hover. 
- trigger right-click must work if it is long press.

_Originally posted by @m0ksem in https://github.com/epicmaxco/vuestic-ui/issues/2109#issuecomment-1192914113_

aluarius:
I'm planning to make a `useTouch` composable a bit later for swiping (carousel) and cases like this.

m0ksem:
Let's just make long press for now. Other features can be added later.

asvae:
I would start from research on this feature. We're not sure we need it.

PR #3150, opened 2023-03-10
Feat/2142 touch events

aluarius:
Closes: #2142

## Description
- [x] `useTouch` composable implementation;
- [x] integration to `VaDropdown`;
- [x] `useEvent` composable improvement;
- [x] refactoring and docs updates. 

## Types of changes


rustem-nasyrov (commented):
- Do `hover` state should work on touch?
- Need to fix demo layout for mobile screens.
- Missing demo for the vue-book.
image

packages/ui/src/composables/useTouch.ts
```diff
@@ -0,0 +1,88 @@
+import { ref, onBeforeUnmount, type Ref } from 'vue'
+
+import { useEvent, useWindow } from './'
```
m0ksem, line 3: Prefer direct import by path to prevent circular dependecies...

packages/ui/src/composables/useTouch.ts
```diff
@@ -0,0 +1,88 @@
...
+  ignoreDoubleClicks = false,
+  timings = { long: 500, double: 250 },
+) => {
+  const window = useWindow()
+
+  if (!window.value || !('ontouchstart' in window.value)) { return }
+
+  const previousTouchTime = ref(0)
+
+  let shortTouchTimer: NodeJS.Timeout | undefined
```
m0ksem, line 40: Types node may not be installed in user's project. 
```suggestion
  let shortTouchTimer: ReturnType<typeof setTimeout> | undefined
```

m0ksem: Would be nice to have eslint rule for that :)

packages/ui/src/composables/useTouch.ts
```diff
@@ -0,0 +1,88 @@
...
+  target: Ref<HTMLElement>,
+  callbacks: UseTouchCallbacks,
+  ignoreDoubleClicks = false,
+  timings = { long: 500, double: 250 },
+) => {
+  const window = useWindow()
+
+  if (!window.value || !('ontouchstart' in window.value)) { return }
+
+  const previousTouchTime = ref(0)
```
m0ksem, line 38: There is nothing depends on this reactive variable, so it can be a let.
```suggestion
  let previousTouchTime = 0
```

packages/ui/src/composables/useTouch.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  let shortTouchTimer: NodeJS.Timeout | undefined
+
+  const touch = (event: TouchEvent) => {
```
m0ksem, line 42:
```suggestion
  const onTouchStart = (event: TouchEvent) => {
```

packages/ui/src/composables/useTouch.ts
```diff
@@ -0,0 +1,88 @@
...
+    }, timings.long)
+
+    const clearTouchTimeouts = () => {
+      if (longTouchTimer) { clearTimeout(longTouchTimer) }
+      if (shortTouchTimer) { clearTimeout(shortTouchTimer) }
+    }
+
+    const cancelTouch = () => {
+      clearTouchTimeouts()
+      removeTouchCancelListeners()
```
m0ksem, line 60: useEvent must be used outside of onTouchStart function and there will no be needed to add removeTouchCancelListeners...

m0ksem (changes requested):
touch fn too complicated because it uses useEvent inside. Must be reworked.

PR #3152, opened 2023-03-10
refactor: rename dev env variable

### LINKED ISSUE #3151: variable name __DEV__ conflict with apollo graphql
**Vuestic-ui version:** 1.6.2

**Steps to reproduce**
build broken if used togather with nuxtjs/apollo

[commonjs--resolver] Unexpected keyword 'false' in
https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/composables/useColors.ts
```
5: import { u as useReactiveComputed } from "./useReactiveComputed.js";
6: import { a as getBoxShadowColor, f as getBoxShadowColorFromBg, c as getHoverColor, b as getFocusColor, g as getGradie...
7: import { _ as __DEV__ } from "../utils/env.js";
                   ^
8: const useColorProps = {
9:   color: {
```

**Other information**
apollo graphql depends on a predefined global variable `__DEV__`
nuxtjs/apollo use vite define feature to replace the variable with `false` in production build

My workaround is replace all `__DEV__` with `__DEV__VUESTIC` using `pnpm patch`

rustem-nasyrov:
Close #3151

## Description
- Renamed dev environment variable.

## Types of changes

### DIFF
--- a/packages/ui/src/utils/env.ts
+++ b/packages/ui/src/utils/env.ts
@@ -2,4 +2,4 @@ const processShim: typeof process = typeof process !== 'undefined' ? process : (
 const envShim = processShim.env || ({} as typeof process.env)
 const nodeEnv = envShim.NODE_ENV || ''
 
-export const __DEV__ = !['prod', 'production'].includes(nodeEnv)
+export const isDev = !['prod', 'production'].includes(nodeEnv)
--- a/packages/ui/src/utils/env.ts
+++ b/packages/ui/src/utils/env.ts
@@ -1,5 +1,8 @@
+// eslint-disable-next-line
+declare var __DEV__: boolean | undefined
+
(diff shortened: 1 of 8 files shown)


packages/ui/src/utils/env.ts
```diff
@@ -2,4 +2,4 @@ const processShim: typeof process = typeof process !== 'undefined' ? process : (
 const envShim = processShim.env || ({} as typeof process.env)
 const nodeEnv = envShim.NODE_ENV || ''
```
m0ksem, line 4: Do we want to handle `typeof __DEV__ !== 'undefined' ? __DEV__ : undefined`

m0ksem (changes requested):

m0ksem (approved):

PR #3154, opened 2023-03-13
Add color prop to VaDivider component and bind it to CSS border color

### LINKED ISSUE #3153: Divider is missing color prop
Solution: ```<va-divider :style="`--va-divider-border-top-color: ${dividerColor}`"/>```, but should have a `color` prop.

Yasir761:
Close #3153

…ers to specify the color of the divider. The color prop is then bound to the CSS border-color property using Vue's v-bind in CSS feature. This enables users to dynamically change the divider color using the color prop.

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-divider/VaDivider.vue
+++ b/packages/ui/src/components/va-divider/VaDivider.vue
@@ -4,6 +4,7 @@
     class="va-divider"
     :class="classComputed"
     :aria-orientation="vertical ? 'vertical' : 'horizontal'"
+    :style="`color: ${color}; --va-divider-border-top-color: ${color}`"
   >
     <div
       v-if="hasSlot && !vertical"
@@ -32,6 +33,7 @@ export default defineComponent({
       default: 'center',
       validator: (value: string) => ['left', 'right', 'center'].includes(value),
     },
+    color: { type: String, default: '' }
   },
   setup: (props, { slots }) => ({
     hasSlot: computed(() => !!slots.default),
@@ -81,7 +83,7 @@ export default defineComponent({
     flex: 1;
     border-top-width: var(--va-divider-line-width);
     border-top-style: var(--va-divider-border-top-style);
-    border-top-color: var(--va-divider-border-top-color);
+    border-top-color: var(--va-divider-border-top-color)
   }
 
   &--dashed {
--- a/packages/ui/src/components/va-divider/VaDivider.vue
+++ b/packages/ui/src/components/va-divider/VaDivider.vue
@@ -17,6 +17,7 @@
 <script lang="ts">
 import { defineComponent, computed, PropType } from 'vue'
 import { useComponentPresetProp } from '../../composables/useComponentPreset'
+import { useColors } from '../../composables/useColors'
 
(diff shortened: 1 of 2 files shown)


Yasir761:
Please let me know, if there is chance to make a change.

rustem-nasyrov (changes requested):
- Redundant `color` prop.
- Let's create a common CSS variable for the border color and use it for horizontal and vertical views. Tip: you can create CSS variable in the `_variables.scss` file at the va-divider directory.

m0ksem:
> * Redundant `color` prop.
> 
>     * Let's create a common CSS variable for the border color and use it for horizontal and vertical views. Tip: you can create CSS variable in the `_variables.scss` file at the va-divider directory.

I would prefer to have `color` prop, the same way we have with any other component (like button, input, swtich etc). Kind of intuitive to have this prop despite it can be done easily with css variables. It is more about DX.

packages/ui/src/components/va-divider/VaDivider.vue
```diff
@@ -32,6 +32,7 @@ export default defineComponent({
       default: 'center',
       validator: (value: string) => ['left', 'right', 'center'].includes(value),
     },
+    color: { type: String, default: 'backgroundBorder' },
```
rustem-nasyrov, line 35: The `default` value won't apply, because it is just a string. You need to handle this with `useColors` hook and after that it will be converted to the CSS color. [Example](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-avatar/VaAvatar.vue#L70-L78), and how it should be [applied](https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/components/va-avatar/VaAvatar.vue#L132) in the `style` tag.

rustem-nasyrov: Because of it we have a regression, here is difference:
<details>
<summary>production</summary>

image
</details>
and
<details>
<summary>your implementation</summary>

image
</details>

rustem-nasyrov (changes requested):
Keep it up! Just need to do a little fix and issue will be solved! :)

packages/ui/src/components/va-divider/VaDivider.vue
m0ksem: Thanks!

Yasir761: @rustem-nasyrov  Should I use the same hook from VaAvatar.vue by importing it or make a new one?

rustem-nasyrov: @Yasir761, you need to import it and use it like it done in `VaAvatar` component.

rustem-nasyrov (approved):

packages/ui/src/components/va-divider/VaDivider.vue
m0ksem: I fixed it to not bother @Yasir761 with it. 👌

PR #3171, opened 2023-03-14
fix(docs): remove client only from page config

m0ksem:


rustem-nasyrov (changes requested):
Great job. But there is a little issue with version titles on the roadmap page.

image

rustem-nasyrov (approved):

PR #3176, opened 2023-03-15
refactor: remove redundant language composable

### LINKED ISSUE #3169: Docs redirects
**Vuestic-ui version:** 1.6.2

Right now when you land on https://vuestic.dev/ru/introduction/overview, you're getting redirected. That should not be happening.

Here's my idea on how this could be done, but I suggest to check other libs docs and see how they did it.
```
/en -> /
/ -> localStorage language or en
/ru

not completely sure, we can check other solutions
```

rustem-nasyrov:
Close #3169

## Description
- Removed redundant language composable.
- Fixed routing redirects on direct navigation.

## TODO
- [ ] We need to migrate from `build:ci` script to the `build` one, because when we use `build:ci` it pre-renders every route and it breaks our routing.

## Types of changes


packages/docs/composables/useLanguageSwitcher.ts
```diff
@@ -1,58 +0,0 @@
...
-      router.push(path)
-    }
-  }
-
-  const currentLanguageName = computed(() => languages.find(({ code }) => code === locale.value)?.name)
-
-  const setHtmlLang = () => {
-    if (!document?.documentElement) { return }
-
-    document.documentElement.setAttribute('lang', locale.value || 'en')
```
m0ksem, line 44: Does nuxt module set lang tag?

packages/docs/modules/page-config/runtime/usePageConfig.ts
```diff
@@ -10,6 +14,13 @@ const files = Object.entries(import.meta.glob<false, string, PageConfigJSModule>
     if (name.split('/').length > 2) { return acc }
```
m0ksem, line 14: If name is `/en/page/something` then length of split by / will be bigger than 2, so code from line 19 will not be executed. I think better solution will be pass name without locale prefix to usePageConfig, instead of extending page configs dictionary with locale - it makes code logic more complex, because page configs now depends on locale...

m0ksem (changes requested):

PR #3177, opened 2023-03-15
Feat/confirm

### LINKED ISSUE #2972: useConfirm composable
Pretty basic thing:

```ts
const { showConfirm } = useConfirm()

const onDeleteButtonClick = () => {
   showConfirm({
       title: 'Are you sure',
       text: 'Deleting this item will destroy our universe',
       okText: `I don't care`
   }).then((ok) => ok && destroyUniverse())
}
```

Might be just wrapper around useModal.

In docs we might destroy CSS and throw a lot of errors if user click "I don't care".

m0ksem:
closes #2972

### DIFF
--- a/packages/docs/locales/en.json
+++ b/packages/docs/locales/en.json
@@ -1671,7 +1671,8 @@
     "colorsClasses": "Colors Classes",
     "skeleton": "Skeleton",
     "unocss": "UnoCSS Integration",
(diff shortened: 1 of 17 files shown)


rustem-nasyrov (changes requested):
API block is empty.
image

PR #3189, opened 2023-03-16
Feat/refactor docs language switcher

rustem-nasyrov:
## Description
- Refactor `useLanguageSwitcher` composable.
- Unify `language-dropdown` component.

## Types of changes


m0ksem (changes requested):
If I change language and reload a page, it will not be saved in localstorage, so if there is no prefix in url it always will be EN.

PR #3192, opened 2023-03-17
fix: issues with rating component

### LINKED ISSUE #3191: VaRating component issues
**Vuestic-ui version:** 1.6.2

- click on space btn triggers scroll down;
- value out of range;
- aria-label issues:
Screenshot 2023-03-17 at 10 10 21

- correct arial-label with correct item value and without backtick symbol
- blockers for values out of range

kushich:
closes #3191
## Description
- set limit for min and max values by using keyboard arrows
- fix aria-label attribute

## Types of changes


kushich:
@rustem-nasyrov but there is the point, arrows have the similar functionality with mouse, we can set value to zero only if we provided `clearable` prop

rustem-nasyrov (approved):

Issue #2924, opened 2023-02-06
Move to Histoire


asvae:
Right now component demos play important role by ensuring high component quality. We use vue-book for that.

The problem with vue-book is that it's not well supported. Fortunately, modern alternative exists, which is pretty much the same functionally. It's called [Histoire](https://histoire.dev/examples/tailwind.html) and it ships with native Vite support.

I reviewed it and I think it's a good fit for us.

This effort is useful for developers as well as @asvae, who'll have a talk on stories.

So here's the plan:
* #3293
* #3323, then @asvae should asyncronously settle things on his side:
  * #3324
  * #3325
- [ ] Play with Histoire for a couple of weeks, every developer could create a couple of demos and see if it works for them. Here's a couple of things we should probably pay attention to:
  * Is performance all right (Histoire uses iframes by default)?
  * Is it ok to debug?
  * and any other suspicion you may think of.
* If it doesn't work - we just rollback to vue-book. Dead end 👺.
* If it works we replace vue-book with Histoire. 
  * `.demo.vue` files could be renamed to `.story.vue`. 
  * Stories should be wrapped in `<story>` according to Histoire syntax. Stories should work without errors. Use `script setup`.
  * Remove vue-cli and anything vue-book related from `ui` package (files, scripts, deps).
  * Add information about histoire to contribution docs.
  * [book staging](http://vuestic-ui-demo.sub.asva.by/#/demo/va-modal/VaModal.demo.vue) should become histoire staging (modify docker file accordingly)

## Next steps
- [ ] Experiment with story/variant docs
- [ ] Add figma link, similarly to storybook
- [ ] Visual and e2e tests
- [ ] Onboarding guide should have detailed readme on how to work with storybook on conceptual level

PR #3225, opened 2023-03-23
feat: update vite

rustem-nasyrov:
Blocker for #2924

## Description
- Update vite version across project packages.
- Fixed paths for CSS imports.

## Types of changes
- [x] Might be breaking change


packages/ui/build/plugins/append-component-css.ts
```diff
@@ -27,15 +27,18 @@ export const appendComponentCss = createDistTransformPlugin({
...
   dir: (outDir) => `${outDir}/src/components`,
 
-  transform: (componentContent, path) => {
+  transform (componentContent, path) {
     if (!isVuesticComponent(path)) { return }
 
     const { name, dir } = parsePath(path)
 
-    const cssFilePath = `${dir}/${name}.css`
+    const distPath = resolve(this.outDir, '..', '..')
```
m0ksem, line 35: I wonder if we can use here the same outDir as in dir fn on line 29.

packages/ui/build/plugins/fabrics/create-dist-transform-plugin.ts
```diff
@@ -23,7 +23,7 @@ export const createDistTransformPlugin = (options: {
 
         const content = await readFile(currentPath, 'utf8')
 
-        const result = await options.transform(content, currentPath)
+        const result = await options.transform.call({ outDir }, content, currentPath)
```
m0ksem, line 26: outDir: options.build.outDir

asvae (approved):

packages/ui/build/plugins/append-component-css.ts
```diff
@@ -27,15 +27,18 @@ export const appendComponentCss = createDistTransformPlugin({
 
   dir: (outDir) => `${outDir}/src/components`,
 
-  transform: (componentContent, path) => {
+  transform (componentContent, path) {
```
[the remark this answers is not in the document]
rustem-nasyrov: Yes, because we provide an `outDir` value to the `transform` function.

Passing the `outDir` value to the function:
image

Getting value by `this.outDir`:
image

Issue #3123, opened 2023-03-04
VaDataTable filter per column


ralpheichelberger:
Please enable the data table to filter for multiple specific columns - ideally with optional regex functionality

**Describe alternatives you've considered**

The alternative would be to make lots of individual API calls for every search within a data table that is already fully loaded in the browser.
This is mostly slower, ineffective on the API side, and offline not possible.

asvae:
Hey,

can you explain how [current solutions](https://vuestic.dev/en/ui-elements/data-table#filtering) doesn't work for you? Maybe you have an example?

ralpheichelberger:
well if there is a number - e.g. EAN codes and on filter there are lines where those same first numbers are in a price, or a name etc. then the list is filtered but includes false positives

PR #3259, opened 2023-03-27
Feat/data table filter per column

rustem-nasyrov:
Related to #3123

## Description
- Added filter per column.
- Extended filter demo.

## Types of changes

### DIFF
--- a/packages/docs/page-config/ui-elements/data-table/examples/Filtering.vue
+++ b/packages/docs/page-config/ui-elements/data-table/examples/Filtering.vue
@@ -15,8 +15,14 @@
 
       <va-checkbox
         v-model="isDebounceInput"
-        class="d-block"
+        class="mb-3 d-block"
         label="Debounce input"
+      />
+
+      <va-checkbox
+        v-model="useFilterByColumn"
+        class="d-block"
+        label="Filter by column"
       />
     </div>
   </div>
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/va-data-table/hooks/useFilterable.ts
```diff
@@ -28,9 +28,15 @@ export const useFilterable = (
...
-        : cell.value.toLowerCase().includes(props.filter.toLowerCase())
+      if (typeof props.filterMethod === 'function') {
+        return props.filterMethod(cell.source)
+      }
+
+      if (!cell.column.filterable) {
+        return false
+      }
+
+      return cell.value.toLowerCase().includes(props.filter.toLowerCase())
```
RVitaly1978, line 39:
```suggestion
      if (!cell.column.filterable) {
        return false
      }
      return typeof props.filterMethod === 'function'
        ? props.filterMethod(cell.source)
        : cell.value.toLowerCase().includes(props.filter.toLowerCase())
```

RVitaly1978 (changes requested):

packages/docs/page-config/ui-elements/data-table/examples/Filtering.vue
asvae, line 1: image

When I click on that checkbox - the whole page freezes.

asvae, line 1: I suggest we have all table demos on the same page, you can split them into multiple components, but it should eventually be displayed on the same page.

The idea is that after change - we want to test the whole component just in case.

That's also a forward motion to histoire, where we'll have all cards split into files

packages/ui/src/components/va-data-table/VaDataTable-filter.demo.vue
```diff
@@ -0,0 +1,244 @@
+<template>
```
asvae, line 1: Can we make this demo shorter?

These are probably not needed:
* checkboxes
* debounce
* alert
* custom filtering function

Demo ideally should be as simple as possible.

packages/docs/page-config/ui-elements/data-table/examples/Filtering.vue
asvae, line 1: I checked several implementations of filter and naive ui does seem the most useful for me.

Maybe we can copy some implementation details from there.

https://www.naiveui.com/en-US/os-theme/components/data-table#filter-and-sorter

asvae, line 1: I rewrote demo and it doesn't really work for simplest case, which is probably something user would be expecting.

asvae (changes requested):
😁

PR #3276, opened 2023-03-29
fix: add string type for tabindex prop

### LINKED ISSUE #2895: Tabindex should accept string as type
We have in input this prop declaration:

```ts
tabindex: { type: Number, default: 0 },
```

But, tabindex also must be String

Might be issue with the same components.

kushich:
closes #2895

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/VaInput.vue
+++ b/packages/ui/src/components/va-input/VaInput.vue
@@ -124,7 +124,7 @@ export default defineComponent({
 
     // input
     placeholder: { type: String, default: '' },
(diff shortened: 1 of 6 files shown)


m0ksem (approved):
I wonder if it is possible to cover this with ts errors.

We could make defineFocusableComponent which must have a tabindex and focus, blur methods exposed. But I would also want to have defineValidatableComponent with validate method and errorMessages prop.

So , instead we can make

defineComponent() satisfies VaFocusableComponent satisfies VaFormComponent

PR #3295, opened 2023-04-03
Feat/a11y avatar

### LINKED ISSUE #3261: a11y: va-avatar
- [x] step 1: investigate a11y guides for this component;
- [x] step 2: discuss with team;
- [ ] step 3: implement required functionality;
- [ ] step 4: add docs page section about a11y.

rustem-nasyrov:
Close #3261

image

## Description
- Removed redundant a11y attributes.
- Refactored component.
- Added accessibility usage example.

## Types of changes


packages/ui/src/components/va-avatar/VaAvatar.vue
```diff
@@ -82,6 +83,12 @@ export default defineComponent({
     const { sizeComputed, fontSizeComputed } = useSize(props, 'VaAvatar')
     const { textColorComputed } = useTextColor()
 
+    const titleAttributeComputed = computed(() => {
```
m0ksem, line 86:
```
<VaAvatar title="You" role="button" @click="openProfile" />
```

I think I must be able to do that (not sure if it is correct a11y usage, btw)

m0ksem (approved):

m0ksem:
@misterpekert, mb you have something to add here. Otherwise, let's merge it.

packages/ui/src/components/va-avatar-group/VaAvatarGroup.vue
```diff
@@ -9,11 +9,11 @@
       :key="idx"
       v-bind="{ ...avatarProps, ...option }"
       role="listitem"
-      tabindex="0"
     />
     <slot name="rest" v-bind="avatarProps">
       <va-avatar
         v-bind="avatarProps"
+        color="secondary"
```
misterpekert, line 16: @rustem-nasyrov, do we always want to use the `secondary` colour here? Maybe, use `color` prop instead?

m0ksem: Nice catch. We need to pass here avatarProps and we want color to be secondary by default, so we need to redeclare color prop in definition.

m0ksem: I added just a new prop for that... It might be solved with css variables, but we don't usually do that.

PR #3322, opened 2023-04-08
fix: dropdown positioning problem

### LINKED ISSUE #2904: Revisit dropdown prevent-overflow false in fixed layout

### LINKED ISSUE #3083: Migrate Dropdown to floating-ui
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
Reduce browser height => click on colors dropdown

at least scroll for colors list 

https://user-images.githubusercontent.com/45711229/221502156-f091fffb-27aa-43c0-806b-eb3065ac1c19.mov

gweesin:
## Description
close #3083 
close #2904
closes https://github.com/epicmaxco/vuestic-ui/issues/3458
I replaced useDropdown with useFloating(`floating-ui/vue`) and implemented a simplified version.

But bro @m0ksem , there are still many details to be addressed:

1. the original placement being 'auto', but there is also an 'autoPlacement' property. Is there any difference between these two?
2. the autoPlacement property is not a real auto placement property, it likes flip (just reverse placement) but auto placement including all placements, right?

## Types of changes


m0ksem (changes requested):
This looks great!

We lost a feature which allows setting a root, that will be used to count auto-placement. Floating UI has an option for it: https://floating-ui.com/docs/detectOverflow#rootboundary ("Ref Target",  "Auto placement" demos.)

Speaking about autoPlacement, autoPlacement=true and placement=auto basically the same. It should work as you expect, not just flit sides. Previous implementation had a bug in it, so it behave more like flip. 
The idea of autoPlacement prop is to make it possible to move dropdown somewhere else if placement is top. E.g. with placement="left" and autoPlacement="true" if it doesn't fit left, it must be moved right ideally, but it also can be moved any available direction.

Also, would be nice to move all of this stuff from VaDropdown, to `useDropdown`, so we can use it in VaPopover (and in VaTooltip later).

Docs "Placement" and "Cursor" demos are broken.

Also, there are hydration mismatches which I will fix after this PR is merged, cuz it is a pain I'll not ask anybody to do.

Thanks! LMK if you need some help here.

gweesin:
> This looks great!
> 
> We lost a feature which allows setting a root, that will be used to count auto-placement. Floating UI has an option for it: https://floating-ui.com/docs/detectOverflow#rootboundary ("Ref Target",  "Auto placement" demos.)
> 
> Speaking about autoPlacement, autoPlacement=true and placement=auto basically the same. It should work as you expect, not just flit sides. Previous implementation had a bug in it, so it behave more like flip. 
> The idea of autoPlacement prop is to make it possible to move dropdown somewhere else if placement is top. E.g. with placement="left" and autoPlacement="true" if it doesn't fit left, it must be moved right ideally, but it also can be moved any available direction.
> 
> Also, would be nice to move all of this stuff from VaDropdown, to `useDropdown`, so we can use it in VaPopover (and in VaTooltip later).
> 
> Docs "Placement" and "Cursor" demos are broken.
> 
> Also, there are hydration mismatches which I will fix after this PR is merged, cuz it is a pain I'll not ask anybody to do.
> 
> Thanks! LMK if you need some help here.

Thank you for pointing out my mistakes.

I'll migrate the implementation of useDropdown and still useDropdown in dropdown component, and fix the ignores and broken points.

I have some remaining questions and would appreciate your assistance or discussion following your response.

Should we need placement can be setting 'auto'?

I believe that providing a placement or position information is essential, even in the case of autoPlacement, as it can provide a preference for orientation, for example, in the case of bottom placement, we would prioritize flipping upwards before other directions.

And if migrate these to useDropdown, I'll broke the original export variables instead of the information about x, y and update function to handle execute(or export all variables computePosition returned).

I have enjoyed collaborating with you, and looking forward to your response!

m0ksem:
> 
> Thank you for pointing out my mistakes.
> 
> I'll migrate the implementation of useDropdown and still useDropdown in dropdown component, and fix the ignores and broken points.
> 
> I have some remaining questions and would appreciate your assistance or discussion following your response.
> 
> Should we need placement can be setting 'auto'?
> 
> I believe that providing a placement or position information is essential, even in the case of autoPlacement, as it can provide a preference for orientation, for example, in the case of bottom placement, we would prioritize flipping upwards before other directions.
> 
> And if migrate these to useDropdown, I'll broke the original export variables instead of the information about x, y and update function to handle execute(or export all variables computePosition returned).
> 
> I have enjoyed collaborating with you, and looking forward to your response!

@gweesin, placement auto basically means "bottom" and autoPlacement="true". I understand your passion to make this thing perfect, so we can get rid of "auto", but in v1.7.0. Until then, it should work as it works now to prevent breaking changes. I'm OK with any change if it will not break existing apps.

You can freely change API of useDropdown composable, because we only use it internally, and do not expose it to the end user. (Here you can find what is exposed: https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/main.ts)

gweesin:
@m0ksem I'm so sorry that the cursor demos already broken now, I try to solve but I could not do it independently. The ResizeObserver could not work with Proxy Element. So, we need to discuss how to reconstruct the `useCursorAnchor`, for example like this:

```js
anchorRef.value = {
  getBoundingClientRect() { /* ... */}
}
```

Another problem is `flip` fn should work as all position, not only opposite position.

m0ksem:
> @m0ksem I'm so sorry that the cursor demos already broken now, I try to solve but I could not do it independently. The ResizeObserver could not work with Proxy Element. So, we need to discuss how to reconstruct the `useCursorAnchor`, for example like this:
> 
> ```js
> anchorRef.value = {
>   getBoundingClientRect() { /* ... */}
> }
> ```
> 
> Another problem is `flip` fn should work as all position, not only opposite position.

Hey. When I wanted to “check something” before answering you, but I just fixed this problem. So cursor anchor works well getter.

m0ksem:
Moved to https://github.com/epicmaxco/vuestic-ui/pull/3484 with resolved conflicts, etc.

PR #3379, opened 2023-04-27
[#3362] Sidebar on mobile

### LINKED ISSUE #3362: Navigation bar shown by default on mobile
**Vuestic-ui version:** 1.6.4
Screenshot_2023-04-22-00-37-10-99_40deb401b9ffe8e1df2f1cc5ba480b12

**Steps to reproduce**

Navigation bar should not cover content

**Other information**

Please check all screen sizes

RVitaly1978:
## Description
close #3362

<details>

https://user-images.githubusercontent.com/55198465/235647320-16f4ae11-6292-49d7-8a81-ba50ae506e7c.mov

</details>

## Types of changes


m0ksem (changes requested):
Makes it always hidden for me... We need maybe use local storage or cookie to show/hide sidebar.

misterpekert (changes requested):
The buttons to toggle sidebar and menu header are not clickable for me. Always hidden and seems that smth is broken :(

packages/docs/components/layout/header/HeaderSelector.vue
```diff
@@ -39,7 +39,7 @@ export default defineComponent({
   props: {
     minimized: {
       type: Boolean,
-      required: true,
+      default: false,
```
m0ksem, line 42: I think we 100% need this to be required, so we handle if sibebar can be toggled. It will be a bug to have this always default to false. If you want this flexable, you can change default to `undefined` and hide the whole button. We don't need this level of flexibility, so this prop must be required.

m0ksem (approved):

Issue #1430, opened 2022-01-20
New Component: Page layout


zvenigorodskaia:
# Layout components

General requirements:

- [ ] Mobile first.
- [ ] Adaptivity and responsivity.
- [ ] Independent nested components.
- [ ] Themeable.
- [ ] Customizability.
- [ ] Semantic markup, by default need to use the `header`, `main`, `aside` and `footer` HTML tags.
- [ ] Keep in mind that web also have the RTL and LTR reading and writing directions.
- [ ] CLI to generate ready layouts for the views.
- [ ] Documentation
- [ ] Tests
- [ ] Implement wireframes, ready to use examples (auth forms, simple admin panel).

## Features
- [ ] Layout builder (constructor), [example](https://quasar.dev/layout-builder).

### VaApp (VaLayout)

Component wrapper for the placing child components: `VaHeader`, [`VaSidebar`](https://vuestic.dev/en/ui-elements/sidebar) (implemented), `VaMain` и `VaFooter`).

### VaHeader
App header, can be used for the navigation.

Requirements:
- [ ] Should be placed on the top.
- [ ] Sticky on the scroll.
- [ ] Collapsible.
- [ ] Hide/show on the scroll.

### VaMain

For the router view.

### VaFooter

App footer, can be used for the navigation.

Requirements:
- [ ] Should be placed on the bottom.
- [ ] Sticky on the scroll.
- [ ] Collapsible.
- [ ] Hide/show on the scroll.

asvae:
Found an article regarding modern layouts: https://ishadeed.com/article/rebuild-featured-news-modern-css/

zvenigorodskaia:
![Image](https://github.com/epicmaxco/vuestic-ui/assets/35342240/1ea75caf-2953-49ea-aa02-189e4458aed0)

asvae:
We have a very small issues:


![Image](https://github.com/epicmaxco/vuestic-ui/assets/5394573/b7a939b1-81c8-4abc-90aa-cb253f5a06dd)

PR #3382, opened 2023-04-28
Feat/layout

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/1430

Features:
- [x] New component: VaLayout
- [x] top, bottom, left, right slots
- [x] order
- [x] animations support
- [x] absolute areas
- [x] mobile friendly
- [x] docs and examples

Default:
image

Order:
image

Absolute:
image


packages/docs/page-config/ui-elements/layout/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    block.paragraph(`
+VaLayout component is used in pair with [VaSidebar](/ui-elements/sidebar)[[target=_blank]], [VaNavbar](/ui-elements/navbar)[[target=_blank]], VaFooter components.
+    `),
+
+    block.example('Default', {
+      title: 'Default usage',
+      description: 'Default usage of Layout component.',
+    }),
+    block.example('AllSlots', {
+      title: 'All slots',
```
asvae, line 14: image

This example doesn't look too nice on mobile.

packages/docs/page-config/navigationRoutes.ts
```diff
@@ -267,9 +267,16 @@ export const navigationRoutes: NavigationRoute[] = [
         name: "color-input",
         displayName: "Color Input",
       },
-
+      
       {
         category: 'Layout',
+        name: 'layout',
```
asvae, line 273: We agreed to add a couple of examples at the beginning, similarly to https://www.naiveui.com/en-US/os-theme/components/layout

It seems to me that ready layout is something the user would hope to bring back from docs.

asvae (approved):
Looks good. I like that component is very minimal and easy to support.

props are not very intuitive, but it doesn't seem there is a way to improve that without a lot of examples / builder / extremely heavy component.

PR #3412, opened 2023-05-10
[#3279] VaDataTable docs improvement

### LINKED ISSUE #3279: Data table improvement suggestions
**Vuestic-ui version:** 1.6.3

From contact form:
> If possible, could you please update [Vuestic UI - Data table](https://vuestic.dev/en/ui-elements/data-table) to explain how to access the row data?
> 
> I was able to eventually find out how, but it was only due to a stack overflow post where the poster had to dig into the source code to figure it out.
> 
> I’ll link the S.O. post bellow for you. 
[vuejs3 - Access current row data in Vuestic UI Data Table from template - Stack Overflow](https://stackoverflow.com/questions/74807956/access-current-row-data-in-vuestic-ui-data-table-from-template)
While doing this, could you also document if there are any more properties you can access in the slot?
> 
> Also a quick suggestion, when using the pagination component it would be really nice to have dropdown next the paginator included for the desired page size. This of course can be accomplished other ways as you’ve shown in your demo, but it does look rather odd compared to other component libraries in Vue or even angular.

RVitaly1978:
## Description
close #3279

- [x] update docs api tables (events, slots)
- [x] update docs examples
- [x] minor types and styling updates for `VaValue` and block components (`alert`, `example`) 

<details>

image

image

https://github.com/epicmaxco/vuestic-ui/assets/55198465/28b76bd5-31d0-47c0-a369-084d313624a6

</details>

## Types of changes


packages/docs/page-config/ui-elements/data-table/examples/Selection.vue
```diff
@@ -107,15 +102,15 @@ export default defineComponent({
...
-      items: users,
+      items,
       columns,
       selectable: true,
-      selectedItems: [users[1]],
+      selectedItems: [items[1]],
       selectedItemsEmitted: [],
       selectMode: "multiple",
-      selectedColor: "#888888",
+      selectedColor: "#EF467F",
```
m0ksem, line 111: Let's have a primary by default, so people will not be scared of pink here.
```suggestion
      selectedColor: "primary",
```

packages/docs/page-config/ui-elements/data-table/examples/StaticSlots.vue
```diff
@@ -108,23 +84,26 @@ export default defineComponent({
...
       columns,
     };
   },
 });
 </script>
 
 <style lang="scss" scoped>
-.table-example-slots__head {
+.table-slots__head {
   background-color: var(--va-info);
```
m0ksem, line 96: It is still hard to read this demo. Let's remove extra text like "CUSTOM CELL WHICH SPAN ALL CELLS " and just show "#headAppend slot" in bold and primary color.

packages/docs/page-config/ui-elements/data-table/examples/Sorting.vue
```diff
@@ -30,25 +27,19 @@
   />
 
   <va-alert
```
m0ksem, line 29: I think it makes demo harder to read and not that important. Let's have a demo in VueBook instead.

packages/docs/page-config/ui-elements/data-table/examples/Styling.vue
```diff
@@ -1,130 +1,122 @@
 <template>
```
m0ksem, line 1: This a demo I always want to scroll down. It is too complicated. Thesee values in select can be hardcoded in example, so one read "column align left - OK!". We don't need that much interactivity here.

packages/docs/page-config/ui-elements/data-table/examples/Sticky.vue
```diff
@@ -60,9 +44,7 @@
     >
       <template #headerPrepend>
         <tr>
-          <th colspan="6">
-            With scroll events
-          </th>
+          <th colspan="6">With scroll events</th>
```
m0ksem, line 47: This is also complicated demo. Why do we need to have infinite scroll in demo related to sticky header? Just show a sticky header and we're done.

m0ksem (changes requested):
There are a lot of request changes from my personal opinion. I find DataTable docs complex and not interesting. I would make more small demos, then having few very big examples. We will add a navigation menu later, so we can scroll trough a lot of examples. 
These requests can be moved to a different PR and Issue. Overall looks much better then before.

m0ksem (approved):

m0ksem (approved):

PR #3429, opened 2023-05-17
Fix(TreeView filtering):

### LINKED ISSUE #3059: Filtering for the complex tree structures doesn't work as expected.
**Vuestic-ui version:** 1.6.1

**Steps to reproduce**
- Create treeView with filtering enabled.
- Add data with depth > 2
- Apply filter

Sandbox with the issue example
[https://codesandbox.io/p/sandbox/wizardly-galileo-2tpviu](https://codesandbox.io/p/sandbox/wizardly-galileo-2tpviu)

All nodes are filtered
image

Only root nodes are filtered
Снимок экрана 2023-02-24 в 09 49 30

**Other information**

misterpekert:
- fix getFilteredNodes method so it correctly filters child nodes.
closes #3059 

## Types of changes


packages/ui/src/components/va-tree-view/types.ts
```diff
@@ -10,6 +10,7 @@ export interface TreeNode {
   hasChildren?: boolean
   matchesFilter?: boolean
   indeterminate?: boolean
+  children: TreeNode[]
```
rustem-nasyrov, line 13: Shouldn't be required, see https://github.com/epicmaxco/vuestic-ui/issues/3077.

rustem-nasyrov (changes requested):

packages/ui/src/components/va-tree-view/types.ts
misterpekert: Hmm, okay. I made it optional. But now I think that we probably need different 'Types' for nodes. The first type is for user data, 2nd type is for our internal node structure. `buildTree` method always returns a node with the `children` property, but we definitely shouldn't force users to put empty arrays.

asvae: User tree node item I would assume should be generic on item type https://blog.vuejs.org/posts/vue-3-3#generic-components.

packages/ui/src/components/va-tree-view/hooks/useTreeView.ts
```diff
@@ -161,7 +161,7 @@ const useTreeView: UseTreeViewFunc = (props, emit) => {
   })
 
   const getFilteredNodes = (nodes: TreeNode[]): TreeNode[] => nodes.slice().filter((node) => {
-    if (node.hasChildren) { getFilteredNodes(getChildren(node)) }
+    if (node.hasChildren) { node.children = getFilteredNodes(node.children || []) }
```
asvae, line 164: Can't we separate `filter` and `map`? Looks very confusing to me.

asvae (changes requested):
Let's have another go at that. Doesn't look good.

PR #3431, opened 2023-05-18
feat#3392: make currentPresetName a writable computed

### LINKED ISSUE #3392: Make currentPresetName a writtableComputed
For now, we have applyPreset method that changes current preset.
Why don't we have DX like

```ts
const { currentPresetName } = useColors()

currentPresetName.value = 'dark'
```

This can be used in components, for example:
```html
<VaSwitch v-model="currentPresetName" true-value="dark" false-value="light" />
```

Evobaso-J:
# Make currentPresetName a WritableComputed property

## Description

closes #3392
A setter was added to the computed property `currentPresetName` in `packages/ui/src/composables/useColors.ts`.
This setter assigns a new string value to the `currentPresetName` property of `globalConfig.value.colors`.

## Markup:

<details>

```ts
 const currentPresetName = computed<string>({
    get: () => globalConfig.value.colors!.currentPresetName,
    set: (v:string) => { globalConfig.value.colors!.currentPresetName = v },
  })
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useColors.ts
+++ b/packages/ui/src/composables/useColors.ts
@@ -144,8 +144,10 @@ export const useColors = () => {
     return getColorLightnessFromCache(color) > globalConfig.value.colors.threshold ? darkColor : lightColor
   }
 
-  const currentPresetName = computed(() => globalConfig.value.colors!.currentPresetName)
-
+  const currentPresetName = computed<string>({
+    get: () => globalConfig.value.colors!.currentPresetName,
+    set: (v:string) => { globalConfig.value.colors!.currentPresetName = v },
+  })
   const presets = computed(() => globalConfig.value.colors!.presets)
 
   const applyPreset = (presetName: string) => {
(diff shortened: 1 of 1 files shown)


packages/ui/src/composables/useColors.ts
```diff
@@ -144,8 +144,10 @@ export const useColors = () => {
     return getColorLightnessFromCache(color) > globalConfig.value.colors.threshold ? darkColor : lightColor
   }
 
-  const currentPresetName = computed(() => globalConfig.value.colors!.currentPresetName)
-
+  const currentPresetName = computed<string>({
+    get: () => globalConfig.value.colors!.currentPresetName,
+    set: (v:string) => { globalConfig.value.colors!.currentPresetName = v },
```
RVitaly1978, line 149: I guess what we want to do is::
```suggestion
    set: (v: string) => { applyPreset(v) },
```

RVitaly1978 (changes requested):

m0ksem (approved):

RVitaly1978 (approved):

m0ksem:
Would be nice to have test coverage here to make sure applyPreset and currentPresetName work correct.

PR #3435, opened 2023-05-22
fix(docs): fix theme switch on SSG

m0ksem:
Now theme switcher work as expected in SSG, so we can switch our docs to SSG.

### DIFF
--- a/packages/docs/layouts/default.vue
+++ b/packages/docs/layouts/default.vue
@@ -1,5 +1,6 @@
 <template>
-  <div class="docs-layout">
+  <div class="docs-layout" :key="currentPresetName + isMounted">
+    <div v-if="!isMounted" class="docs-layout__loader" />
     <div
       ref="header"
       class="docs-layout__header"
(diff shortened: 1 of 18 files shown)


packages/ui/src/services/global-config/plugin/index.ts
```diff
@@ -4,11 +4,9 @@ import { PartialGlobalConfig } from '../types'
 import { defineGlobalProperty, defineVuesticPlugin } from '../../vue-plugin/utils'
 
 /** Provides global configuration to Vuestic components */
-export const GlobalConfigPlugin = defineVuesticPlugin((config: PartialGlobalConfig | undefined) => ({
+export const GlobalConfigPlugin = defineVuesticPlugin((config: PartialGlobalConfig | undefined = {}) => ({
```
rustem-nasyrov, line 7:
```suggestion
export const GlobalConfigPlugin = defineVuesticPlugin((config?: PartialGlobalConfig = {}) => ({
```

rustem-nasyrov (approved):

packages/docs/layouts/default.vue
```diff
@@ -31,16 +32,14 @@
 <script setup lang="ts">
 import { useColors } from 'vuestic-ui'
 import { useDocsScroll } from '../composables/useDocsScroll';
+import { useIsMounted } from 'vuestic-ui/src/composables/useIsMounted'
 
-const cookie = useCookie('vuestic-theme')
```
RVitaly1978, line 35: How do we set the theme that the user has last selected?
Instead we need to use localStorage. This can be done in this task or you need to open a new one.

RVitaly1978 (changes requested):

PR #3353, opened 2023-04-19
chore: Use tailwind instead of inline styles

a-3140:
Some notable changes that aren't 1:1 from inline styles to tailwind usage:

1. From padding-x: 30px -> px-8 ( px-8 is 32px )
2. height: 300px -> h-72 which is 288px
3. height: 120px -> h-32 which is 128px

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

PR #3447, opened 2023-05-27
docs: replace inline styles with tailwind classes in docs examples

### LINKED ISSUE #3321: Replace inline styles with tailwind classes
We need to get rid of inline style in all docs examples if it's possible. :) Context:
> There are several instances inside the `docs/page-config/ui-elements/` like this:
image
Can be replaced with tailwind `max-w-sm` which is 24rem ( 384px ) or if we want to maintain the 400px, `max-w-[400px]`.
There are other files within the folder that uses inline styles. I can create a new PR and transform them into tailwind classes instead of inline styles for code uniformity, if we're good to go with these changes?

_Originally posted by @a-3140 in https://github.com/epicmaxco/vuestic-ui/issues/3314#issuecomment-1500849782_

xiongmao86:
Handles docs sections other than ui-element. Partly solve #3321.

## Description
Replace inline styles with tailwind classes in examples

## Types of changes


packages/docs/page-config/styles/colors/components/components/PaletteGridCard.vue
```diff
@@ -7,6 +7,8 @@
       <p>
         {{ name }}
       </p>
+      <!-- Is there any way to specify font-size in px in tailwind? I found text-sm
+      change font-size in rem. change it to text-sm cause this p element to move.-->
       <p style="font-size: 14px;">
```
xiongmao86, line 12: `text-sm` changes both `font-size` and `line-height`, in the unit of `rem`. Tailwind can be configured but I don't know this is a good way, and even it is, I haven't found any `tailwind.config.js` exist under docs. Is there any way to solve this?

packages/docs/page-config/styles/css-variables/examples/profile.vue
```diff
@@ -25,10 +25,12 @@
         </div>
 
         <va-card-actions class="flex pt-2">
+          <!-- class -ml-2 and -ml-[0.5rem] makes the Follow button move left a little bit
+          how should I specify in class? -->
           <va-button
             preset="secondary"
             color="#fff"
-            style="margin-left: -0.5rem;"
+            class="-ml-[0.5rem]"
```
xiongmao86, line 33: Setting class to `-ml-2` and `-ml-[0.5rem]` cause the follow button to move left a little bit, why is that?

m0ksem:
Let's have it. https://github.com/epicmaxco/vuestic-ui/pull/3353 is missing styles for services and ag-grid, etc. (e.g. it has only fixed for ui-elements section).

xiongmao86:
Thanks very much, @m0ksem, I'll try to accommodate my branch.

xiongmao86:
All inline style of the document site is changed to tailwind class except ui-element. It's ready for review. Thanks.

m0ksem (approved):

m0ksem:
Looks perfect as always.

xiongmao86:
Thanks very much, @m0ksem.

Issue #3390, opened 2023-05-01
Automatically close sidebar when clicking outside of sidebar.


Bond-Addict:
**Describe the solution you'd like**
Please add a directive to automatically close Sidebar when you click outside of it (setting the v-model to false).

When you open the sidebar, if you click on the main content the sidebar should close.

**Current Solution:**
```
onMounted(() => {
  window.addEventListener('click', (e: any) => {
    if (e.target?.id !== 'menu' && e.target?.className !== 'va-sidebar__menu' && e.target?.className.indexOf('va-collapse') === -1 && show.value == true)
      show.value = false;
  });
});
``

PR #3455, opened 2023-05-31
Automatically close sidebar when clicking outside of sidebar

igorprodev1:
## Description

There is new prop in Sidebar which is called `close-on-click-outside` (false by default). Demo was added.

close: [#3390](https://github.com/epicmaxco/vuestic-ui/issues/3390)

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-sidebar/VaSidebar.vue
+++ b/packages/ui/src/components/va-sidebar/VaSidebar.vue
@@ -58,8 +58,6 @@ export default defineComponent({
   components: { VaConfig },
 
   setup (props, { emit }) {
-    const rootElement = shallowRef<HTMLElement>()
-
     const { getColor } = useColors()
     useSidebar(props)
 
@@ -98,6 +96,8 @@ export default defineComponent({
     const updateHoverState = (newHoverState: boolean) => {
       isHovered.value = props.hoverable && newHoverState
     }
+
+    const rootElement = shallowRef<HTMLElement>()
 
     useClickOutside([rootElement], () => {
       if (props.closeOnClickOutside && props.modelValue) {
--- a/packages/ui/src/components/va-sidebar/VaSidebar.vue
+++ b/packages/ui/src/components/va-sidebar/VaSidebar.vue
@@ -1,5 +1,6 @@
 <template>
   <aside
+    ref="rootElement"
     class="va-sidebar"
     :class="computedClass"
     :style="computedStyle"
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-sidebar/VaSidebar.vue
```diff
@@ -93,11 +99,20 @@ export default defineComponent({
       isHovered.value = props.hoverable && newHoverState
     }
 
+    useClickOutside([rootElement], () => {
```
m0ksem, line 102:
```suggestion
    const rootElement = shallowRef<HTMLElement>()

    useClickOutside([rootElement], () => {
```

packages/ui/src/components/va-sidebar/VaSidebar.vue
```diff
@@ -49,11 +50,16 @@ export default defineComponent({
...
   },
 
+  emits: ['update:modelValue'],
+
   components: { VaConfig },
 
-  setup (props) {
+  setup (props, { emit }) {
+    const rootElement = shallowRef<HTMLElement>()
+
```
m0ksem, line 62:
```suggestion
```

m0ksem (approved):

packages/ui/src/components/va-sidebar/VaSidebar.vue
```diff
@@ -93,11 +97,22 @@ export default defineComponent({
       isHovered.value = props.hoverable && newHoverState
     }
 
+    const rootElement = shallowRef<HTMLElement>()
+
+    useClickOutside([rootElement], () => {
+      if (props.closeOnClickOutside && props.modelValue) {
+        setTimeout(() => {
```
m0ksem, line 104: Why is setTimeout here?

igorprodev1: It fixes an issue when you try to hide sidebar by clicking to checkbox, in this case we have two sources that change modeValue and Sidebar can't be closed. setTimeout with 0 delay creates macrotask, and this means that close-outside handler does't fire update:modelValue immediately. So, the delay is enough to change modelValue from checkbox source (or button) and after that click-outside handler can't brake right behaviour

m0ksem: Maybe add `if modelValue != false  { emit(modelValue, false)` to prevent firing update event twice?

Issue #3442, opened 2023-05-26
Remove CSS variable from root(html element)


shiina7:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
1. vuestic-ui places CSS variable in the style attribute of the root(html) element.

2. that's no problem, but only when there are few CSS variables, when there are many (e.g. integrated tailwindcss).

3. when I use browser(e.g. chrome) debug, CSS variables will fill the console.

4. I know it doesn't affect the functionality, but the experience is very bad (at least for me it affects my debugging).

5. Please move these CSS variables to a style tag if you can.

**Describe alternatives you've considered**

**Additional context**
Maybe this can be changed through configuration but I missed it, please let me know thanks.

shiina7:
There is a problem with gitbub which can't upload pictures right now, I will add screenshots later.

m0ksem:
Yeah. We 100% can do that. Must be easy to change "updateColors" method. 
https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/services/color/plugin/create-color-config-plugin.ts

shiina7:
> Yeah. We 100% can do that. Must be easy to change "updateColors" method. https://github.com/epicmaxco/vuestic-ui/blob/develop/packages/ui/src/services/color/plugin/create-color-config-plugin.ts

I create a pull request #3466 for this issue. If you check after no problems please merge it. Thanks.

shiina7:
added new feature for this. #3466

PR #3466, opened 2023-06-02
feature: support configure css variables generation location

shiina7:
# support configure css variables generation location

## Description

pull request for #3442 
support configure css variables generation location, root element(html) or new style tag.

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useColors.ts
+++ b/packages/ui/src/composables/useColors.ts
@@ -162,6 +162,8 @@ export const useColors = () => {
     }
   }
 
+  const styleTag = computed(() => globalConfig.value.colors?.styleTag ?? false)
+
   return {
     colors,
     currentPresetName,
@@ -182,6 +184,7 @@ export const useColors = () => {
     colorsToCSSVariable,
     colorToRgba,
     getStateMaskGradientBackground,
+    styleTag,
   }
 }
 
--- a/packages/ui/src/services/color/plugin/create-color-config-plugin.ts
+++ b/packages/ui/src/services/color/plugin/create-color-config-plugin.ts
@@ -1,16 +1,22 @@
 import { PartialGlobalConfig } from './../../global-config/types'
 import { ColorVariables } from './../types'
-import { App, watch } from 'vue'
+import { App, watch, computed } from 'vue'
 import { isServer } from '../../../utils/ssr'
 import { cssVariableName } from '../utils'
 import { useColors } from '../../../composables'
+import { generateUniqueId } from '../../../utils/uuid'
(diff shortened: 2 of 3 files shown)


shiina7:
> This is look great.
> 
> We also need to handle it in `nuxt/packages`. So instead of htmlAttrs we need to use content in `useHead`. Otherwise both style tag and html attribute will be rendered.

No, html arribute is default. It just adds an optional configuration so that we can choose to transfer the color variables to a new style tag.
I don't think that I can decide where the color variables are generated (this decision should be done the vuestic-ui's owner). so I just add new optional configuration for those who need it.

shiina7:
I got some screenshot for this.

1. before, all color variables in the root (html) attributes.

Snipaste_2023-06-06_15-05-54

2. when configuration `styleTag` is `true`. It's look tidy better.

Snipaste_2023-06-06_15-04-57

3. and the color variable will be placed in the new generated style tag.

Snipaste_2023-06-06_15-05-25

m0ksem:
@shiina7, would you like to write docs for this feature under Services/ColorConfig?

shiina7:
> @shiina7, would you like to write docs for this feature under Services/ColorConfig?

@m0ksem sure, I'd love it! I will finish it in two days (before Thursday). Thank you for your additions for nuxt and merge request.

asvae:
@shiina7 does it make any point to keep previous solution and not just keep yours? (and get rid of `styleTag`)

shiina7:
> @shiina7 does it make any point to keep previous solution and not just keep yours? (and get rid of `styleTag`)

@asvae my first point is to expend it, not change. beacuse I think that I can't decide it, but you(author) can. If you think it's okay to just keep the new style tag, I'm glad to change it.
But I think it's up to you to decide it. That's why added an option to configuration.

shiina7:
> > @shiina7 does it make any point to keep previous solution and not just keep yours? (and get rid of `styleTag`)
> 
> @asvae my first point is to expend it, not change. beacuse I think that I can't decide it, but you(author) can. If you think it's okay to just keep the new style tag, I'm glad to change it. But I think it's up to you to decide it. That's why added an option to configuration.

Perhap someone needs this option and we can keep, If we want keep new tag we can set the `styleTag` to true by default. BTW

m0ksem:
I think we can remove previous solution. It was made asap and not documented, so we can just drop it.

shiina7:
> I think we can remove previous solution. It was made asap and not documented, so we can just drop it.

Okay, got it. It's will be change it tomorrow.

m0ksem (approved):

PR #3469, opened 2023-06-06
feat(ui): useDropdownable composable and close-on-value-update prop

### LINKED ISSUE #3459: Date Input `close-on-value-updated` prop
**Vuestic-ui version:** 1.6.5
**Steps to reproduce**
Set date input behavior using close-on-content-click.
You should be able to cycle through months using the arrows and clicking on a date should close the popup and select the date.
You cannot cycle through the months, but clicking on a date does close the popup and select the date.

m0ksem:
closes #3459

Moved all stuff for components like VaTimeInput, VaDateInput, VaSelect to separate composable. Might be used later in VaColorPicker, etc.


m0ksem:
Check if it works with manual-input pickers.

packages/ui/src/composables/useDropdownable.ts
```diff
@@ -0,0 +1,36 @@
...
+import { VaDropdown } from '../components/va-dropdown'
+
+const VaDropdownProps = extractComponentProps(VaDropdown,
+  ['innerAnchorSelector', 'stateful', 'keyboardNavigation', 'modelValue', 'ariaLabel'],
+)
+
+export const useDropdownableProps = {
+  ...VaDropdownProps,
+  modelValue: {},
+  closeOnValueUpdate: { type: Boolean, default: false },
```
asvae, line 13: I think this should be 3 states: `false`, `true`, `null`. 
* `false` - would never close on select
* `true` - close on select
* `null` - default component behavior

Here's couple of examples why this matters.

For select:
* single select should close on input by default
* multiselect should **not** close on input by default

For time picker
* when you have only one column - we need to close on input by default
* when you have multiple columns - we shoudn't close by default

Probably also applicable to date picker - where for single non range input we want to close by default.

asvae: Also I read the user issue and it seems their concern was with defaults, not with ability to force closing on select.

asvae (changes requested):
Let's spend more time on this.

packages/ui/src/composables/useDropdownable.ts
m0ksem: Defaults can be configured trough GlobalConfig.

asvae: @m0ksem they can't be if you need different defaults depending on other props.

packages/ui/src/components/va-date-input/VaDateInput.demo.vue
```diff
@@ -192,6 +192,10 @@
       <va-button @click="focusDateInput">Focus</va-button>
       <va-button @click="blurDateInput">Blur</va-button>
     </VbCard>
+
+    <VbCard title="Close on value update">
+      <va-date-input v-model="value" close-on-value-update />
```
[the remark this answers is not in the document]
asvae: Do we have a reason to prefer "close-on-update" to "close-on-select"?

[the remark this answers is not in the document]
m0ksem: The only that it behaves on value update, not actual select. So if value changed outside of component (no select from user) dropdown is closed.

PR #3477, opened 2023-06-07
fix(docs): add vue-gtag

### LINKED ISSUE #3476: Improving vuestic analytics
Right now vuestic analytics are working through some unknown ways (and also not exactly correctly).

### Problems
* Analytics are not atached to vuestic.dev.
* We don't have a process around analytics, though they can provide us with useful numbers on docs improvements. Right now we only check them on big releasese and that's about it

### Doables
- [ ] Attach analytics to vuestic.dev.
<details>
  <summary>For that we would need to follow instructions</summary>
  Below is the Google tag for this account. Copy and paste it in the code of every page of your website, immediately after the 
  <head> element. Don’t add more than one Google tag to each page.

  ```html
  <!-- Google tag (gtag.js) -->
  <script async src="https://www.googletagmanager.com/gtag/js?id=G-G4S900KP26"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
  
    gtag('config', 'G-G4S900KP26');
  </script>
  ```

  Use Google Tag Manager?
  For additional features including multiple tag management and enterprise workflows, install and manage with Google Tag 
  Manager. [Learn More](https://support.google.com/tagmanager/answer/12811173)
</details>

Analytics should only be enabled on all environments (so that we can test without pushing to prod). Key should be passed from railway variables. Here's an example of variables:

```
GTM_ENABLED=true
GTM_KEY=G-G4S900KP26
```
- [ ] Maybe add a redirect from `/get-started`, for some reason we're getting a lot of traffic to that url.
- [ ] Create a process to have analysis once in a month (maybe sync with retro).

m0ksem:
closes #3476

### DIFF
--- a/packages/docs/nuxt.config.ts
+++ b/packages/docs/nuxt.config.ts
@@ -70,8 +70,13 @@ export default defineNuxtConfig({
     compressPublicAssets: true,
   },
 
+  googleAnalytics: {
+    id: 'G-G4S900KP26'
+  },
+
   modules: [
     // './modules/banner',
+    './modules/google-analytics',
     './modules/vuestic',
     './modules/page-config',
     '@nuxtjs/google-fonts',
--- a/packages/docs/modules/google-analytics/index.ts
+++ b/packages/docs/modules/google-analytics/index.ts
@@ -1,0 +1,1 @@
+export { default as default } from './module'
--- a/packages/docs/modules/google-analytics/module.ts
+++ b/packages/docs/modules/google-analytics/module.ts
@@ -1,0 +1,23 @@
+import { defineNuxtModule, addPluginTemplate } from '@nuxt/kit';
+import { resolve } from 'pathe'
+
+/** Module used to add current file path to page-config */
+export default defineNuxtModule<any>({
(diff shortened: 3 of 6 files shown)


packages/docs/nuxt.config.ts
```diff
@@ -70,8 +70,13 @@ export default defineNuxtConfig({
     compressPublicAssets: true,
   },
 
+  googleAnalytics: {
+    id: 'G-G4S900KP26'
```
asvae, line 74: This should go to railway variables. No need to have this locally.

asvae:
> closes https://github.com/epicmaxco/vuestic-ui/issues/3476

It only solves one task out of 3 in the issue...

asvae:
I updated the issue to mention we should support all environments (except local)

asvae:
Seems good. We can update it if something doesn't work in prod.

asvae (approved):

asvae:
I used the same key fall all apps.

PR #3481, opened 2023-06-08
docs: structure contribution docs

### LINKED ISSUE #1907: Vuestic contribution manual
Docs for contributors.

In progress and must be moved to vuestic.dev as page, not issue.

m0ksem:
closes #1907

image


asvae (approved):
I updated docs text (was easier to fix than to explain) :). Hope that's ok.

I'm not completely sure about structure, but we can fix that after we get that understanding.

PR #3484, opened 2023-06-08
Feat/floating UI migartion 3083

ilearnjs:
## Description
Migrate Dropdown to floating-ui (#3083)

Before converting from the draft:
- [x] test once again
- [x] to check other pr's can be fixed/affected after the new implementation
- [x] ~to fix inner anchor issue (probably the old one) **Need help** 🆘~

## What is going next?
After initial code review and approve we are going to replace VaDropdown component with the current implementation.

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes


m0ksem (changes requested):
Placement prop is not reactive: see docs placement example (try to play with select).

packages/ui/src/components/va-dropdown/VaDropdown.demo.vue
```diff
@@ -113,6 +113,9 @@
         <va-dropdown
           anchor-selector="#anchor-selector"
         >
+          <template #anchor>
```
m0ksem, line 116: Should work without #anchor slot, because you have anchor-selector. Maybe remove warning and let it be empty.

packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -174,14 +174,15 @@ export default defineComponent({
     })
 
     const anchorComputed = computed(() => {
-      return props.cursor ? cursorAnchor.value : anchor.value
+      if (!props.cursor) { return anchor.value }
+
+      return useCursorAnchor(anchor, valueComputed).value
     })
```
m0ksem: Composables mustn't be used outside setup() execution flow. useCursorAnchor will create computed on each recalculation of anchorComputed. Computed stores effect globally, so there might be potential memory leaks.

```ts
    const cursor = useCursorAnchor(anchor, valueComputed)

    const anchorComputed = computed(() => {
      if (!props.cursor) { return anchor.value }
      return cursor.value
    })
```

m0ksem (changes requested):

packages/ui/src/components/va-dropdown/VaDropdown.vue
```diff
@@ -1,88 +1,64 @@
...
-      default: 'click',
-      validator: (value: string) => ['click', 'right-click', 'hover', 'dblclick', 'none'].includes(value),
-    },
+    autoPlacement: { type: Boolean, default: true },
+    stickToEdges: { type: Boolean, default: false },
+    /** Viewport where dropdown will be rendered if preventOverflow is true. Autoplacement will be calculated relative to `target` */
+    target: { type: [String, Object] as PropType<MaybeHTMLElementOrSelector>, default: undefined },
+    /** Element where dropdown content will be rendered. */
+    teleport: { type: [String, Object] as PropType<MaybeHTMLElementOrSelector>, default: undefined },
+    preventOverflow: { type: Boolean, default: false },
```
m0ksem, line 53: Let's turn it on by default and maybe remove this prop, because it doesn't make any sense to not prevent overflow...

m0ksem (approved):

PR #3495, opened 2023-06-14
fix: change title and subtitle styles to look different on mobile resolution

### LINKED ISSUE #3486: On mobile title and subtitle look the same
**Vuestic-ui version:** 1.6.5

They should look different. Maybe make subtitle and title proportionally smaller on mobile? It's even more confusing given line height stays the same.

image

Here's from desktop:
image

pogrib0k:
closes #3486

I changed the `title` and `sub-title` styles to look different on mobile resolution.

### On desktop:
image

### On mobile:
image


asvae:
Excellent! Thanks.

PR #3505, opened 2023-06-19
Feat/extra vuesitc plugins

m0ksem:
Removed grid styles
Removed colors helper plugin from default registration, but it is still available for registration.


m0ksem:
TLDR: Tailwind is going to be fully replacement for our grid and color classes plugin.

This PR removes grid.css and color helpers classes like `.va-primary`. It makes sense if we assume that user will depend on Tailwind. We'll need to update create-vuestic and docs.

asvae:
> We'll need to update create-vuestic and docs

I suggest we do this in context of this PR. As it is - it's incomplete.

PR #3516, opened 2023-06-20
fix: props descriptions

asvae:
After translations removal all descriptions disappeared. I tried to fix it, but typing is somewhat broken with a number of loose ends.

So @m0ksem maybe you have a context to finish this one without much of a trouble?

That's mission critical for release!

image

Note:

image
Here we define meta as an object, but we get array in the prop!


m0ksem (approved):
I'd like to combine api description and manual options into one config file, but it looks OK for now.

asvae:
> I'd like to combine api description and manual options into one config file, but it looks OK for now.

Could you create an issue for that?

asvae:
Thanks!

PR #3520, opened 2023-06-21
feat(docs): add redirect from language pages

ilearnjs:
Add redirects from language pages

## Description
Remove language code from the url

## Types of changes

### DIFF
--- a/packages/docs/middleware/redirects.global.ts
+++ b/packages/docs/middleware/redirects.global.ts
@@ -1,0 +1,15 @@
+// Is used to test following paths /en, /en/, /en/something, /ru, /ru/, /ru/something
+const langTestRegex = /^\/(en|ru)\/?$|^\/(en|ru)\/.*/i
+// Is used to replace /en, /en/, /ru, /ru/ from the beginning of the path
+const langReplaceRegex = /^\/(en|ru)\/?/i
+
+export default defineNuxtRouteMiddleware((to) => {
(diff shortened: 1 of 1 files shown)


packages/docs/middleware/redirects.global.ts
```diff
@@ -0,0 +1,15 @@
+// Is used to test following paths /en, /en/, /en/something, /ru, /ru/, /ru/something
+const langTestRegex = /^\/(en|ru)\/?$|^\/(en|ru)\/.*/i
```
m0ksem, line 2:
```suggestion
const langTestRegex = /^\/(en|ru)/i
```

If this regex is used in `.test` it checks if string starts with `/en` or `/ru`. Removed part doesn't make sense to me.

You can use `/^\/(en|ru)\/?/i` for both check if string starts with prefix and to remove this prefix.

m0ksem (approved):

packages/docs/middleware/redirects.global.ts
ilearnjs: It doens't work for the case when you have something like `/rumba` or `/end`. It will be replaced with `/mba` and `/d`

ilearnjs: There is an exact check for `/en` or `/en/` OR for `/en/something`

asvae (approved):
Checked on experimental - looks good.

PR #3522, opened 2023-06-22
fix: make focus input more visible

### LINKED ISSUE #3512: Input colors do not work well with selection
On screenshot second input is on focus, yet the selection is barely visible. We should make it more pronounced.

image

I think we should reuse highlight [from other elements](https://www.figma.com/file/2yH0KFiWds3q1B5NA5dELZ/(wip)-Vuestic-UI-library?type=design&node-id=636-1036&t=NXMezl3pYN2YBZuL-4). 

@xx13 maybe you have a comment on that.

xiongmao86:
Make focus input more visible.

Fixes #3512.

## Types of changes


asvae:
Because of outline increase - input changes size a bit. That's not great.

I suggest we wait for design before progressing with that - it's not exactly clear to me what implementation would look like.

xiongmao86:
Sure, thanks @asvae.

PR #3523, opened 2023-06-22
fix(ui): select, date-input, time-input keyboard navigation

m0ksem:
Previously it was impossible to open select, date-input, time-input with space and enter for some reason.
closes https://github.com/epicmaxco/vuestic-ui/issues/3533
closes https://github.com/epicmaxco/vuestic-ui/issues/3530


packages/ui/src/components/va-date-input/VaDateInput.vue
```diff
@@ -133,7 +133,7 @@ import { VaIcon } from '../va-icon'
 const VaInputWrapperProps = extractComponentProps(VaInputWrapper, ['focused', 'maxLength', 'counterValue'])
```
asvae, line 133: image

Date input doesn't close on click outside.

asvae: Same issue for time input

image

asvae (changes requested):
Checked locally - there are some issues.

pogrib0k (changes requested):
I looked it up like everything is ok.
But `DateInput` allows focus outside the component, so sometimes something like this happens:
image

Also, because of this, I can open many "dropdowns" at the same time (see screenshot), and when I press **ESC** they all close at the same time.
image

ilearnjs (approved):

PR #3524, opened 2023-06-22
feat(ui): dropdown remove prevent overflow property

ilearnjs:
Remove VaDropdown  prevent-overflow property

## Types of changes

### DIFF
--- a/packages/docs/components/ComponentPlayground.vue
+++ b/packages/docs/components/ComponentPlayground.vue
@@ -79,7 +79,6 @@ const doShowComponent = false; // TODO: Temporarily disabled
             :label="option.key"
             :rules="option.rules"
             clearable
-            prevent-overflow
           />
           <va-select
             v-if="option.type === 'multiselect'"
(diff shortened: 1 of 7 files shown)


m0ksem (approved):

m0ksem:
@asvae, this prop is useless, so we're going to remove it without deprecation warning. It actually does nothing and will be turned on by default.

ilearnjs:
After we moved dropdowns to the document body sometimes it is visible they change positions not instantly (e.g. after scroll).
If you think that is an issue it is possible to try this: https://floating-ui.com/docs/autoUpdate#animationframe
Vue example from docs:
```
whileElementsMounted(...args) {
    const cleanup = autoUpdate(...args, {animationFrame: true});
    // Important! Always return the cleanup function.
    return cleanup;
  },
```

asvae (commented):
Checked locally. Things seem to work ok, but didn't delve too deep.

PR #3538, opened 2023-06-23
feat: automated display of badges

### LINKED ISSUE #3528: Automated display of badges
Right now for pages we use badges like this:

```js
{
  name: "confirm",
  displayName: "Confirm",
  meta: {
    badge: {
      type: 'new',
      text: 'new',
      title: 'Recently added',
    },
  }
},
```

We then have to manage them and delete accurately before release. It's also not entirely clear when this badge was applied.

- [ ] What I suggest to do instead is to declare these conditions instead.

```
badge: navigationBadge.new('1.6.6'),
```

What that line should do is - until new major version (`1.7.0`) show "new" badge. After `1.7.0` - hide that badge.

- [ ] We would also need to apply this to other label types.

Conceptually the idea is to solve this problem in advance. We'll also have additional information about the version.

- [ ] Let's also remove `GENERATOR_ADD` lines in context of this issue, we're not using them.

Also see https://github.com/epicmaxco/vuestic-ui/issues/3504. It's somewhat tied to release and deprecation schedule.

thesheppard:
## Description

closes #3528 
closes https://github.com/epicmaxco/vuestic-ui/issues/3546

## Types of changes

### DIFF
--- a/packages/docs/components/layout/Sidebar.vue
+++ b/packages/docs/components/layout/Sidebar.vue
@@ -50,7 +50,7 @@
                   placement="right-center"
                   size="small"
                   offset="-5px"
-                  :text="childRoute.meta && childRoute.meta.badge.text"
+                  :text="childRoute.meta?.badge?.text"
                   :color="childRoute.meta && childRoute.meta.badge && badgeColors[childRoute.meta.badge.type]"
                   :visible-empty="false"
                 >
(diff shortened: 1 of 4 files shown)


thesheppard:
@asvae Not too sure how to assign somebody to this PR
Tagging you since you created the issue 🖖🏾

m0ksem:
Hi, @thesheppard! Thanks for PR! I simplified a bit these changes. It will be much easier for us to have badge logic outside of navigationRoute file, so we store only configuration here.

m0ksem (approved):

Issue #3506, opened 2023-06-19
va-radio component does not support validation


sir-kain:
I could add the validation support if someone can agree with this issue

asvae:
Hey @sir-kain, that would be absolutely nice to have 😁.

PR #3542, opened 2023-06-25
Add validation support for the va-radio component

sir-kain:
## Description

Following #3506 , this PR enables validation for the `va-radio` component using attributes such as `rules`, `error`, and `error-messages`, similar to other components like `va-input`, `va-checkbox`, etc.

## Types of changes


packages/ui/src/components/va-radio/VaRadio.vue
```diff
@@ -1,116 +1,183 @@
...
     }))
 
+    const selectOption = (option: option['value']) => {
+      emit('update:modelValue', option)
+    }
+
+    const labelStyle = computed(() => {
+      return {
+        color: computedError.value ? getColor('danger') : '',
+        padding: props.leftLabel ? '0 0.5rem 0 0' : '0 0 0 0.5rem',
```
m0ksem, line 124: Would be nice to move it to CSS variable.

```suggestion
        padding: props.leftLabel ? '0 var(--va-radio-label-padding) 0 0' : '0 0 0 var(--va-radio-label-padding)',
```

packages/ui/src/components/va-radio/VaRadio.vue
```diff
@@ -1,116 +1,183 @@
...
-      aria-hidden="true"
-      class="va-radio__icon"
-      :style="iconComputedStyles"
-    >
-      <span
-        class="va-radio__icon__background"
-        :style="iconBackgroundComputedStyles"
+      <input
+        ref="input"
+        :style="inputStyle"
```
m0ksem, line 21: I know it was as it before, but it would be nice to move styles to `v-bind` in css. I will make a different issue for it.

packages/ui/src/components/va-radio/VaRadio-validation.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <VbCard title="va-input validation">
+    <va-radio
```
m0ksem, line 3: Let's have this demo in the `VaRadio.demo.vue`. Later, we will do a script to transform all of these demos to storybook, so it is better it the way other demos have.

packages/ui/src/components/va-radio/types.ts
```diff
@@ -0,0 +1,4 @@
+export type option = {
```
m0ksem, line 1:
```suggestion
export type VaRadioOption = {
```

We can export this type to the end user.

```ts
const options: VaRadioOptions[] = []
```

m0ksem (approved):

igorprodev1 (approved):

PR #3589, opened 2023-07-09
fix: node 16 module resolution error

bsssshhhhhhh:
## Description

Adds a "types" field to the `exports` block in package.json to allow TypeScript's Node16 resolution to work properly.

Without it, TypeScript throws this error:

```
Could not find a declaration file for module 'vuestic-ui'. '/workspace/node_modules/vuestic-ui/dist/es/main.js' implicitly has an 'any' type.
  There are types at '/workspace/node_modules/vuestic-ui/dist/types/main.d.ts', but this result could not be resolved when respecting package.json "exports". The 'vuestic-ui' library may need to update its package.json or typings
```

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -148,7 +148,8 @@
     ".": {
       "import": {
         "node": "./dist/esm-node/main.mjs",
-        "default": "./dist/es/main.js"
+        "default": "./dist/es/main.js",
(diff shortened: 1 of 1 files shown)


m0ksem (approved):
Thanks!

PR #3609, opened 2023-07-12
feat(library-builder): builder

m0ksem:


m0ksem:
I decided I don't need this ASAP, so we can put it on hold.

m0ksem:
This PR actually reduces bundle size fixing lodash issues, lol. Need to be finished.

PR #3631, opened 2023-07-24
fix: weird arrow showing for external redirect links

### LINKED ISSUE #3628: Weird arrow for external redirect
**Vuestic-ui version:** 1.7.2

image

SiddharthaSarma:
closes #3628 

## Types of changes

### DIFF
--- a/packages/docs/modules/page-config/blocks/shared/markdown/MarkdownView.vue
+++ b/packages/docs/modules/page-config/blocks/shared/markdown/MarkdownView.vue
@@ -109,7 +109,7 @@ const codeBorder = computed(() => {
     position: relative;
 
     &::after {
-      content: "\279A";
+      content: "";
       position: relative;
       opacity: 0.35;
       line-height: 1;


packages/docs/modules/page-config/blocks/shared/markdown/MarkdownView.vue
```diff
@@ -109,7 +109,7 @@ const codeBorder = computed(() => {
     position: relative;
 
     &::after {
-      content: "\279A";
+      content: "";
```
m0ksem, line 112: How about removing whole `::after` element CSS?

m0ksem (approved):

packages/docs/modules/page-config/blocks/shared/markdown/MarkdownView.vue
SiddharthaSarma: sure, let me update the PR

SiddharthaSarma: updated the PR

SiddharthaSarma:
> Maybe we can fix it instead of removing? It seems pretty important for accessibility. Could be some other svg, or maybe smaller arrow (current arrow just looks wrong).
> 
> https://codersblock.com/blog/external-links-new-tabs-and-accessibility/

sure, will take a look

SiddharthaSarma:
I am not able to find the right unicode for the external link and closest one which I can find is U+1F517
Is it ok if I use that one or any svg for external link?

m0ksem:
Hi. I think it is okay to use svg. Or we can use https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid if it is possible.

PR #3632, opened 2023-07-24
fix(dropdown): more demos and fix multiple nodes rendered in proxy component

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3487

SVG, v-if, v-for, v-slot, component and multiple nodes demos

image

I don't have more ideas how dropdown can be used...


asvae (approved):
🤟

Issue #3496, opened 2023-06-14
Cannot make screenshots on any pages apart from landing.


asvae:
**Vuestic-ui version:** 1.6.6

Here you can see 2 screenshots taken by percy. Landing looks ok, but documentation page only shows top part.
<img width="291" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/5394573/45480482-ddd6-462d-9e74-591d9f4f654a">

And here's a screenshot I took via "Capture node screenshot"
<img width="759" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/5394573/148e6243-5424-482d-a07f-a4b5f686b002">
![ui vuestic dev_](https://github.com/epicmaxco/vuestic-ui/assets/5394573/cea465c1-bdff-45a5-a161-b228c4d84942)

As you can see - for docs - both screenshots are cut.

That's problematic because we can't:
* Make a screenshot of full documentation page, which is useful for pull requests.
* Make automated visual testing for docs.

asvae:
Maybe has something to do with #1430

asvae:
Confirming that's fixed in #3633. Works beautifullly!

PR #3633, opened 2023-07-24
feat: layout

m0ksem:
NEW PR with Sidebar and Navbar fixes

Introduces breaking changes in Navbar

closes https://github.com/epicmaxco/vuestic-ui/issues/1430
closes https://github.com/epicmaxco/vuestic-ui/issues/3496

Features:
- [x] New component: VaLayout
- [x] top, bottom, left, right slots
- [x] order
- [x] animations support
- [x] absolute areas
- [x] fixes areas
- [x] mobile friendly
- [x] docs and examples
- [x] window scroll (see https://github.com/epicmaxco/vuestic-admin/pull/927#discussion_r1273842969)

Default:
image

Order:
image

Absolute:
image


packages/docs/page-config/ui-elements/layout/index.ts
```diff
@@ -0,0 +1,67 @@
+export default definePageConfig({
+  blocks: [
+    block.title('Layout'),
+    block.paragraph('Component is used for building App layout. It is based on CSS Grid and provides a simple API for building complex layouts.'),
+    block.paragraph(`
+VaLayout component is used in pair with [VaSidebar](/ui-elements/sidebar)[[target=_blank]], [VaNavbar](/ui-elements/navbar)[[target=_blank]] components.
```
asvae, line 6:
```suggestion
VaLayout component could be used in pair with [VaSidebar](/ui-elements/sidebar)[[target=_blank]], [VaNavbar](/ui-elements/navbar)[[target=_blank]] components.
```

packages/docs/page-config/ui-elements/navbar/examples/Default.vue
```diff
@@ -10,10 +10,11 @@
         Right slot
       </va-navbar-item>
     </template>
-
-    <va-navbar-item class="navbar-item-slot">
-      Default slot (center)
-    </va-navbar-item>
+    <template #center>
```
asvae, line 13: Menu button does nothing:

image

packages/docs/page-config/ui-elements/layout/examples/AllSlots.vue
```diff
@@ -0,0 +1,38 @@
+<script setup lang="ts">
+  const showLeftSidebar = ref(true)
+  const showRightSidebar = ref(true)
+</script>
+
+<template>
```
asvae, line 6: Would be nice to have something that looks all right on mobile:

image

packages/docs/page-config/ui-elements/layout/examples/Order.vue
```diff
@@ -0,0 +1,102 @@
+<script setup lang="ts">
```
asvae, line 1: Also would be nice to have a plausible mobile example.

image

Maybe we can just make sidebars proportional on mobile?

asvae (changes requested):
Found a couple of bugs.

I checked experimental.

m0ksem:
Wondering if we want to improve Navbar so it can have shadow and bordered class.

Also, looks like having VaFooter (the same as Sidebar, but with top shadow and border) will be a huge DX improvement.

asvae:
Moved story issue here - #3734

packages/docs/page-config/ui-elements/layout/examples/Default.vue
```diff
@@ -0,0 +1,77 @@
+<script setup lang="ts">
```
asvae, line 1: Text eloped

image

asvae:
@m0ksem button visual bug is still relevant

I suggest we create a button with right 5rem or so, so that it doesn't overlap with close.
We also probably want to make this button have white half-transparent background and make it stick on scroll.

image

asvae (approved):

PR #3637, opened 2023-07-25
Feat#1417/add new menu component

pogrib0k:
## Types of changes
- [x] Add new Menu component
  - [x] Dropdownable list of click options
  - [x] Icon option (left, right)
  - [x] Groups of options in menu
  - [x] Double wrapping functionality
  - [x] Keyboard navigation
  - [x] MenuItem improvement (more slots like "prepend" etc.)
  - [x] Props for: disabled, position, activator slot, alighment, close delay, close on click outside, close on content click, content-class, disable keybord, render on mounted, max and min sizes, offset-overflow, open on click / on focus / on hover, transition, z-index, value, rounded
  - [x] Activator and default slot
  - [x] Refactor all code


packages/ui/src/components/va-menu/components/VaMenuItem.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <div class="va-menu-item"
+    :class="{
+      'va-menu-item--disabled': disabled
+    }"
+    @click.stop="disabled ? '' : $emit('option-click', $event)"
+  >
+    <VaIcon class="va-menu-item__icon--left" v-if="icon" :name="icon" />
+    <a class="va-menu-item__content" :style="{
```
m0ksem, line 9: Content must be aligned to the left.

packages/ui/src/components/va-menu/VaMenu.vue
```diff
@@ -0,0 +1,53 @@
...
+    <template #anchor>
+      <slot name="anchor">
+        <VaButton>
+          Слава Україні!!!
+        </VaButton>
+      </slot>
+    </template>
+    <template #default>
+      <VaDropdownContent :no-padding="true">
+        <va-menu-list :options="options" @option-click="$emit('option-click', $event)">
```
m0ksem, line 12: We need to cover this component with slots:

- option - replace whole option
- left-icon - replace left icon (for example with Avatar)
- right-icon - replace right icon
- option-text - replace text
- options - replace whole options list v-for, so user can render VaMenuItem manually

```html
<VaMenu>
   <template #options>
       <VaMenuItem>My profile</VaMenuItem>
       <VaMenuItem>Settings</VaMenuItem>
       <VaDivider />
       <VaMenuItem style="color: red">Log out</VaMenuItem>
   </template>
</VaMenu>
```

packages/ui/src/components/va-menu/VaMenu.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+export default {
+  title: 'VaMenu',
+  component: VaMenuList,
+}
+
+export const Default = () => ({
+  components: { VaMenuList },
+  data: () => ({
+    options: [
+      { id: '0', text: 'one', value: 'one' },
```
asvae, line 16: do we need all 3 params here? 

maybe this could be default:

```ts
options: [
  'one',
  'one',
  'one',
]
```

packages/ui/src/components/va-menu/VaMenu.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+      { id: '1', text: 'two', value: 'two' },
+      { id: '2', text: 'three', value: 'three' },
+    ],
+    value: false,
```
asvae, line 20: I think the default for menu should include selected action. As that would be its main usecase.

packages/ui/.storybook/assets/fonts.scss
```diff
@@ -9,7 +9,4 @@ body {
   font-size: 1rem;
   font-weight: 300;
   line-height: 1.5;
-  color: #34495e;
-  text-align: left;
-  background-color: #eeeeee;
```
asvae, line 14: Maybe as a separate pr?

We'll get a lot of [changes](https://www.chromatic.com/build?appId=64d608ab34b67e621ab50cfd&number=432)

packages/ui/src/components/va-menu-list/VaMenuList.stories.ts
```diff
@@ -0,0 +1,162 @@
...
+      { id: '4', text: 'five', value: 'five', icon: 'search', rightIcon: '', group: 'A2' },
+    ],
+    value: false,
+  }),
+  template: `
+  <VaMenuList :options="options">
+  </VaMenuList>
+  `,
+})
+addText(
```
asvae, line 61: I added my notes to story descriptions. We probably need to fix these before we merge.

asvae (changes requested):
Some small changes would be welcome

packages/ui/src/components/va-menu-list/VaMenuList.vue
```diff
@@ -0,0 +1,150 @@
+<template>
+  <table class="va-menu-list" ref="container" v-bind="makeMenuContainerAttributes()">
```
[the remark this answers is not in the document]
m0ksem: Table make sense when you work with columns which need to be aligned automatically. So, if you put here avatar, all other icons will be aligned correctly.

packages/ui/src/components/va-menu-list/VaMenuList.stories.ts
```diff
@@ -0,0 +1,162 @@
...
+  data: () => ({
+    options: ['Option 1', 'Option 2', 'Option 3'],
+  }),
+  template: `
+    <VaMenuList color="warning" :options="options" />
+  `,
+})
+addText(
+  Groups,
+  `I don't think we have any other component where 'color' means 'hover-color', let's make it explicit.` ,
```
m0ksem, line 97: Color means component accent color in general. Color doesn't mean hover-color, otherwise hover color would be exactly primary (hover color have less opacity). In case you need to change hover-color I'll add a css variable for that.

asvae: image

Just to double check we're on the same page - this is a story for color. And this color is set on hover.

So color doesn't create any coloration by default - it requires hover. Which I believe is never how color works for other components. I would expect font color to change or background or something, but not hover color.

asvae: I checked a couple of other frameworks - and they don't have color for menu at all. So how about we remove it and only keep variable?

packages/docs/page-config/ui-elements/menu/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    block.title("Menu"),
+    block.paragraph("The `VaMenu` component is used to display structured information under the menu button."),
+
+    block.link("VaMenuList", "/ui-elements/menu-list", { preText: "See " }),
+
+    block.subtitle("Examples"),
+
+    block.example("Default", {
+      title: "Basic usage",
+      description: "Required to have a anchor. Anchor can be anything, component will automatically listen for clicks, keyboard navigation and open/close menu. To show menu items you can use `options` prop or `VaMenuItem` component.",
```
asvae, line 12:
```suggestion
      description: "Required to have a anchor. Component will automatically listen for clicks, keyboard navigation and open/close menu. To show menu items you can use `options` prop or `VaMenuItem` component.",
```

packages/docs/page-config/ui-elements/menu/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    }),
+    block.example("MenuItem"),
```
asvae, line 14: Let's add description for this component. Right now it's not clear why this example exists.

"You can also use slot and `VaMenuItem` component to achieve better flexibility:"

asvae (changes requested):
I believe we can merge after fixes. Looks good!

packages/docs/page-config/ui-elements/menu/index.ts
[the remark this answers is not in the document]
asvae: image

PR #3641, opened 2023-07-27
Added Linear stepper and custom step actions

### LINKED ISSUE #3635: Added linear (validation dependent) option for VaStepper
A user is able to skip around steps which can result in data issues.

Below is an example of how I accomplished this, but I'm sure it could be improved when dealing with the stepper component its self. 

The ``fromForm`` ref is necessary because the watch is called to intercept the step change and modify the step value based on the validity of the form. This causes the watch to get called indefinitely. The only time that the validation logic is called is if it was triggered by clicking either the next or previous button. Again, this can probably be simplified from the inside the actual stepper component. 

Code can be found in Stepper.vue

The stepper should allow for linear navigation. This requires that the step (form) is valid before being able to proceed to the next step. This is crucial if you are using the stepper for a wizard as users can currently skip around which could cause important steps to be missed.

Bond-Addict:
## Description

closes #3635
This change will introduce two useful features:

- Linear Mode- This forces form validation to be valid before navigation will occur.
- Custom step actions- This add a ``save`` property on the Step interface. This allows the user to specify save actions for each step. This goes hand in hand with linear because it makes it where the stepper can be treated like a true wizard ensuring a previous step has completed successfully.  Example: You create a user on step 1 and then you assign multiple roles on step 2. If the user didn't exist when you get to step 2 you wouldn't be able to add the roles. You won't navigate to the desired step unless the save function succeeds. 

- Required validator- Adds a required function that accepts a custom error message but provides a default one.

## Types of changes


Bond-Addict:
The whitespace changes where made automatically by prettier

m0ksem:
> The whitespace changes where made automatically by prettier

I understand it. But better do not use any formatter, so it is easier to review the code...

Maybe we should have a standardized Prettier config, actually.

packages/ui/package.json
```diff
@@ -46,7 +46,8 @@
     "@types/lodash": "^4.14.161",
     "cleave.js": "^1.6.0",
     "colortranslator": "^1.9.2",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "vue-router": "^4.2.4"
```
m0ksem, line 50: Why would we need vue-router?

packages/ui/src/components/va-stepper/VaStepper.vue
```diff
@@ -225,6 +179,23 @@ export default defineComponent({
...
 
+    const save = (targetIndex: number) => {
+      const save = props.steps[modelValue.value].save;
+      let errored = false;
+      if (!validate() && props.linear) {
+        return;
+      }
+      if (save && !errored) {
+        try {
+          save();
```
m0ksem, line 190: I would better make a validate method which returns true/false instead of throwing error. The same way we have for inputs.

packages/ui/src/components/va-stepper/VaStepper.vue
```diff
@@ -1,100 +1,51 @@
...
-  <div
-    class="va-stepper"
-    :class="{ 'va-stepper--vertical': $props.vertical }"
-    v-bind="ariaAttributesComputed"
-  >
-    <ol
-      class="va-stepper__navigation"
-      ref="stepperNavigation"
-      :class="{ 'va-stepper__navigation--vertical': $props.vertical }"
+  <VaForm ref="form">
```
m0ksem, line 2: It feels like Form must be used outside of stepper. Better register each step as form field and add validation method to it.

packages/ui/src/components/va-stepper/VaStepperControls.vue
```diff
@@ -1,23 +1,12 @@
 <template>
   <div class="va-stepper__default-controls">
-    <va-button
-      preset="primary"
-      :disabled="$props.modelValue <= 0"
-      @click="$props.stepControls.prevStep()"
-    >
+    <va-button preset="primary" :disabled="$props.modelValue <= 0" @click="$props.stepControls.prevStep()">
```
m0ksem, line 3: This file seems to be changed with prettier only, so we can revert this changes?

m0ksem (changes requested):
The package-lock file must be removed, because we use yarn.

packages/ui/package.json
Bond-Addict: This was only added because book-main.ts requires it and it was missing causing the build to fail.

packages/ui/src/components/va-stepper/VaStepper.vue
Bond-Addict: When you say must, what do you mean? Functionally? Would doing it this way cause performance issues?

Can you explain more what you're suggesting? Do you mean that each step should be a va-form? How would you recommend doing that?

packages/ui/package.json
m0ksem: You're right. Vue book requires vue-router to work, but not `vuestic-ui` package. We should move `vue-router` from `/docs` to `/ui`, and it must be dev-dependency because it used only in our demos.

packages/ui/src/components/va-stepper/VaStepper.vue
m0ksem: The interface must be like

```html
<VaForm ref="form">
  <VaStepper>
      ...
  </VaStepper>
</VaForm>
```
It is more tree-shaking efficient, because not all steps requires a form inside. Also, with VaForm used as parent for VaStepper allows user to access form with `useForm`.

Usage must be like:
```vue
<template>
  <VaForm ref="stepperForm">
    <VaStepper v-model="step" :steps="steps" :next-disabled="!isValid" @before-next="(next) => next(validate)">
      <template #step-content-0>
        <va-input
          v-model="model.a"
          :rules="[(v) => v.length > 0 || 'Required']"
          label="A"
        >
        </va-input>
      </template>
      <template #step-content-1>
        <va-input
          v-model="model.b"
          :rules="[(v) => v === '13' || 'Value must be 13']"
          label="B"
        ></va-input>
      </template>
      <template #step-content-2>
        <va-input v-model="model.c" label="C"></va-input>
      </template>
    </VaStepper>
  </VaForm>
</template>

<script setup lang="ts">
import { ref } from "vue";
import { useForm } from "vuestic-ui";

const { isValid, validate } = useForm("stepperForm");

const step = ref();
const model = ref({ a: "", b: "", c: "" });

const steps = [{ label: "One" }, { label: "Two" }, { label: "Three" }];
</script>
```

It does the same thing you have with `$props.linear` and also disables nextButton visually.
The only thing we need to add is event `before-next` where we can validate form (`@before-next="(next) => next(validate)` like vue-router guard). But it is actually not required, because you can pass custom buttons.

Here is a working example: https://codesandbox.io/p/sandbox/vibrant-orla-sqrjc4?file=/src/App.vue:3,46-3,64 (Idk why codesandbox draws controls two times, btw)

Do you feel like you still need this behavior to be done with prop or we can move validation process out of component?

Bond-Addict: One thing to note, in your example navigation is disabled. It would be nice to still retain that functionality

Bond-Addict: @m0ksem can you check out my last wip change? I'm trying to pass form as ref. It does show as a ref, but validate is always undefined. I've tried passing the actual ref and the string name for the form.

m0ksem: I'm not sure how to implement validation with navigation enabled...

If you have step1 valid, you must be able to go to step2.
In case you're back to step1 and made it invalid, you mustn't be able to go to step2 again.

Also, VaForm validates only mounted components, so we're not able to check if step1 valid after moved to step2.


---
I think this can be fixed if nextDisabled will disable step2 if step1 is invalid, even if you typed something in form. Looks like correct behavior for me.

Bond-Addict: I most ideal (imo) would be to have a prop for the form on the stepper. If the form is passed, then it knows validation should occur. This would negate the need for linear property. But Vue automatically unwrapping the ref, and then useForm expecting a string or Ref<typeof VaForm> is causing an issue. I'm searching S.O and open issues, but I'm not seeing a resolution. I've tried nesting the formRef inside an object (so it isnt unwrapped) but that did not work. Other suggestions were to use makeRaw but that also does not work. 

I think this type of approach would provide a simpler and thus more enjoyable development experience if possible.

Bond-Addict: I think this would be accomplished by adding a prop of completed on each step. When you successfully validate and then update the current step index, then you also set that step as completed. Then you cannot proceed to a step unless the previous step is completed and the current step is valid. Thoughts? I think this is basically how angular material works. They have a linear mode but you can also use the upper navigation.

Bond-Addict: > I'm not sure how to implement validation with navigation enabled...
> 
> If you have step1 valid, you must be able to go to step2. In case you're back to step1 and made it invalid, you mustn't be able to go to step2 again.
> 
> Also, VaForm validates only mounted components, so we're not able to check if step1 valid after moved to step2.
> 
> I think this can be fixed if nextDisabled will disable step2 if step1 is invalid, even if you typed something in form. Looks like correct behavior for me.

I think this would be accomplished by adding a prop of completed on each step. When you successfully validate and then update the current step index, then you also set that step as completed. Then you cannot proceed to a step unless the previous step is completed and the current step is valid. Thoughts? I think this is basically how angular material works. They have a linear mode but you can also use the upper navigation.

m0ksem: It looks like just didn't register VaForm in demo component. 
Also, notice that you mustn't use `useForm` in methods, only during setup fn execution. You can directly access validate method in VaForm ref. 
You also don't need to use `FormRef`, even thought vue unwrapps ref in props `props.form` is still reactive, it is because props are`reactive`. 
```ts
const form = toRef(props, 'form')
```
`form` will not lose reactivity this way.

In template, all refs are unwrapped, so in case `<VaStepper :form="myCustomForm" />` will be compiled like
```ts
h(VaStepper, {
   form: myCustomForm.value,
})
```
(actually it will use ctx.myCustomForm which is reactive too, so code above is just analogy)

Bond-Addict: Okay cool. That does work exactly as intended! Sorry, I'm still getting a total hang of Vue's reactivity.

packages/ui/src/components/va-stepper/VaStepper.vue
Bond-Addict: Is my use of adding the isValid function acceptable?

Bond-Addict:
The only conflict I have is the yarn lock. Should I do the same as the package.lock and delete the yarn.lock?

m0ksem:
> The only conflict I have is the yarn lock. Should I do the same as the package.lock and delete the yarn.lock?

I can handle it before merge.

packages/ui/src/components/va-stepper/VaStepper.vue
Bond-Addict: @m0ksem  can this be resolved?

packages/ui/src/components/va-stepper/VaStepper.vue
Bond-Addict: @m0ksem after playing around a bit with it, I did opt to not have the stepper trigger form validation automatically. You of course can still wrap the stepper with the form and then complete the step or set the step as errored if validation fails.

Bond-Addict:
@m0ksem I think this is ready for final review

Bond-Addict:
@m0ksem Are there any more adjustments needed?

packages/ui/src/components/va-stepper/VaStepper.vue
```diff
@@ -307,5 +338,17 @@ export default defineComponent({
     display: flex;
     gap: var(--va-stepper-controls-gap);
   }
+
+  .invalid-navigation.va-stepper__step-button--active{
+    color: var(--va-danger);
+    --30407a18-stepperColor: var(--va-danger) !important;
```
m0ksem, line 344: You  mustn't do that... 
We use v-bind in vue which adds a uniq hash for css variable. It means this variable will be different every build. You need to pass `danger` to Button via props.

packages/ui/src/components/va-stepper/types.ts
```diff
@@ -2,6 +2,9 @@ export type Step = {
   label: string
   icon?: string
   disabled?: boolean
+  save?: Function
+  hasError?: boolean
+  completed?: boolean
```
m0ksem, line 7:
```suggestion
  /** Will be set to true if user completed step without validation error */
  completed?: boolean
```

packages/ui/src/components/va-stepper/types.ts
```diff
@@ -2,6 +2,9 @@ export type Step = {
   label: string
   icon?: string
   disabled?: boolean
+  save?: Function
+  hasError?: boolean
```
m0ksem, line 6:
```suggestion
  /** Will be set to true if user completed step with validation error */
  hasError?: boolean
```

packages/ui/src/components/va-stepper/types.ts
```diff
@@ -2,6 +2,9 @@ export type Step = {
   label: string
   icon?: string
   disabled?: boolean
+  save?: Function
```
m0ksem, line 5:
```suggestion
  save?: (step: Omit<Step, 'save'>) => boolean
```

Let's specify function type. Let's say save first argument is going to be a step, so we can do:

```
save(step) {
  if (validate()) {
    step.complete = true
  } else {
    step.hasError = true
  }
}
```

I also omited save method, so user will not be able to call save in save and in save. (There will be TS error, actually he can)

m0ksem (changes requested):
Hi! Sorry for later response!

2023-08-10 at 7 27 34 - Sapphire Sparrow
Demo show error class only once.

I'd say calling component method in most cases in bad practice. Would be nice to be able to set step as complete or as with error directly in save function. And let's call it not save, but "beforeNext". I'd assume save is called when stepper is complete (all steps are completed).

Also, I think we don't need `complete` property. We can change step type to
```ts
type Step {
   [key: string]: any,
   label:string,
   // other properties
}
```

This way in save fn you can add to step anything you want and need.
The UI framework will only provide a way to visually display error:

```ts
save(step) {
   step.error = validate()
}
```

Let's also discuss what step.error does:
- It must show step with error class you made.
- It must prevent moving out from this step if $props.linear is true (so navigation is disabled if there is error in current step).

Otherwise, I like it.

m0ksem:
I've updated yark.lcok and resolved merge conflict. 

Notice that instead of `yarn serve` you need to use `yarn serve:storybook` now.

packages/ui/src/components/va-stepper/types.ts
Bond-Addict: I did remove the form validation from the stepper its self. Is there an additional change I need to make?

packages/ui/src/components/va-stepper/VaStepper.vue
Bond-Addict: Can you check my last commit? I would think this would work, but the button color doesnt seem to react to the new button color. I figured that I could use a computed prop, but from what I'm seeing I cant pass a parameter to a computed prop.

Bond-Addict:
> Hi! Sorry for later response!
> 
> 2023-08-10 at 7 27 34 - Sapphire Sparrow Demo show error class only once.
> 
> I'd say calling component method in most cases in bad practice. Would be nice to be able to set step as complete or as with error directly in save function. And let's call it not save, but "beforeNext". I'd assume save is called when stepper is complete (all steps are completed).
> 
> Also, I think we don't need `complete` property. We can change step type to
> 
> ```ts
> type Step {
>    [key: string]: any,
>    label:string,
>    // other properties
> }
> ```
> 
> This way in save fn you can add to step anything you want and need. The UI framework will only provide a way to visually display error:
> 
> ```ts
> save(step) {
>    step.error = validate()
> }
> ```
> 
> Let's also discuss what step.error does:
> 
> * It must show step with error class you made.
> * It must prevent moving out from this step if $props.linear is true (so navigation is disabled if there is error in current step).
> 
> Otherwise, I like it.

Btw, the demo class showing as an error once was actually by design. I should have mentioned that with the example. There is a retry count with that demo. After a retry is made, hasError is then set false on the next try.

m0ksem:
@Bond-Addict, I made demos less confusing and also added few extra features. Please, review if I did something wrong. Otherwise, it looks good for me. The last step is to update docs demo if you're ok with changes I made.

Bond-Addict:
> @Bond-Addict, I made demos less confusing and also added few extra features. Please, review if I did something wrong. Otherwise, it looks good for me. The last step is to update docs demo if you're ok with changes I made.

I looks great dude. Thanks for the assist! Just for my understanding, what was I doing wrong with the button color part? Was there some vue fundamental I was messing up? I'll get the docs updated as soon as possible.

m0ksem:
> > @Bond-Addict, I made demos less confusing and also added few extra features. Please, review if I did something wrong. Otherwise, it looks good for me. The last step is to update docs demo if you're ok with changes I made.
> 
> I looks great dude. Thanks for the assist! Just for my understanding, what was I doing wrong with the button color part? Was there some vue fundamental I was messing up? I'll get the docs updated as soon as possible.

There was a bug on line https://github.com/epicmaxco/vuestic-ui/pull/3641/files#diff-330631d4f7a4d81d8d3d14e31f746d0e1c148d4e5bb9c063913166c6a7b5addeL50. Because it wasn't computed, color wasn't updating if you change it later... So it always was primary.

Bond-Addict:
> > > @Bond-Addict, I made demos less confusing and also added few extra features. Please, review if I did something wrong. Otherwise, it looks good for me. The last step is to update docs demo if you're ok with changes I made.
> > 
> > I looks great dude. Thanks for the assist! Just for my understanding, what was I doing wrong with the button color part? Was there some vue fundamental I was messing up? I'll get the docs updated as soon as possible.
> 
> There was a bug on line https://github.com/epicmaxco/vuestic-ui/pull/3641/files#diff-330631d4f7a4d81d8d3d14e31f746d0e1c148d4e5bb9c063913166c6a7b5addeL50. Because it wasn't computed, color wasn't updating if you change it later... So it always was primary.

Ah okay. I had tried to use a computed earlier but there must have been something messed up. I'm glad to know I was on the right track though. I'll try and get the docs update Monday morning. Have a good weekend!

Bond-Addict:
@m0ksem Can you explain the correct steps to get the docs to run? I've gotten it to generate one time, but I have no idea how I did it. Once I can get them to run I'll get the docs updated. I've been looking at the contribution docs, but cant seem to figure it out. I might be doing something wrong, but I cant get ``yarn serve:docs`` to work in either docs or ui.

m0ksem:
> @m0ksem Can you explain the correct steps to get the docs to run? I've gotten it to generate one time, but I have no idea how I did it. Once I can get them to run I'll get the docs updated. I've been looking at the contribution docs, but cant seem to figure it out. I might be doing something wrong, but I cant get `yarn serve:docs` to work in either docs or ui.

Maybe you need to reinstall deps. Run `yarn && yarn serve:docs` in root of the project and you should be good.

Bond-Addict:
> > I this error normal?
> > `Failed to resolve import "fsevents" from "node_modules\.cache\vite\client\deps\chunk-5CLDZUOX.js?v=3a618fd0". Does the file exist?`
> 
> Nope. Try to clean node_modules/.cache or .nuxt folder. Likely something is cached. Or try yarn --force.

I'll try clearing cache and nuxt folder but I have tried a couple times to run yarn --force

Bond-Addict:
> Author

@m0ksem I'm unable to get the docs updated due to this issue. In light of this, would you be able to update the doc? My project is dependent on this feature so its at a stand standstill.

m0ksem:
> > Author
> 
> @m0ksem I'm unable to get the docs updated due to this issue. In light of this, would you be able to update the doc? My project is dependent on this feature so its at a stand standstill.

Yeah, sure. I wanted to handle the rest of pr before next release and I hope @asvae gonna have some time to review it.

Bond-Addict:
> > > Author
> > 
> > 
> > @m0ksem I'm unable to get the docs updated due to this issue. In light of this, would you be able to update the doc? My project is dependent on this feature so its at a stand standstill.
> 
> Yeah, sure. I wanted to handle the rest of pr before next release and I hope @asvae gonna have some time to review it.

Awesome, thank you so much. I appreciate the continued support. I'll try and ensure the next change goes a bit smoother.

packages/ui/src/components/va-stepper/VaStepper.demo.vue
```diff
@@ -94,4 +167,11 @@ const stepsWithCustomIcons = [
   { label: 'Four', icon: 'grade' },
   { label: 'Five', icon: 'list' },
 ]
+
+const stepsWithNextAction = ref([
+  { label: 'One' },
+  { label: 'Two', beforeSave: (step) => { step.hasError = true } },
```
asvae, line 173: I wouldn't cal it save. No saving is happening.

Maybe something like `onOut`?

asvae (approved):
Thanks!

This looks good enough for now, enabling form functionality for stepper. Later we might want to rework the component - some things are not done optimally.

Bond-Addict:
@m0ksem  How do I resolve your requested change? The pr is currently failing the "Changes requested" precheck.

asvae:
@Bond-Addict hey, I actually coded it quickly myself (hope you don't mind the intrusion 🙏).

I can merge it no problems. :)

m0ksem (approved):

Bond-Addict:
> @Bond-Addict hey, I actually coded it quickly myself (hope you don't mind the intrusion 🙏).
> 
> I can merge it no problems. :)

I don't mind at all. I appreciate the assist from yall to get this feature in. I look forward to seeing the future optimizations as I'm still learning Vue and want to learn the best practices 😃

m0ksem:
> @m0ksem How do I resolve your requested change? The pr is currently failing the "Changes requested" precheck.

Hi, we good. I'm not sure if you saw that I pinged you in Discord, so you could get this feature released in an experimental build. 

We're currently in a process to move dev discussions to discord, where we can communicate faster with contributors, help them and ship code faster. If you were interested in real-time communication instead of GitHub comments, you're welcome!

* We about to release it, but it actually was in version `0.0.0-experimental-5d49247e6-20230821` already.

Bond-Addict:
> > @m0ksem How do I resolve your requested change? The pr is currently failing the "Changes requested" precheck.
> 
> Hi, we good. I'm not sure if you saw that I pinged you in Discord, so you could get this feature released in an experimental build.
> 
> We're currently in a process to move dev discussions to discord, where we can communicate faster with contributors, help them and ship code faster. If you were interested in real-time communication instead of GitHub comments, you're welcome!
> 
> * We about to release it, but it actually was in version `0.0.0-experimental-5d49247e6-20230821` already.

Ah okay I must have missed that message. Thanks for all the assistance getting this out. You rock! 😄

Issue #3625, opened 2023-07-20
Badge docs improvements


asvae:
We did badge improvements (#3246), but there are a couple of things not covered by issue and PRs:
- [ ] `offset` prop is extremely poorly documented, we need to document all usage cases (single number, array, strings, rems/percents etc)
- [ ] It's not possible to set different border color as it's getting overriden by `:styles`. We either want to dispose of borders or most probably make them operational.
- [ ] Default example makes no sense for docs (it makes sense for demos though). I suggest removing it for now, and later on - we can have either rich example or playground instead.
- [ ] Right now in examples to make badge work with icon and avatar components - we use offset. That's not great - these components should work properly with each other out of the box. It could be achieved by either some styling tweak when particular component is detected or by changing default behaviour of badge.
- [ ] `placement` prop description (in table) should have full information about the prop usage. Example could have only high level description.
- [ ] `preset` prop description should have link to preset usage page.

PR #3647, opened 2023-07-31
feat: VaBadge move to storybook

asvae:
Moved va-badge demos fully to storybook. Now they look something like this:

image

I moved demos improvements to #3625. This PR mostly focuses on architecture and story structure.

This storybook structure is not final design, I would probably need 7-10 more iterations at least.


asvae:
Was simple enough, so I decided to merge on my own.

PR #3654, opened 2023-08-03
docs: fix prop types

### LINKED ISSUE #3624: We lost types somewhere
I checked back to 1.7.0 and there are no types. So they disappeared earlier it seems.

We have to bring them back.

image

m0ksem:
closes #3624


asvae:
Confirming fix.

image

PR #3663, opened 2023-08-07
fix#3652/docs design 

### LINKED ISSUE #3652: Docs Design Issues
Sidebar > image

Roman4437:
closes #3652 

Moved both components into a div container and applied mt-2 space-x-2 to it


asvae:
Great!

In the future please add screenshot before/after. So that it doesn't require local setup to test.

PR #3667, opened 2023-08-08
fix#3662/VaDropdown overflow bug

### LINKED ISSUE #3662: VaDropdown: Dropdown ignores page structure
**Vuestic-ui version:** 1.7.1

My guess would be is that we changed default behaviour for location. Dropdown is in body, so overflow: hidden won't make a difference.

Notice change was introduced in 1.7.1.

We have a couple of broken examples (but please check other places and other examples):

image

image

Roman4437:
closes #3662 

predefined `teleport: 'body'` makes each `<va-dropdown/>` to be atached to `<body></body>` regardless of whether teleport was defined or not, by removing this line, we will fix all the strange behavior with overflow.

### DIFF
--- a/packages/docs/config/vuestic-config.ts
+++ b/packages/docs/config/vuestic-config.ts
@@ -19,7 +19,6 @@ export const VuesticConfig = defineVuesticConfig({
     },
     VaDropdown: {
       target: scrollWrapperSelector,
-      teleport: 'body',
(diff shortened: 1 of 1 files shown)


m0ksem (approved):

m0ksem:
Thanks!

PR #3677, opened 2023-08-09
accordion stories refactor

Roman4437:
Is it good?

### DIFF
--- a/packages/ui/src/components/va-accordion/VaAccordion.stories.ts
+++ b/packages/ui/src/components/va-accordion/VaAccordion.stories.ts
@@ -1,12 +1,167 @@
-import { defineComponent } from 'vue'
-import VaAccordion from './VaAccordion.demo.vue'
+import { VaAccordion } from './'
+import { VaCollapse } from '../va-collapse'
+
+interface Args {
+  modelValue: boolean[]
+  multiple: boolean
+  inset: boolean
+  popout: boolean
+}
+
+const VaAccordionSetup = {
+  value: [false, false, false],
+  valueArray: [false, true, false],
+  collapses: [
+    { title: 'First collapse', content: 'first collapse content' },
+    { title: 'Second collapse', content: 'second collapse content' },
+    { title: 'Third collapse', content: 'third collapse content' },
+  ],
+}
 
 export default {
   title: 'VaAccordion',
   component: VaAccordion,
+  tags: ['autodocs'],
 }
 
-export const Default = defineComponent({
-  components: { VaAccordion },
-  template: '<VaAccordion/>',
+export const Default = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const DefaultWithCustomHeader = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="value" style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+      >
+        <template #header="{ value }">
+          {{ value ? 'opened' : 'closed' }}
+        </template>
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const Multiple = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="valueArray" multiple style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const Inset = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="value" inset style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const InsetMultiple = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="valueArray" inset multiple style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const Popout = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="value" popout style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const PopoutMultiple = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion v-model="valueArray" popout multiple style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
 })


packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+import { VaCollapse } from '../va-collapse'
+
+interface Args {
+  modelValue: boolean[]
+  multiple: boolean
+  inset: boolean
+  popout: boolean
+}
+
+const VaAccordionSetup = {
```
asvae, line 11:
```suggestion
const accordionSetup = {
```

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
-import { defineComponent } from 'vue'
-import VaAccordion from './VaAccordion.demo.vue'
+import { VaAccordion } from './'
+import { VaCollapse } from '../va-collapse'
+
+interface Args {
```
asvae, line 4: Let's not use args for now. None of the demos require them.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+  multiple: boolean
+  inset: boolean
+  popout: boolean
+}
+
+const VaAccordionSetup = {
+  value: [false, false, false],
+  valueArray: [false, true, false],
+  collapses: [
+    { title: 'First collapse', content: 'first collapse content' },
```
asvae, line 15: Let's use the same text for all titles and contents.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+  tags: ['autodocs'],
 }
 
-export const Default = defineComponent({
-  components: { VaAccordion },
-  template: '<VaAccordion/>',
+export const Default = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
```
asvae, line 30: Each story should be independent and have relevant data for that particular story.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
-import { defineComponent } from 'vue'
-import VaAccordion from './VaAccordion.demo.vue'
+import { VaAccordion } from './'
```
asvae, line 1: Let's delete `.demo.vue` file. It's not needed if this PR is merged.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion style="width: 400px;">
+      <va-collapse
+        v-for="(collapse, index) in collapses"
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
```
asvae, line 39: Do we need div here?

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
-export const Default = defineComponent({
-  components: { VaAccordion },
-  template: '<VaAccordion/>',
+export const Default = (args: Args) => ({
+  components: { VaAccordion, VaCollapse },
+  setup () {
+    return VaAccordionSetup
+  },
+  template: `
+    <va-accordion style="width: 400px;">
```
asvae, line 33: Do we need width here?

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+        :key="index"
+        :header="collapse.title"
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
 })
```
asvae, line 167: We need additional story for `stateful` prop.

All props should be covered by stories. You can get a list of props from docs: https://ui.vuestic.dev/ui-elements/accordion#props.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const Inset = (args: Args) => ({
```
asvae, line 89: We can leave only one story for inset. That looks like this: 

[open]
[open]
[closed]
[closed]
[open]

That way we'll handle all visual interactions between item open-closed, closed-open, closed-closed, open-open.

Same for Popout.

We can apply same open/closed config to multiple story.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+export const DefaultWithCustomHeader = (args: Args) => ({
```
asvae, line 47: We can probably remove this demo as it has nothing to do with accordion. (it's collapse specific)

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,167 @@
...
 export default {
   title: 'VaAccordion',
   component: VaAccordion,
+  tags: ['autodocs'],
 }
 
-export const Default = defineComponent({
-  components: { VaAccordion },
-  template: '<VaAccordion/>',
+export const Default = (args: Args) => ({
```
asvae, line 27: Would be very nice to add interaction tests for default, multiple and stateful stories.

asvae (changes requested):
Nice start!

Left some feedback.

m0ksem:
@asvae, FYI, I'm working on visual changes for VaCollapse and VaAccordion components. Double check if it is a good idea to refactor stories right now. Some demos will be removed.

Roman4437:
image
image
image
image

Roman4437:
stateful propriety seemingly doesn't work properly

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+      modelValue: [false, true, false],
+      collapses: [
+        { title: 'Collapse', content: 'Content' },
+        { title: 'Collapse', content: 'Content' },
+        { title: 'Collapse', content: 'Content' },
+      ],
+    }
+  },
+  template: `
+    <va-accordion v-model="modelValue" multiple>
```
asvae, line 105: We can actually shortcut template even more with following syntax:

```html
<va-accordion>
      <va-collapse
        v-for="i in 3"
        :key="i"
        header="Collapse"
      >
        Content
      </va-collapse>
    </va-accordion>
```

That way you don't need data even.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+Stateful.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const collapses = await canvas.getAllByRole('button', { name: 'Collapse' }) as HTMLElement[]
```
asvae, line 83: await not needed here

```suggestion
  const collapses = canvas.getAllByRole('button', { name: 'Collapse' }) as HTMLElement[]
```

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+    </va-accordion>
+  `,
 })
+
+Multiple.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const collapses = await canvas.getAllByRole('button', { name: 'Collapse' }) as HTMLElement[]
+  await DispatchEventWithDelay(collapses[1], userEvent.click, 500)
+
+  for (const collapse of collapses) {
```
asvae, line 124: We need asserts here. So that it's fully automated test and not just visual demo.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+    await DispatchEventWithDelay(collapse, userEvent.click, 500)
+  }
+
```
asvae, line 127: I suggest to add comments on what each line does, similarly to example that I sent you in discord.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
 })
+
+Multiple.play = async ({ canvasElement }) => {
```
asvae, line 119: For collapse I would image we'd need interaction test for default and multiple collapses:

For default collapse:
* When I click on collapse - it opens.
* When I click on another collapse - that collapse opens and previous collapse closes.
* When I click on opened collapse - it closes.

For multiple collapse:
* When I click on collapse - it opens.
* When I click on another collapse - that collapse opens and previous collapse stays open.
* When I click on opened collapse - it closes.

Interactions 1 and 3 could be reused. 2nd is unique.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+      >
+        <div>
+          {{ collapse.content }}
+        </div>
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+
+Popout.play = async ({ canvasElement }) => {
```
asvae, line 216: For popup/inset I'm struggling to see any value from interactions demos. Do you see some?

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -1,12 +1,237 @@
...
+    return {
+      collapses: [
+        { title: 'Collapse', content: 'Content' },
+        { title: 'Collapse', content: 'Content' },
+        { title: 'Collapse', content: 'Content' },
+      ],
+    }
+  },
+  template: `
+    <va-accordion stateful>
```
asvae, line 67: For stateful demos I would recommend something like this

```html
<p> true </p>
<va-accordion stateful>
... collapses

<p> false </p>
<va-accordion>
... collapses
```

Similar to [this](https://ui-storybook-develop.vuestic.dev/?path=/story/vabadge--multi-line&globals=backgrounds.grid:!false) example.

In both cases story should not use v-model.

Interaction should be:
* for stateful true accordion: on click - should open;
* for stateful false accordion: on click - nothing should happen;

asvae (changes requested):
Some more things. Very close to perfect.

asvae:
> stateful property seemingly doesn't work properly

You can try to fix this in context of this PR, or you can create a failing interaction test - then create a separate issue.

Roman4437:
> 

The second one

Roman4437:
> > stateful property seemingly doesn't work properly
> 
> You can try to fix this in context of this PR, or you can create a failing interaction test - then create a separate issue.

The second one

Issue #3676, opened 2023-08-09
Stories rework


asvae:
Right now stories are in bad state. A lot of them either have too much of stuff in them or not enough. Treated somewhat as a temporary code. This hampers storybook usage and leads to regressions.

After rework we should expect development speed to increase and regression rate to decrease.

We'll also be able to publicly share storybook without feeling shame 😞.

We also don't have solid strategy on how to shape stories (#3673), which is a requirement to consider stories "polished"

## How to progress:

* You can create single pr mentioning this issue, or pr for each component you plan to rework.
* Update this issue only after your changes are merged to `develop`.

## Legend:

* 🟥- stories in `.demo.vue` file;
* 🟨- stories in `.stories.ts` file, but not polished;
* 🟩- stories in `.stories.ts` file and polished.

Polished means:
* tests for all applicable stories
* no [stale] stories
* fits the guide (#3673)

### Components

* 🟥 va-accordion
* 🟥 va-affix
* 🟥 va-alert
* 🟥 va-app-bar
* 🟥 va-aspect-ratio
* 🟥 va-avatar
* 🟥 va-avatar-group
* 🟥 va-backtop
* 🟨 va-badge
* 🟥 va-breadcrumbs
* 🟥 va-button
* 🟥 va-button-dropdown
* 🟥 va-button-group
* 🟥 va-button-toggle
* 🟥 va-card
* 🟥 va-carousel
* 🟥 va-checkbox
* 🟥 va-chip
* 🟥 va-collapse
* 🟥 va-color-indicator
* 🟥 va-config
* 🟥 va-content
* 🟥 va-counter
* 🟥 va-data-table
* 🟥 va-date-input
* 🟥 va-date-picker
* 🟥 va-divider
* 🟥 va-dropdown
* 🟥 va-file-upload
* 🟥 va-form
* 🟥 va-hover
* 🟥 va-icon
* 🟥 va-image
* 🟥 va-infinite-scroll
* 🟥 va-inner-loading
* 🟥 va-input
* 🟥 va-input/components/VaMessageList
* 🟥 va-list
* 🟥 va-modal
* 🟥 va-navbar
* 🟥 va-option-list
* 🟥 va-pagination
* 🟥 va-parallax
* 🟥 va-popover
* 🟥 va-progress-bar
* 🟥 va-progress-circle
* 🟥 va-radio
* 🟥 va-rating
* 🟥 va-rating/components/VaRatingItem
* 🟥 va-scroll-container
[...]

PR #3688, opened 2023-08-14
VaCarousel stories

Roman4437:
Made a few adjustments

See #3676


packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+}
+
+export const Stateful = () => ({
+  components: { VaCarousel },
+  ...carouselSetup,
+  template: `
+    [true] - should slide
+    <VaCarousel :items="items" stateful />
+    [false] - should not slide
+    <VaCarousel :items="items" />
```
asvae, line 65: We need to use `:stateful="false"` here.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+      v-model="areArrowsVisible"
+      class="mb-6"
+      label="Arrows"
+      size="small"
+    />
+    <VaCarousel :items="items" v-model="currentSlide" :arrows="areArrowsVisible" />
+  `,
+})
+
+export const Indicators = () => ({
```
asvae, line 109: Let's split `indicators` and `indicator-trigger` props into different stories and handle them similar to `stateful`.

`indicator-trigger` might also need story for `none`?

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+export const Infinite = () => ({
+  components: { VaCarousel, VaSwitch },
+  ...carouselSetup,
+  template: `
+    <VaCarousel :items="items" v-model="currentSlide" infinite />
+  `,
+})
+
+Infinite.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
```
asvae, line 138: We need to handle this similar to `stateful`.

One story with infinite (where navigate left works).
Another story without infinite (where navigate left doesn't work).

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+  })
 }
 
-export const Default = defineComponent({
+export const Autoscroll = () => ({
+  components: { VaCarousel, VaSwitch },
+  ...carouselSetup,
+  template: `
+    [default]
+    <VaCarousel :items="items" stateful autoscroll />
```
asvae, line 157: Let's have only one example here.

Autoscroll should behave as infinite always.

There is no case, where we need to go:
1,2,3,2,1.

We always need:
1,2,3,1

You can create a bug report or fix it on your own.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+    userEvent.click(slideFive)
+    await sleep()
+    userEvent.click(goForward)
+    await sleep()
+    expect(carousel[0].getAttribute('aria-current')).toEqual('true')
+  })
 }
 
-export const Default = defineComponent({
+export const Autoscroll = () => ({
```
asvae, line 152: Would be nice to have some basic interaction test here, maybe just wait a bit and check if slide changed. (slide changes without interaction)

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+  ...carouselSetup,
+  template: `
+    [default]
+    <VaCarousel :items="items" stateful autoscroll />
+    [infinite]
+    <VaCarousel :items="items" stateful autoscroll infinite />
+  `,
+})
+
+export const Fade = () => ({
```
asvae, line 163: Also please do 2 components (same as stateful).

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,203 @@
...
+  template: `
+    <VaCarousel :items="items" v-model="currentSlide" effect="fade" />
+  `,
+})
+
+export const Ratio = () => ({
+  components: { VaCarousel, VaSwitch },
+  ...carouselSetup,
+  template: `
+    <VaCarousel :items="items" v-model="currentSlide" :ratio="16 / 9" />
```
asvae, line 175: Let's make it 1/3 so that it's obvious ratio is changed. Right now it's hard to notice.

asvae (changes requested):
I only reviewed carousel for now.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
Roman4437: I think 1/1 should be better

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
+    <VaCarousel :items="items" :indicators="false" :arrows="false" v-model="current"/>
+  `,
+})
+
+Arrows.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
+  const slides = document.querySelectorAll('div.va-carousel__slide')
+  const NavigatesOnArrowClick = canvas.getByRole('button', { name: 'go next slide' }) as HTMLElement
+
+  await step('Should head to slide 2 by clicking on arrow', async () => {
```
asvae, line 51: Can we use test step naming like this? - "Changes slide on arrow click".

When test breaks it would be easier to understand what was broken. (as very likely if it doesn't navigate to slide 2 - then it doesn't navigate at all!)

Same idea apply for all other stories.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
+  const slides = document.querySelectorAll('div.va-carousel__slide')
+  const NavigatesOnPageClick = canvas.getByRole('button', { name: 'go slide 2' }) as HTMLElement
+
+  await step('Should head to slide 2 by clicking on indicator', async () => {
+    await userEvent.click(NavigatesOnPageClick)
+    expect(slides[1].getAttribute('aria-current')).toEqual('true')
+  })
+}
+
+export const IndicatorsTrigger = () => ({
```
asvae, line 79: My gut sense is that we should have

```
[click]
[hover]
[none]
```

It's totally possible for indicators to be non interactive. `click` is default.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
 
-export const Default = defineComponent({
+export const Infinite = () => ({
   components: { VaCarousel },
-  template: '<VaCarousel/>',
+  data: () => ({ items: getItems() }),
+  template: `
+    [infinite]
+    <VaCarousel :items="items" infinite />
+    [default]
```
asvae, line 114: please use actual values here: `true` and `false`. 

`infinite` in template actually means `infinite=true`, and no infinite at all means `infininte=false`. Even if doesn't look so.

It would probably be for the best to use infinite=true by default. I checked vuetify and naive ui and it seems meaningful from my experience.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
+}
+
+export const Stateful = () => ({
+  components: { VaCarousel },
+  data: () => ({ items: getItems() }),
+  template: `
+    [true] - should slide
+    <VaCarousel :items="items" />
+    [false] - should not slide
+    <VaCarousel :items="items" :stateful="false"/>
```
asvae, line 141: image

That seems broken to me. Even if you don't pass value - there should be a slide selected.
In this case - it should default to 1st slide.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
+    await userEvent.click(NavigatesOnPageClick)
+    expect(slides[4].getAttribute('aria-current')).toEqual('false')
+  })
+}
+
+export const Autoscroll = () => ({
+  components: { VaCarousel },
+  data: () => ({ items: getItems() }),
+  template: `
+    <VaCarousel :items="items" autoscroll :arrows="false" :indicators="false" />
```
asvae, line 165: Autoscroll is too fast right now. Should be once every 5s or so by default.

asvae, line 165: We also need a prop to define autoscroll-interval.

You can check implementation in here.
https://www.naiveui.com/en-US/os-theme/components/carousel

Maybe we can make it 500ms so that test goes faster.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
+    expect(slides[2].getAttribute('aria-current')).toEqual('true')
+  })
+}
+
+export const Fade = () => ({
+  components: { VaCarousel },
+  data: () => ({ items: getItems() }),
+  template: `
+    [fade]
+    <VaCarousel :items="items" effect="fade" />
```
asvae, line 186: There is no fade transition, so fade is actually broken.

You can mark stories as [stale] similarly to what I did in badges, then create an issue (here's issue I did for badges leftovers - #3625).

This way we'll have an easier time focusing on stories instead of spreading too thin.

packages/ui/src/components/va-carousel/VaCarousel.vue
```diff
@@ -127,7 +127,7 @@ export default defineComponent({
     ...useComponentPresetProp,
     ...VaImageProps,
 
-    modelValue: { type: Number, default: 0 },
```
asvae, line 130: Model value default would probably fix issue with stateful=false story.

asvae (changes requested):
Some more things.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
-import { defineComponent } from 'vue'
```
asvae, line 1: Please add a story with infinite for 0, 1, 2 slides. We can catch some edge cases there.

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
```diff
@@ -1,12 +1,224 @@
...
-import VaCarousel from './VaCarousel.demo.vue'
```
asvae, line 2: We need a story for swipe. Same as always (true - works, false - doesn't).

asvae (changes requested):

packages/ui/src/components/va-carousel/VaCarousel.stories.ts
asvae: Sorry for not being clear.

I meant
* `[infinite]` replace with `[true]`
*  and `[default]` replace with `[false]` 

Template overall was fine. Ping me in discord if it's unclear why this is needed.

image

asvae (approved):
Great work. Thanks!

Issue #3687, opened 2023-08-11
Accordion stateful doesn't work


asvae:
**Vuestic-ui version:** 1.7.4

https://www.chromatic.com/test?appId=64d608ab34b67e621ab50cfd&id=64d66e4526233b45a7bf232e

PR #3689, opened 2023-08-14
Fix#3687/stateful-accordion

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3687

### DIFF
--- a/packages/ui/src/components/va-accordion/VaAccordion.stories.ts
+++ b/packages/ui/src/components/va-accordion/VaAccordion.stories.ts
@@ -12,8 +12,9 @@ export default {
 
 export const Default = () => ({
   components: { VaAccordion, VaCollapse },
+  data: () => ({ value: [] }),
   template: `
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -12,8 +12,9 @@ export default {
 
 export const Default = () => ({
   components: { VaAccordion, VaCollapse },
+  data: () => ({ value: [] }),
   template: `
-    <va-accordion>
```
asvae, line 16: I suggest we make accordion stateful by default. (then you can make it false if you need that sweet sweet control)

I'd say - in 90% cases you as user don't need control over collapses.

asvae (changes requested):
Just one thing.

asvae (approved):
Perfect!

PR #3694, opened 2023-08-16
feat: add privacy policy

### LINKED ISSUE #3601: We need a privacy policy to use google analytics
Right now we don't have cookie notification which we need to do in order to use google analytics and potentially some other info-gathering services. So it's **mission critical** to have that consent as in europe you're obliged by law to do so. 

Should include 2 things:
- [ ] Cookie banner (show on first visit)
image

- [ ] Privacy policy page, similar to https://epicmax.co/privacy. Maybe check other frameworks docs to have a fuller picture.

Please in context of this effort also add for google-analytics module `enable: boolean` option. That's a better pattern for module. So the logic would be:

* enabled: true, id: specified - everything works
* enabled: true, id: unspecified - throws an error (not a warn).
* enabled: false, id: any - disables google-anallytics completely, doesn't show a warn or error.

- [ ] `enabled` flag is introduced.
- [ ] variables are updated in railway.app (ping @asvae or @m0ksem to do so).

asvae:
closes #3601

I chose a very simple approach to privacy policy - just add links to headers.
There are ways to improve it (add consent banner etc), but other ui solutions don't do it. It will also be harmful for user experience.

Links lead to iubenda - a site where said policies for each tool used (analytics, cdn etc) are being regularly updated, so we won't need to follow these up on our own.

image
image

### DIFF
--- a/packages/docs/components/landing/Footer.vue
+++ b/packages/docs/components/landing/Footer.vue
@@ -66,6 +66,10 @@
           >
             hello@epicmax.co
           </a>
+          <br>
+          <a href="https://www.iubenda.com/privacy-policy/50799764" title="Privacy Policy ">Privacy Policy</a>
+          •
+          <a href="https://www.iubenda.com/privacy-policy/50799764/cookie-policy" title="Privacy Policy ">Cookie Policy</a>
         </div>
       </div>
       <!--<LandingNewsBanner />-->


asvae:
@xx13 @zvenigorodskaia if you have an idea of better placement - I'm happy to hear.

asvae:
Maksim says it's ok.

PR #3699, opened 2023-08-16
added form stories

Roman4437:
Form stories


packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+        v-model="input"
+        label="with error message"
+        :rules="[value => 'error']"
+      />
+    </va-form>
+  `,
+})
+
+HideErrorMessages.play = async ({ step }) => {
+  const [firstDefinedInput, secondDefineInput] = document.querySelectorAll('div.va-input-wrapper > div.va-message-list')
```
asvae, line 179: I would suggest you either check for text directly or use aria-role. Classes are mostly for style and might change in unpredictable ways. Role stays stable. Text also does (if you define it).

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  data: () => ({ input: 'error' }),
+  template: `
+    [hidden]
+    <va-form
+      class="flex max-w-sm"
+      immediate
+      :hideErrors="true"
+    >
+      <va-input
+        v-model="input"
```
asvae, line 201: No need to use v-model. We're here not to test particular inputs.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+        :rules="[value => 'error']"
+      />
+    </va-form>
+  `,
+})
+
+HideErrors.play = async ({ step }) => {
+  const [firstInput, secondInput] = document.querySelectorAll('div.va-input-wrapper')
+
+  await step('Does not highlight invalid input', async () => {
```
asvae, line 224: If we don't have a role for errored input - that could be considered input's bug.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  await step('Highlights invalid input', async () => {
+    expect(secondInput).toHaveClass('va-input-wrapper--error')
+  })
+}
+
+export const Focus = () => ({
+  components: { VaForm, VaInput, VaButton },
+  template: `
+    <va-form 
+      class="flex space-x-2 mb-6 max-w-sm"
```
asvae, line 237: Probably we can also do without classes here.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+
+export const Focus = () => ({
+  components: { VaForm, VaInput, VaButton },
+  template: `
+    <va-form 
+      class="flex space-x-2 mb-6 max-w-sm" 
+      ref="focusForm"
+    >
+      <va-input
+        label="should be focused"
```
asvae, line 241: We can remove labels.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  const firstInput = canvas.getByRole('textbox', { name: 'should be focused' }) as HTMLElement
+  const secondInput = canvas.getByRole('textbox', { name: 'should not be focused' }) as HTMLElement
+  const button = canvas.getByRole('button', { name: 'Focus first input' }) as HTMLElement
+
+  await step('Focusses first input', async () => {
+    await userEvent.click(button)
+    expect(firstInput).toHaveFocus()
+  })
+
+  await step('Does not focus first input', async () => {
```
asvae, line 264: We can remove this interaction. Focus could be only on one element at once. So if it's on first - then it won't be on second.

We also don't need to test for switching focus on subsequent `focus()`, as that's just too hard to implement to be a random regression.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  const firstInput = canvas.getByRole('textbox', { name: 'valid' }) as HTMLElement
+  const secondInput = canvas.getByRole('textbox', { name: 'invalid' }) as HTMLElement
+  const button = canvas.getByRole('button', { name: 'Focus invalid' }) as HTMLElement
+
+  await step('Focusses invalid input', async () => {
+    await userEvent.click(button)
+    expect(secondInput).toHaveFocus()
+  })
+
+  await step('Does not focus invalid input', async () => {
```
asvae, line 302: Again, only one check is needed.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  })
+}
+
+export const ValidateAndResetValidation = () => ({
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ input: 'do not reset' }),
+  template: `
+    <va-form 
+      class="flex space-x-2 mb-6 max-w-sm" 
+      ref="validateAndResetValidationForm"
```
asvae, line 314: Let's replace all refs with "form". We used long names to prevent conflict, but there won't be a conflict with storybook as all stories are isolated.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+    >
+      <va-input
```
asvae, line 316: I don't think we need this input here.

It basically checks "validate method doesn't validate inputs without rules". Which is weird thing to check and also has no particular relation to method.

In this story we only need to check that methods do what they meant to do.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ username: 'root', password: 'admin' }),
+  template: `
+    <va-form
+      class="flex flex-col max-w-sm"
+      tag="form"
+      ref="resetForm"
+      immediate
+    >
+      <va-input
```
asvae, line 366: We can test all of this on one input, no need to have 3.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  })
+}
+
+export const Reset = () => ({
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ username: 'root', password: 'admin' }),
+  template: `
+    <va-form
+      class="flex flex-col max-w-sm"
+      tag="form"
```
asvae, line 362: Tag is not useful. (or shouldn't be)

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ isSubmit: false }),
+  template: `
+    <va-form
+      class="flex flex-col max-w-sm"
+      tag="form"
+      ref="formSubmit"
+      @submit.prevent="isSubmit = true"
+    >
+      <va-input
```
asvae, line 416: Let's just have one input here and one button. Remove label, class, unnecessary rules, div at the bottom.

asvae (changes requested):
Functionality is mostly fine, but stories are extremely bloated. We can cut about 50% of code without losing quality.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,443 @@
...
+Autofocus.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
+  const input = canvas.getByRole('textbox', { name: 'should have focus' }) as HTMLElement
+
+  await step('Focus first input', async () => {
+    expect(input).toHaveFocus()
+  })
+}
+
+export const Stateful = () => ({
```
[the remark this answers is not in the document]
m0ksem: Yeah. Stateful form means all its form fields are stateful.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
 }
 
-export const Default = defineComponent({
-  components: { VaForm },
-  template: '<VaForm/>',
+export const Default = () => ({
+  components: { VaForm, VaInput, VaButton },
+  template: `
+    <va-form ref="form">
+      <va-input :rules="[value => !!value || 'input is required']"/>
```
asvae, line 20: We can replace with `:rules="[Boolean]"` in all cases.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+    expect(input).toHaveFocus()
+  })
+}
+
+export const Stateful = () => ({
+  components: { VaForm, VaCheckbox },
+  template: `
+    [true]
+    <va-form stateful>
+      <va-checkbox role/>
```
asvae, line 57: Why role?

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+    </va-form>
+    [false]
+    <va-form :stateful="false">
```
asvae, line 60: The reason we want to keep stories like this looking like:
```
// for default false
va-form stateful
va-form
```
or
```
// for default true
va-form
va-form stateful=false
```

instead of 

```
va-form stateful=true
va-form stateful=false
```

is so that we can catch default change if someone does it in component. Default change could be considered a breaking change, so important to notice early.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+export const HideErrorMessages = () => ({
+  components: { VaForm, VaInput },
+  template: `
+    [true]
+    <va-form
+      role
+      immediate
+      hideErrorMessages
+    >
+      <va-input :rules="[value => !!value || 'input is required']"/>
```
asvae, line 118: We can use:

```
<va-input error error-messages="message"/>
```

and remove `immediate`.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+    </va-form>
+  `,
+})
+
+HideErrorMessages.play = async ({ step }) => {
+  const [firstDefinedInput, secondDefineInput] = document.querySelectorAll('[role] > div > div:nth-child(2)')
+
+  await step('Does not display validation error', async () => {
```
asvae, line 134: Displays error message. Not necessarily have anything to do with validation.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+export const HideErrors = () => ({
+  components: { VaForm, VaInput },
+  template: `
+    [true]
+    <va-form
+      role
+      immediate
+      hideErrors
+    >
+      <va-input :rules="[value => !!value || 'input is required']"/>
```
asvae, line 152: Same here `<va-input error error-messages="message"/>`

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+  await step('Highlights invalid input', async () => {
+    expect(secondInput).toHaveClass('va-input-wrapper--error')
+  })
+}
+
+export const Focus = () => ({
+  components: { VaForm, VaInput, VaButton },
+  template: `
+    <va-form ref="form">
+      <va-input />
```
asvae, line 181: For Focus and FocusInvalid we need 2 inputs in template.

If feature doesn't work properly - focus might select not first input.
For focusInvalid - it might be indifferent to input valid state - and we won't know from story.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+    </va-form>
+    <va-button @click="$refs.form.focusInvalidField()">
+      Focus invalid
+    </va-button>
+  `,
+})
+
+FocusInvalid.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
+  const input = canvas.getByRole('textbox', { name: '' }) as HTMLElement
```
asvae, line 214: Maybe you can think of some shortcuts for selectors like this?

I'm thinking something like:

```
const input = helper.get('input')
const button = helper.get('button')
```

or maybe even
```
const [input, button] = helper.get('input', 'button')
```

There is only one button and one input in story, so we can definitely add some sugar.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+    expect(input).toHaveFocus()
+  })
+}
+
+export const ValidateAndResetValidation = () => ({
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ input: 'do not reset' }),
+  template: `
+    <va-form ref="form">
+      <va-input v-model="input"/>
```
asvae, line 228: I don't see how this input is useful.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+      <va-input :rules="[() => 'error']"/>
```
asvae, line 229: `:rules="[Boolean]"` is fine here, we don't have to check for validation message.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+
+export const Reset = () => ({
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ username: 'reset' }),
+  template: `
+    <va-form
+      ref="form"
+      immediate
+    >
+      <va-input v-model="username"/>
```
asvae, line 267: one input is enough here

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,286 @@
...
+      <va-input :rules="[() => 'error']"/>
+    </va-form>
+    <va-button @click.prevent="$refs.form.reset()">
```
asvae, line 270: We need 2 buttons in total:
* `Fill` - (set value and run validation)
* `Reset` - refs.form.reset

Right now - it's not possible to interact manually.

asvae (changes requested):
Polish mostly. We're close.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,276 @@
...
+      <va-input :rules="[false]"/>
+    </va-form>
+  `,
+})
+
+Immediate.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
+  const [firstInput, secondInput] = canvas.getAllByRole('textbox', { name: '' }) as HTMLElement[]
+
+  await step('Displays error message', async () => {
```
m0ksem, line 100:
```suggestion
  await step('First input displays error message', async () => {
```

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,276 @@
...
+  })
+}
+
+export const Immediate = () => ({
+  components: { VaForm, VaInput },
+  template: `
+    [true]
+    <va-form immediate>
+      <va-input :rules="[false]"/>
+    </va-form>
```
m0ksem, line 88: Passing false directly prevents any interaction within the form in stories. I'd like to be able to play with the story and make sure there is no error if I typed something. Also, having a small message makes it easier to understand the story without digging into source code.

```suggestion
    <va-form immediate>
      Input is required and must show error if empty
      <va-input :rules="[(v) => v?.length > 0 || 'required']"/>
    </va-form>
```

m0ksem: I meant to do this to Default demo, not immediate.

packages/ui/src/components/va-form/VaFormNew.demo.vue
```diff
@@ -1,108 +0,0 @@
-<template>
```
asvae, line 1: Name `FormFormItemInteraction` and please keep it. This story tests all form components work all right with form.

Let's mark it as stale, as it should have better code (maybe some reuse from component stories and form stories). It's not mission critical though as it's fine to develop.

packages/ui/src/components/va-form/VaFormValidationLib.demo.vue
```diff
@@ -1,73 +0,0 @@
-<template>
```
asvae, line 1: Also should be a separate file: `VaFormVuelidate`

packages/ui/src/components/va-form/VaFormNew.stories.ts
```diff
@@ -1,12 +0,0 @@
-import { defineComponent } from 'vue'
```
asvae, line 1: We need to bring back these stories fully. Right now they won't be visible in storybook.

packages/ui/src/components/va-form/VaForm.stories.ts
```diff
@@ -1,12 +1,280 @@
...
+    await userEvent.click(resetButton)
+    expect(input.getAttribute('aria-invalid')).toEqual('false')
+  })
+}
+
+export const Reset = () => ({
+  components: { VaForm, VaInput, VaButton },
+  data: () => ({ data: '' }),
+  methods: {
+    set () {
```
asvae, line 250: Let's use different name here, maybe `fillForm`. `set` is a reserved word.

packages/ui/src/components/va-form/VaFormNew.stories.ts
asvae: We also need to polish these stories and move them to storybook. (you can move to different PR or leave an issue, but there should be a follow up - we can't mark form stories as reworked)

asvae (changes requested):
We need to decide what to do with validation lib and form interaction stories. These are not finished.

PR #3704, opened 2023-08-17
Fix/sidebar fixes

m0ksem:
👋


packages/docs/components/layout/Sidebar.vue
```diff
@@ -168,8 +168,6 @@ export default defineComponent({
 @import "@/assets/smart-grid.scss";
```
asvae, line 168: Add breaking change to the log.

packages/ui/src/components/va-sidebar/VaSidebar.vue
```diff
@@ -7,7 +7,7 @@
     @mouseenter="updateHoverState(true)"
     @mouseleave="updateHoverState(false)"
   >
-    <div class="va-sidebar__menu" :style="`width: ${computedWidth};`">
```
asvae, line 10: Sidebar "right" story is broken (was before your pr as well).

asvae, line 10: Sidebar "100% width" story is broken (aslwo was before your pr).

image

asvae, line 10: Minimized width story is broken (is less than 25%)

image

asvae, line 10: Hoverable minimized story seems to be broken (Should do somethin on hover? right now - does nothing)

image

asvae (changes requested):
Too many stories are broken right now. I see docs are all right, but still.

Not saying we need them perfect (that's something to pursue in stories rework).

m0ksem:
Because this PR makes breaking changes and all of this generally made for VaLayout component I'll merge this in feat/layout pr and close tihs PR.

PR #3712, opened 2023-08-18
Fix/chromatic false changes

### LINKED ISSUE #3686: Chromatic false negatives
We have a couple of places where we get false positive changes from chromatic:

image

You can also check it here:
https://www.chromatic.com/test?appId=64d608ab34b67e621ab50cfd&id=64d63d8ef9181a6ea5053d8e

* With images and transparent labels - the problem probably is with compression or something along that. I'm not sure how to solve it.
* Also we sometimes use randomly generated text or time. In that case - we can maybe keep randomness, but bind it to the seed, so that it stays the same for story.

My idea is to have a short investigation if we can handle this issue elegantly. And if we don't - we can exclude some stories  from checks or something else radicaly. End goal is to have no false positives for chromatic and not spend too much time.

asvae:
fix #3686 + tweaked stories a bit.


asvae:
Merged on my own. Too small.

PR #3713, opened 2023-08-19
Stories rework: VaAlert

Roman4437:
Alert stories


packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+      Message
+    </va-alert>
+  `,
+})
+
+export const Stateful = () => ({
+  components: { VaAlert },
+  template: `
+    [true]
+    <va-alert stateful closeable>
```
asvae, line 87: Alert should be stateful on default as well...

We forgot about stateful in a lot of places, sorry for this mess 🫠.

packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+  args: {
+    center: true,
+    title: 'Title',
+    description: 'Message',
+  },
+}
+
+export const Icon = {
+  args: {
+    icon: 'Icon',
```
asvae, line 43: This format is dangerous and I would avoid it for now. Somehow - what you pass to args - is not used as props, but as slots instead. You can compare to this story.

```
export const Icon = () => ({
  components: { VaAlert },
  template: `
    <va-alert icon="home" description="Message"/>
  `,
})
```

We probably have to declare something in meta (`export default`) for this format to work properly. Probably if component has both slot and prop - it prioritises slot. (wild guess)

packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+      Message
+    </va-alert>
+    [#ce6e67]
+    <va-alert border="left" border-color="#ce6e67">
+      Message
+    </va-alert>
+  `,
+})
+
+export const Sloted = () => ({
```
asvae, line 178:
```suggestion
export const Slotted = () => ({
```

packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+  `,
+})
+
+export const Hide = () => ({
+  components: { VaAlert, VaButton },
+  template: `
+    <va-alert stateful ref="alert">
+      Message
+    </va-alert>
+    <va-button @click="$refs.alert.hide()">
```
asvae, line 202: Let's add `show` method to Alert so that we can test it fully manually (right now you can't unhide alert).

asvae (changes requested):
This time it's very close to done on first try! Great progress.

packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+}
+
+export const Title = {
+  args: {
+    title: 'Title',
+    description: 'Message',
+  },
+}
+
+export const Center = {
```
asvae, line 33: For center we also need component with icon and close, something like this (ignore colors and messages, both should have title and description):

image

packages/ui/src/components/va-alert/VaAlert.stories.ts
```diff
@@ -1,12 +1,217 @@
...
+      Message
+    </va-alert>
+    [false]
+    <va-alert>
+      Message
+    </va-alert>
+  `,
+})
+
+export const CloseText = () => ({
```
asvae, line 69: We need another story for `closeIcon`. This prop is not mentioned in docs, so we need to mention it there as well - https://ui.vuestic.dev/ui-elements/alert#props.

asvae (approved):
🫡

PR #3714, opened 2023-08-19
Stories rewok: VaAppBar

Roman4437:
AppBar stories


packages/ui/src/components/va-app-bar/VaAppBar.stories.ts
```diff
@@ -1,12 +1,132 @@
...
+        id="target1" 
+        class="pb-[500px] overflow-auto"
+      >
+        {{ $vb.lorem(2500) }}
+      </div>
+    </div>
+  `,
+})
+
+export const ShadowOnScroll = () => ({
```
asvae, line 76: Would also be nice to have interaction demo.

packages/ui/src/components/va-app-bar/VaAppBar.stories.ts
```diff
@@ -1,12 +1,132 @@
...
+        id="target2" 
+        class="pb-[500px] overflow-auto"
+      >
+        {{ $vb.lorem(2500) }}
+      </div>
+    </div>
+  `,
+})
+
+export const ShadowColor = () => ({
```
asvae, line 96: Can we scroll this by default somehow? I'd like visual test to capture shadow color.

asvae (changes requested):
Pretty good.

Let's work on this some more.

packages/ui/src/components/va-app-bar/VaAppBar.stories.ts
Roman4437: it can't be helped

asvae (approved):

PR #3720, opened 2023-08-21
va-skip-link

Roman4437:
VaSkipLink component and stories

### DIFF
--- a/packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
+++ b/packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
@@ -1,0 +1,60 @@
+import { VaSkipLink } from './'
+
+export default {
+  title: 'VaSkipLink',
+  component: VaSkipLink,
+}
+
+export const Target = () => ({
+  components: { VaSkipLink },
+  template: `
+    <va-skip-link target="#target">
+      [skip link]
+    </va-skip-link>
+    <div :style="{ height: '100dvh' }">
+      [not target]
+    </div>
+    <div 
+      id="target" 
+      :style="{ height: '100dvh' }"
+    >
+      [target]
+    </div>
+  `,
+})
+
+export const PositionTopLeft = () => ({
+  components: { VaSkipLink },
+  template: `
+    <va-skip-link position="top-left">
+      [top-left]
+    </va-skip-link>
+  `,
+})
+
+export const PositionTopRight = () => ({
+  components: { VaSkipLink },
+  template: `
+    <va-skip-link position="top-right">
+      [top-right]
+    </va-skip-link>
+  `,
+})
+
+export const PositionBottomLeft = () => ({
+  components: { VaSkipLink },
+  template: `
+    <va-skip-link position="bottom-left">
+      [bottom-left]
+    </va-skip-link>
+  `,
+})
+
+export const PositionBottomRight = () => ({
+  components: { VaSkipLink },
+  template: `
+    <va-skip-link position="bottom-right">
+      [bottom-right]
+    </va-skip-link>
+  `,
+})
--- a/packages/ui/src/components/va-skip-link/VaSkipLink.vue
+++ b/packages/ui/src/components/va-skip-link/VaSkipLink.vue
@@ -1,0 +1,47 @@
+<template>
+  <a
+    class="va-skip-link"
+    role="link"
+    :href="target"
+    :style="position"
+  >
+    <slot />
+</a>
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue'
+import { usePosition } from './hooks/usePosition'
+
+export default defineComponent({
+  name: 'VaSkipLink',
+  props: {
+    target: { type: String, default: '' },
+    position: {
+      type: String as PropType<'top-right' | 'top-left' | 'bottom-right' | 'bottom-left'>,
+      default: 'bottom-right',
+      validator: (v: string) => ['top-right', 'top-left', 'bottom-right', 'bottom-left'].includes(v),
+    },
+  },
+
+  setup (props) {
+    const { position } = usePosition(props)
+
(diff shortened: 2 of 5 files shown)


packages/ui/src/components/va-skip-link/VaSkipLink.vue
```diff
@@ -0,0 +1,47 @@
...
+<script lang="ts">
+import { PropType, defineComponent } from 'vue'
+import { usePosition } from './hooks/usePosition'
+
+export default defineComponent({
+  name: 'VaSkipLink',
+  props: {
+    target: { type: String, default: '' },
+    position: {
+      type: String as PropType<'top-right' | 'top-left' | 'bottom-right' | 'bottom-left'>,
```
asvae, line 21: Don't we have this list defined for some other component? Seems like something we can reuse.

packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
```diff
@@ -0,0 +1,60 @@
+import { VaSkipLink } from './'
```
asvae, line 1: Please create a docs page in navigation section. Overall any changes should be combined with docs updates.

packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
```diff
@@ -0,0 +1,60 @@
...
+    <div 
+      id="target" 
+      :style="{ height: '100dvh' }"
+    >
+      [target]
+    </div>
+  `,
+})
+
+export const PositionTopLeft = () => ({
```
asvae, line 26: Maybe 4 stories for one prop is too much? We probably can leave only one + have a select or radio to select other positions. (that's an exception, that's not recommended approach in other stories as it's better to just position them side by side, which is not possible for skip link)

packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
```diff
@@ -0,0 +1,60 @@
+import { VaSkipLink } from './'
+
+export default {
+  title: 'VaSkipLink',
+  component: VaSkipLink,
+}
+
+export const Target = () => ({
```
asvae, line 8: Needs default demo. (this one seems pretty default to me) skip link doesn't work without target and slot.

packages/ui/src/components/va-skip-link/VaSkipLink.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <a
```
asvae, line 2: Can we also apply it to vuestic docs? Would be interesting to test if this works properly (doesn't trigger reload etc).

asvae (changes requested):
Missing docs are a major point! Implementation looks pretty good.

packages/docs/page-config/navigationRoutes.ts
```diff
@@ -342,6 +342,10 @@ export const navigationRoutes: NavigationRoute[] = [
         name: "sidebar-item",
         displayName: "Sidebar Item",
       },
+      {
```
asvae, line 345: Needs NEW label.

image

You can check https://github.com/epicmaxco/vuestic-ui/pull/3633

packages/ui/src/components/va-skip-link/VaSkipLink.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+      :style="{ height: '100dvh' }"
+    >
+      [target]
+    </div>
+  `,
+})
+
+const OPTIONS = ['top-left', 'top-right', 'bottom-left', 'bottom-right']
+
+export const Position = () => ({
```
asvae, line 29: Can't we add an easy interaction test here?

Maybe even only for one case.

The problem is that we have solid visual tests, but we won't click around for every story. But we need to somehow ensure core functionality stands.

packages/docs/page-config/ui-elements/skip-link/examples/Default.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <va-skip-link
```
asvae, line 2: This doesn't seem to work for me and I have an idea on how we can document this component without too much trouble.

* Let's make it work for docs globally.
* Then on this page - just note: "To test this component - click `Esc`, then press `Tab` and you should be able to see a button that would get you straight to content on `Enter`. That's perfect for users who rely on keyboard navigation." (you can improve text with chatgpt if you want).
* And we can remove example.

asvae (changes requested):
Doing great! Let's continue.

packages/docs/page-config/ui-elements/skip-link/index.ts
```diff
@@ -0,0 +1,20 @@
...
+import apiDescription from './api-description';
+
+export default definePageConfig({
+  blocks: [
+    block.title("Skip Link"),
+    block.paragraph("`va-skip-link` skip to the target component."),
+
+    block.subtitle("Basic usage"),
+
+    block.paragraph("By default, `va-skip-link` needs a target and position, please keep in mind that the component should be placed as the very first element."),
```
asvae, line 11: Can we add a short code example here? Just code. Right now it's not clear at all how to use component.

packages/docs/page-config/ui-elements/skip-link/index.ts
```diff
@@ -0,0 +1,20 @@
+import apiOptions from "./api-options";
+import apiDescription from './api-description';
+
+export default definePageConfig({
+  blocks: [
+    block.title("Skip Link"),
+    block.paragraph("`va-skip-link` skip to the target component."),
```
asvae, line 7: We need more text here. You can describe in technical terms what component does - then ask chat gpt to get you some text.

asvae:
We agreed I'll help with docs.

PR #3731, opened 2023-08-23
va-dropdown fixes

### LINKED ISSUE #3725: VaDropdown must pass slot attributes like isOpened, etc.
Must be a demo like

```html
<VaDropdown>
  <template #anchor="{ isOpen }">
      <va-icon :name="isOpen ? 'va-arrow-up' : 'va-arrow-down'" />
  </template>
  
  aaaa
</VaDropdown>
```

image

### LINKED ISSUE #3728: VaDropdown keep-anchor-width must have min-width=100%
If keep-anchor-width is set to true now only max-width is applied. Means, small menu still small, not the same width as anchor.

How to fix:
In useDropdown please add minWidth here:
image

Roman4437:
fixes #3728
fixes #3725

### DIFF
--- a/packages/ui/src/components/va-dropdown/hooks/useDropdown.ts
+++ b/packages/ui/src/components/va-dropdown/hooks/useDropdown.ts
@@ -74,8 +74,7 @@ export const useDropdown = (
             const reference = elements.reference
             const availableWidth = reference.getBoundingClientRect().width
             Object.assign(elements.floating.style, {
-              maxWidth: `${availableWidth}px`,
-              minWidth: '100%',
+              width: `${availableWidth}px`,
             })
           }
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/va-dropdown/hooks/useDropdown.ts
```diff
@@ -75,6 +75,7 @@ export const useDropdown = (
             const availableWidth = reference.getBoundingClientRect().width
             Object.assign(elements.floating.style, {
               maxWidth: `${availableWidth}px`,
+              minWidth: '100%',
```
m0ksem: minWidth: 100% for some reason have more importance than maxWidth.

image


```suggestion
              width: `${availableWidth}px`,
```

m0ksem (approved):
I've added docs and two more methods (`getAnchorWidth`, `getAnchorHeight`), so it is easier to render whatever user wants.

image

PR #3739, opened 2023-08-25
fix(dropdown): prevent dropdown content from taking 100%

m0ksem:


asvae:
Didn't reproduce for me, but also no harm. So I'll merge.

Next time please get clearer reproduction.

asvae (approved):

m0ksem:
> Didn't reproduce for me, but also no harm. So I'll merge.
> 
> Next time please get clearer reproduction.

It was like this for me before:
image

PR #3747, opened 2023-08-27
VaDatePicker a11y

### LINKED ISSUE #3744: a11y: va-date-picker

Roman4437:
closes #3744 

added essential roles and attributes

### DIFF
--- a/packages/ui/src/components/va-date-picker/components/VaDayPicker/VaDayPicker.vue
+++ b/packages/ui/src/components/va-date-picker/components/VaDayPicker/VaDayPicker.vue
@@ -1,9 +1,14 @@
 <template>
-  <div class="va-day-picker" v-bind="containerAttributes">
+  <div 
+    class="va-day-picker" 
+    v-bind="containerAttributes"
+    role="grid"
+  >
     <template v-if="!hideWeekDays">
       <div
         v-for="weekday in weekdayNamesComputed" :key="weekday"
         class="va-day-picker__weekday"
+        role="gridcell"
       >
         <slot name="weekday">
           {{ weekday }}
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-date-picker/components/VaDayPicker/VaDayPicker.vue
```diff
@@ -17,6 +22,8 @@
       :key="index"
       @mouseenter="hoveredIndex = index"
       @mouseleave="hoveredIndex = -1"
+      :aria-label="date.toLocaleDateString(undefined, { year: 'numeric', month: 'long', day: 'numeric' })"
```
m0ksem, line 25: There must be `format` prop, which you can use here.

```suggestion
      :aria-label="$props.format(date)"
```

And probably aria-label must be applied to va-data-picker-cell, which has @click. And most likely date-picker-cell with @click must have `role="button"`.

m0ksem (changes requested):
Please use format prop in aria-label.

Also, why did you skip role "presentation", "aria-live " and "aria-atomic"? We change grid using arrow buttons, so it must be applied here I think (Im not sure).

These changes also must be applied to months picker and something must be changed in year picker.

m0ksem:
I checked out that gridcell must be used in row. It is a requirement, otherwise gridcell will not be recognized. We're using css grid here, so we don't have an actual row element... 

I think we need to change component layout from css grid, to more general approach with row

roles:
```
grid
  row
     columnheader <- weekdays
  rowgroup
     row
       gridcell <- date
 ```
 
What if we make VaGrid component, which will implement all of these with keyboard navigation? I think it might be useful for people who make something like cinema chair picker or whatever requires navigation with grid. We could reuse VaGrid in DatePicker. I think for this PR we can simply remove wrong roles and implement VaGrid component later.

Issue #3746, opened 2023-08-27
Vuestic x Nuxt 3.7 - Not coming out of terminal – stuck in npm install 


mrniamster:
**Vuestic-ui version:** 
 "@vuestic/nuxt": "^1.0.14",
 
**Steps to reproduce**
Removing "@vuestic/nuxt" from modules in nuxt.config.ts fixes the issue

https://stackblitz.com/edit/github-khexbq?file=nuxt.config.ts,package.json

**What is the expected behavior?**
Should be working with NUXT 3.7, and able to install .

**What is the current behavior?**
Stuck in installing dependencies,
<img width="667" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/52563039/db6b5517-5f66-4bd2-8ad8-cf5581a3965c">

**Other information**
Previously reported: https://github.com/nuxt/nuxt/issues/22844

mrniamster:
it was working well in nuxt 3.5.2 release && same "@vuestic/nuxt": "^1.0.14",

m0ksem:
Hi, @mrniamster. Thanks for report. Can you check if `@vuestic/nuxt@1.0.15-pre-release` version works great?

PR #3748, opened 2023-08-27
fix(nuxt): use built-in `watch` option to restart nuxt

danielroe:
## Description
closes https://github.com/epicmaxco/vuestic-ui/issues/3746

Since https://github.com/nuxt/nuxt/pull/19530, we have a built-in way to watch files + restart Nuxt when they change. This means we don't need to register additional watchers in modules.

Spotted the issue in https://github.com/nuxt/nuxt/issues/22844. The watcher should probably have been set to be closed when nuxt closed (something like `nuxt.hook('close', () => watcher.close())`. Nevertheless, it's likely a bug we need to fix in nuxt/cli.

However, this PR works around the issue and also should improve performance in vuestic/nuxt by reducing watchers required.

## Types of changes

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -28,7 +28,6 @@
   },
   "dependencies": {
     "@nuxt/kit": "^3.0.0",
-    "chokidar": "*",
     "nuxt": "^3.1.1",
     "pathe": "^0.3.5",
     "vuestic-ui": "^1.4.10"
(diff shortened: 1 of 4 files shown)


m0ksem (approved):
Thanks!

mrniamster:
Thanks. @m0ksem  @danielroe  <3

PR #3774, opened 2023-08-30
feat: github label sync

asvae:
Small utility to sync labels between repos.

Meant to make managing multiple repos less annoying.


m0ksem (approved):
Maybe move it to separate repo? It is not related to UI and repo is big enough already. It might be painful to sync deps in future.

Otherwise, looks OK, but I haven't run it locally.

asvae:
@m0ksem it uses same deps as release script and I don't envision any change with that.

Did you have a problem with monorepo dependencies related to anything except for vue?

I would hope to add more small projects to vuestic-ui repo.

With github-label - right now it's more of a general script, but I see it being used as part of automation/check of some sorts.

Issue #1870, opened 2022-05-30
`VaModal` overlay when there are nested modals


RVitaly1978:
**Vuestic-ui version:** 1.4.1

**Steps to reproduce**
- use nested `VaModal`: the second modal should be smaller than the first one

**What is the expected behavior?**
- when only the first modal is opened, the overlay should be below the first modal container. And when the second modal is open, the overlay should be under the second modal container, but before the first one

**What is the current behavior?**
![chrome-capture-2022-4-30 (1)](https://user-images.githubusercontent.com/55198465/171012432-f6bc5a0c-7d31-44cc-b464-c371b418436e.gif)

m0ksem:
It depends on preference... Must be a configurable with prop.

Udaay:
Picking this one

m0ksem:
> Picking this one

Make sure this feature will be as separate prop.

Udaay:
Hello @m0ksem @RVitaly1978 
I wanted to share with you the results of my investigation regarding the issue of the overlay not showing up for the `nested modal`, or the `lowest level modal.` After some digging, I found that this problem is caused by a [this line of code ](https://github.com/epicmaxco/vuestic-ui/blob/92eb61c24df6ce29c1a6561b8096e26d815a48b6/packages/ui/src/components/va-modal/VaModal.vue#L240)in the VaModal.vue file. Specifically, the line of code does not add a background to the overlay when it is the lowest level modal.
Removing this  [`!isLowestLevelModal.value`]  from `if`condition solves the issue.

m0ksem:
> Hello @m0ksem @RVitaly1978 I wanted to share with you the results of my investigation regarding the issue of the overlay not showing up for the `nested modal`, or the `lowest level modal.` After some digging, I found that this problem is caused by a [this line of code ](https://github.com/epicmaxco/vuestic-ui/blob/92eb61c24df6ce29c1a6561b8096e26d815a48b6/packages/ui/src/components/va-modal/VaModal.vue#L240)in the VaModal.vue file. Specifically, the line of code does not add a background to the overlay when it is the lowest level modal. Removing this [`!isLowestLevelModal.value`] from `if`condition solves the issue.

[...]

PR #3788, opened 2023-09-03
fix: added props based support for nested overlay

Udaay:
## Description
Added a props `isShowNestedOverlay` props by default, this props is set to 'false'. When you enable it for a modal, an overlay will be displayed for that modal and will not get removed in any case.

Discussion : #1870 

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/va-modal/VaModal.vue
@@ -198,6 +198,7 @@ export default defineComponent({
     withoutTransitions: { type: Boolean, default: false },
     overlay: { type: Boolean, default: true },
     overlayOpacity: { type: [Number, String], default: 0.6 },
+    isShowNestedOverlay: { type: Boolean, default: false },
     blur: { type: Boolean, default: false },
     zIndex: { type: [Number, String], default: undefined },
     backgroundColor: { type: String, default: 'background-secondary' },
@@ -242,12 +243,15 @@ export default defineComponent({
       // Supposedly solves some case when background wasn't shown.
       // As a side effect removes background from nested modals.
 
-      if (!props.overlay || !isLowestLevelModal.value) { return }
+      if (!props.overlay) { return }
 
-      return {
-        'background-color': `rgba(0, 0, 0, ${props.overlayOpacity})`,
-        'z-index': props.zIndex && Number(props.zIndex) - 1,
-      } as StyleValue
+      if (isTopLevelModal.value || props.isShowNestedOverlay) {
+        return {
+          'background-color': `rgba(0, 0, 0, ${props.overlayOpacity})`,
+          'z-index': props.zIndex && Number(props.zIndex) - 1,
+        } as StyleValue
+      }
+      return ''
     })
 
     const show = () => { valueComputed.value = true }
--- a/packages/ui/src/components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/va-modal/VaModal.vue
@@ -12,7 +12,7 @@
     </div>
 
     <teleport :to="attachElement" :disabled="$props.disableAttachment">
-      <modal-element
+      <WithTransition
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -198,6 +198,7 @@ export default defineComponent({
     withoutTransitions: { type: Boolean, default: false },
     overlay: { type: Boolean, default: true },
     overlayOpacity: { type: [Number, String], default: 0.6 },
+    isShowNestedOverlay: { type: Boolean, default: false },
```
m0ksem, line 201:
```suggestion
    showNestedOverlay: { type: Boolean, default: false },
```

packages/ui/src/components/va-modal/VaModal.vue
```diff
@@ -242,12 +243,15 @@ export default defineComponent({
...
-      if (!props.overlay || !isLowestLevelModal.value) { return }
+      if (!props.overlay) { return }
 
-      return {
-        'background-color': `rgba(0, 0, 0, ${props.overlayOpacity})`,
-        'z-index': props.zIndex && Number(props.zIndex) - 1,
-      } as StyleValue
+      if (isTopLevelModal.value || props.isShowNestedOverlay) {
+        return {
+          'background-color': `rgba(0, 0, 0, ${props.overlayOpacity})`,
```
m0ksem, line 250: Would be nice if we can make nestedOverlay opacity smaller... Because with 3 nested modals overlay become so much darken. 


 How about using css variables here instead of hardcoding?
 
 ```
 :root {
    --va-modal-overlay-color: rgb(0, 0, 0);
    --va-modal-overlay-opacity: 0.6;
    --va-modal-overlay-nested-opacity: 0.1;
 }
 ```
 
 We can remove backgroundOpacity prop.

m0ksem (changes requested):
Looks good. Thanks! I have few suggestions.

packages/ui/src/components/va-modal/VaModal.vue
Udaay: By removing `overlayOpacity` from props, one cannot control the overlay that will be okay ?

m0ksem: I think user still have `overlay` option to turn it off. In case user wants opacity to be different from 0.6, it can be changed using CSS variables.

Udaay:
```
 if (isTopLevelModal.value || props.ShowNestedOverlay) {
        return {
          'background-color': 'var(--va-modal-overlay-color)',
          opacity: `${props.ShowNestedOverlay ? 'var(--va-modal-overlay-nested-opacity)' : 'var(--va-modal-overlay-opacity)'} `,
          'z-index': props.zIndex && Number(props.zIndex) - 1,
        } as StyleValue
      }
```

Something like this ?

Udaay: https://github.com/epicmaxco/vuestic-ui/assets/26655564/329cd724-7669-41b0-8a77-ec27a706fd31

By using this approach I'm able to see the black background flickering on Clicking Cancel of Nested Dialog

m0ksem: > Screen.Recording.2023-09-05.at.9.39.24.PM.mov 
> By using this approach I'm able to see the black background flickering on Clicking Cancel of Nested Dialog

If you want me to help you, please update your PR.

Udaay: have updated my PR

m0ksem (approved):
It was a little bit of pain to make animations works. There was a blink because of Transition element. We need to hide overlay immediately if nested modal is disappearing, otherwise two overlays visible at the time.

PR #3800, opened 2023-09-06
Stories rework: VaAvatarGroup

Roman4437:


asvae (approved):
Thanks. I tweaked stories a bit and moved leftovers to #3879

Issue #3789, opened 2023-09-04
Accordion v-model not effective


iynehz:
**Vuestic-ui version:** 1.7.6

**Steps to reproduce**

Take an accordion example like the default one https://ui.vuestic.dev/ui-elements/accordion#default , change that “value” array from `[false, false, false]` to like `[true, false, false]`, restart server and see if the first collapse expands or not.

**What is the expected behavior?**

A default true value in accordion's v-model array should expand the corresponding collapse on page load.

**What is the current behavior?**

A true value in accordion's v-model array is not effective.

**Other information**

version 1.7.5 is good. So it breaks since 1.7.6.

m0ksem:
Must be fixed in 1.7.7 and 1.7.8

PR #3833, opened 2023-09-07
fix(#3789): accordion ignores v-model

m0ksem:
## Description

closes(#3789 )

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes


packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -52,6 +52,38 @@ Default.play = async ({ canvasElement, step }) => {
...
 
+export const vModel = () => ({
+  components: { VaAccordion, VaCollapse },
+  data: () => ({ value: [true, false, true] }),
+  template: `
+    <va-accordion v-model="value" multiple>
+      <va-collapse
+        v-for="i in 3"
+        :key="i"
+        :header="value[i - 1] ? 'Must be opened' : 'Must be closed'"
```
asvae, line 63: Let's not have any text description in story apart from what is absolutely necessary
```suggestion
        header="Collapse"
```

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
```diff
@@ -52,6 +52,38 @@ Default.play = async ({ canvasElement, step }) => {
...
+      >
+        Content
+      </va-collapse>
+    </va-accordion>
+  `,
+})
+vModel.play = async ({ canvasElement, step }) => {
```
asvae, line 70: Do we need this story at all? `Multiple` story does pretty much the same and even more.

packages/ui/src/components/va-accordion/VaAccordion.stories.ts
m0ksem: Previously it was hidden even if header was "Must be opened". So, at least it helped me to fix a bug. More, it probably will prevent this bug in future.

PR #3835, opened 2023-09-07
Stories rework: VaBacktop

Roman4437:
VaBacktop stories


packages/ui/src/components/va-backtop/VaBacktop.stories.ts
```diff
@@ -1,12 +1,79 @@
...
+  template: `
+    <p class="m-1">{{$vb.lorem(10000)}}</p>
+    <va-backtop 
+      horizontalOffset="4rem" 
+      verticalOffset="4rem"
+    />
+  `,
+})
+
+const OPTIONS_X = ['left', 'right']
```
asvae, line 36: I suggest we move these to data block. So that it's clear OPTIONS are going to be used only for that story and not in some other stories.


```
data: () => {
  const OPTIONS_X = ['left', 'right']
  const OPTIONS_Y = ['top', 'bottom']
  return { optionsY: OPTIONS_Y, valueY: OPTIONS_Y[1], optionsX: OPTIONS_X, valueX: OPTIONS_X[1]}
},
```

packages/ui/src/components/va-backtop/VaBacktop.stories.ts
```diff
@@ -1,12 +1,79 @@
...
+    <va-radio v-model="valueX" :options="optionsX"/>
+    <p class="m-1">{{$vb.lorem(10000)}}</p>
+    <va-backtop 
+      :horizontalPosition="valueX"
+      :verticalPosition="valueY"
+    />
+  `,
+})
+
+export const ClikcEvent = () => ({
```
asvae, line 55:
```suggestion
export const ClickEvent = () => ({
```

packages/ui/src/components/va-backtop/VaBacktop.stories.ts
```diff
@@ -1,12 +1,79 @@
...
+import { VaBacktop } from './'
+import { VaRadio } from '../va-radio'
 
 export default {
   title: 'VaBacktop',
   component: VaBacktop,
 }
 
-export const Default = defineComponent({
+export const Default = () => ({
```
asvae, line 9: I think we need an interaction test for this story (all others are fine without).

packages/ui/src/components/va-backtop/VaBacktop.stories.ts
```diff
@@ -1,12 +1,79 @@
...
   components: { VaBacktop },
-  template: '<VaBacktop/>',
+  template: `
+    <p class="m-1">{{$vb.lorem(10000)}}</p>
+    <va-backtop/>
+  `,
 })
+
+export const Color = () => ({
```
asvae, line 17: I think for all stories we need some sort of "scroll into view" initial setup.

I think the preferable way to do this would be in mounted hook, but it's probably also fine to do as part of interaction script.

The problem right now is that for every story - you have to scroll manually for component to appear, otherwise you only see lorem.

This makes visual tests useless and testing cumbersome.

asvae (changes requested):
I have a couple of ideas how we can achieve a bit more polish.

Looks nice overall.

packages/ui/src/components/va-backtop/VaBacktop.stories.ts
```diff
@@ -1,12 +1,79 @@
...
+      :verticalPosition="valueY"
+    />
+  `,
+})
+
+export const ClikcEvent = () => ({
+  components: { VaBacktop },
+  template: `
+    <p class="m-1">{{$vb.lorem(10000)}}</p>
+    <va-backtop ref="backtop"/>
```
asvae, line 59: This demo is incorrect, event should be on va-backtop. 

So something like this:
```
va-backtop @click="clicked = true"

clicked: {{clicked}}
```

Interaction demo would also be nice here

asvae:
@Roman4437 would be cool if you could review my changes, was late, might have missed something :).

asvae (approved):
🤟

PR #3836, opened 2023-09-07
Stories rework: VaBreadcrumbs 

Roman4437:
VaBreadcrumbs stories


packages/ui/src/components/va-breadcrumbs/VaBreadcrumbs.demo.vue
```diff
@@ -1,194 +0,0 @@
...
-        </template>
-
-        <va-breadcrumbs-item label="One" />
-        <va-breadcrumbs-item label="Two" />
-        <va-breadcrumbs-item label="Three" />
-      </va-breadcrumbs>
-      <br>
-    </VbCard>
-
-    <VbCard title="Disabled">
```
asvae, line 133: Disabled prop is not covered right now.

asvae (approved):
Looks good!

Let's add one demo for Default and call it done.

Issue #3838, opened 2023-09-08
`VaDataTable` `no-data` class has no default paddings


RVitaly1978:
**Vuestic-ui version:** 0.0.0-experimental-f060346c3-20230830

- `no-data` class by default must have the same paddings as table cell (`td`)

![image](https://github.com/epicmaxco/vuestic-ui/assets/55198465/e11a8fc8-97e0-405c-96be-41f39e6ec088)

![image](https://github.com/epicmaxco/vuestic-ui/assets/55198465/4d24c11a-0169-4c54-8bd4-a150a308728c)

Issue #3837, opened 2023-09-08
`VaDataTable` does not emit `filtered` immediately


RVitaly1978:
**Vuestic-ui version:** 0.0.0-experimental-f060346c3-20230830

`immediate: true` must be added for `watch` in the **useFilterable** hook

![image](https://github.com/epicmaxco/vuestic-ui/assets/55198465/be5462dc-2a75-47a3-b15a-616dd81b7abc)

PR #3851, opened 2023-09-09
fix(#3837): emit filter event immidiate if needed

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3838
closes https://github.com/epicmaxco/vuestic-ui/issues/3837


packages/ui/src/components/va-data-table/VaDataTable.stories.ts
```diff
@@ -1,12 +1,45 @@
...
-  component: VaDataTable,
+  component: VaDataTableDemo,
 }
 
+const columns = defineVaDataTableColumns([
+  { key: 'name' },
+  { key: 'email' },
+])
+
+const items = defineVaDataTableItems<typeof columns>([
```
m0ksem, line 16: Create satisfies VaDataTableColumns<Items> and satisfies VaDataTableItems<Columns>

PR #3852, opened 2023-09-11
fix: docs related sliders overflow

### LINKED ISSUE #3819: VaSlider goes overflow
**Vuestic-ui version:** 1.7.6

When you click first slider to the right - sliders of the bottom example overflow.
https://ui-develop.vuestic.dev/ui-elements/slider#track-label

image

Stars22:
- fixes related slider overflow: when the top slider value(set to min) becomes greater then the max value of bottom sliders close #3819

image


packages/docs/page-config/ui-elements/slider/examples/TrackLabel.vue
```diff
@@ -1,12 +1,12 @@
 <template>
```
asvae, line 1: Doesn't seem to fix it.

image

asvae (changes requested):

packages/docs/page-config/ui-elements/slider/examples/TrackLabel.vue
Stars22: yep, my bad. missed the second slider. fixed

asvae: Ok, so we fixed the docs example. We didn't fix the underlying issue.

Range should be operational even if min/max are swapped. It's reasonable to expect user would do something similar in app.

We could either fix it in this PR or move to another issue, but this has to be addressed.

Stars22: ok. I think it's right to have fix in this pr. I'll check what we can do.

Stars22: I've created a possible solution,  I can't say how it's optimal as it was not a trivial task to begin with )

Stars22: story is kind of draft not intended for merge,  maybe will be helpful to play around with sliders

Issue #3845, opened 2023-09-08
Remove `background-tertiary` from docs


m0ksem:
We don't have this color anymore. Use `background-element` or `background-border`. Please make it look great!

peoray:
this should be closed

PR #3855, opened 2023-09-11
fix: remove background-tertiary from docs

thesheppard:
fixes use of non existent color `background-tertiary` https://github.com/epicmaxco/vuestic-ui/issues/3845

## Types of changes


m0ksem (approved):
Thanks!

PR #3860, opened 2023-09-12
Story rework: VaVirtualScroller

xiongmao86:
Transform VaVirtualScroller to storybook.
Part of #3676.

## Types of changes


xiongmao86:
Hi, I have most props test by step in storybook. But there are two issue I don't know how to handle:
1. `track-by` prop is assign to `key` attribute, which is handle and remove by `vue`, and won't show up in the DOM of browser, how can I test for it?
2. In the story of `DifferentSizesAndMargins` the bottom padding and button size is different in even and odd list items, I am not sure how to write test for it, I don't know how the size are calculated, I can work it out by reading the code, but I am not sure whether that is a good way, would it be rigid and not flex?

Is there anything I have missed to do?

xiongmao86:
Is it normal for UI Tests to fail? I haven't see anything related to `VaVirtualScroller` in details?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+      :wrapper-size="200"
+      v-slot="{item}"
+    >
+      <div>{{ item }}</div>
+    </va-virtual-scroller>
+  `,
+})
+
+Default.play = async ({ step }) => {
+  const scroller = document.querySelector('.va-virtual-scroller') as HTMLElement
```
asvae, line 34: I suggest we use `canvasElement.querySelector` instead of `document`. It's more local.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+
+  await step('wrapper height is 200px', async () => {
```
asvae, line 36: We can not check height in interaction tests - as that would be tested in visual tests in chromatic anyway.

That's applicable to all tests.

asvae: We generally want to test only functionality that's relevant to the story. So in case of `disabled` that would be disabled-ness itself 🙃 etc.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+  data: () => ({ hugeArray }),
+  template: `
+    <p>[bench = 0]</p>
+    <va-virtual-scroller
+      :items="hugeArray"
+      :bench="0"
+      :wrapper-size="200"
+      v-slot="{item, index}"
+    >
+      <div>{{ index }} - {{ item }}</div>
```
asvae, line 130: Let's leave `item` here. We're not testing index.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+
+CustomKey.play = async ({ step }) => {
+  const scroller = document.querySelector('.va-virtual-scroller') as HTMLElement
+
+  await step('wrapper height is 200px', async () => {
+    expect(scroller.style.height).toEqual('200px')
+  })
+}
+
+export const LoopingComponent = () => ({
```
asvae, line 183: We can remove this story. It's not useful.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+  const canvas = within(canvasElement)
+  const list = canvas.findByRole('list')
+  const scroller = document.querySelector('.va-virtual-scroller') as HTMLElement
+
+  await step('wrapper height is 200px', async () => {
+    expect(scroller.style.height).toEqual('200px')
+  })
+}
+
+export const RemWrapperAndItemSizeValue = () => ({
```
asvae, line 235: This story is not great. Instead we need:
* One story for `wrapper-size`
  * You can make it similarly to [this story](https://ui-storybook-develop.vuestic.dev/?path=/story/vaalert--border).
  * Should have: [100] (default, please don't define a prop), [auto], [number px], [number rem].
* One story for `item-size`
  * 0 [default], [number], [number px], [number rem]
* You can select any number you see appropriate. 
* Let's also remove other wrapper size and item-size demos as they become redundant.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+const hugeArray = hugeArrayBase.fill(null).map((_, index) => index)
+const hugeObjectsArray = hugeArrayBase.fill(null).map((el, index) => ({ value: index }))
+
+export const Default = () => ({
+  components: { VaVirtualScroller },
+  data: () => ({ hugeArray }),
+  template: `
+    <va-virtual-scroller
+      :items="hugeArray"
+      :wrapper-size="200"
```
asvae, line 25: Let's remove this line from everywhere except for `wrapper-size` story.

asvae (changes requested):
Hey, great work!

There is a bit or redundancy in old demos and this PR, apart from that - things are perfect.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
xiongmao86: Hi, @asvae, thanks for reviewing my pr. After reading your comment, I am assuming story without play function are worth having because the story is taken in considerate in visual tests, is that correct?

When reading documents at storybook, I have only vague understanding of visual test. I know it's testing the visual feature of the component, I know it's going to run in chromatic very pr is made. But that's all.

Could you tell me how do you get to know the specific knowledge to make that call?

asvae:
> Is it normal for UI Tests to fail? I haven't see anything related to `VaVirtualScroller` in details?

Generally no (tests pass on develop). But you can ignore that for this PR 😁 as they have nothing to do with your changes (likely tests are broken because you branched off a while ago).

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+}
+
+export const CustomKey = () => ({
+  components: { VaVirtualScroller },
+  data: () => ({ hugeObjectsArray }),
+  template: `
+    <va-virtual-scroller
+      :items="hugeObjectsArray"
+      :bench="10"
+      track-by="value"
```
asvae, line 166: You asked about how to test this story. I think we can ignore this for now.

I moved your concern to #3865, which is like 10 times less prio than current effort.

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
asvae: @xiongmao86 
> I am assuming story without play function are worth having
Absolutely yes. Actually, that's majority of value they provide. Developers would check relevant stories manually while working on components. Visual tests and interaction tests - are mostly safeguards to prevent paranoia :).

I'll try to answer your question if I understood it correctly.

Right now we have visual and interaction tests in [chromatic](https://www.chromatic.com/build?appId=64d608ab34b67e621ab50cfd&number=303).

All stories are covered with visual tests and chromatic shows us diffs between latest stable version and current version. So that we can approve change. That's enabled for all stories currently.

Visual tests motivate us to make stories visual in a sense that when feature breaks - something visual will change. That would allow chromatic to inform us and likely this change won't pass review and be bounced back.

In case of virtual scroll container size - if feature breaks - container size would change and that would be visible in chromatic.

Another thing storybook does - is interaction tests. (that's not the same as visual test) These are pretty much short e2e tests and allow us to catch stuff that is not detectable by visual tests or set up the story in specific way (scroll to certain point to make component visible etc).

We want to have interaction tests only in critical places and in pretty basic form.

I.e. we have `color` feature and it's big feature and you'll have a big interaction test if you want to test it fully. But in stories we only want to test "color feature works at all", not delve into details. Details should be handled in unit tests or specialized stories that pretty much test composabled directly (I don't have an example of that sadly, just an idea for now 😬).

Hope that helps.

xiongmao86: I am not very sure what do you mean by testing disabled-ness, 

I have found 3 occurrence of `disabled` prop. If `disabled` is true: 
1. `useEvent('scroll', handleScroll, true)` is not called, 
2. items is not dynamically loaded
3. `isDisabledVirtualTable` is false
by testing disabled-ness do you mean by testing on the scroll event in point 1?

xiongmao86: I am not very sure what do you mean by testing disabled-ness, 

I have found 3 occurrence of `disabled` prop. If `disabled` is true: 
1. `useEvent('scroll', handleScroll, true)` is not called, 
2. items is not dynamically loaded
3. `isDisabledVirtualTable` is false
by testing disabled-ness do you mean by testing on the scroll event in point 1?

 I have checked that `disabled` scroller rolls in respond to mouse wheels just like the non disabled scroller, how do I test for that?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+  const lists = canvas.getAllByRole('list')
+
+  await step('wrapper height is 200px', async () => {
+    expect((scrollers[0] as HTMLElement).style.height).toEqual('200px')
+    expect((scrollers[1] as HTMLElement).style.height).toEqual('200px')
+  })
+
+  await step('the virtual scroller that has 20 bench is 40 items larger than no bench virual scroller', async () => {
+    expect(lists[1].children.length - lists[0].children.length).toBe(40)
+  })
```
xiongmao86, line 156: Should I remove this? Doesn't seem to be a interaction test?

packages/ui/src/components/va-virtual-scroller/VaVirtualScroller.stories.ts
```diff
@@ -1,12 +1,285 @@
...
+    </va-virtual-scroller>
+  `,
+})
+
+NoItemsPassed.play = async ({ canvasElement, step }) => {
+  const canvas = within(canvasElement)
+  const list = canvas.getByRole('list')
+
+  await step('There is no item to show', async () => {
+    expect(list.children.length).toBe(0)
```
xiongmao86, line 283: This doesn't looks like an interaction test, should I remove it?

xiongmao86:
Thanks, @asvae, How about the code now, anything I have missed?

asvae:
I did a couple of small changes, but overall it's perfect 🤗.

xiongmao86:
Thank you very much, @asvae.

PR #3863, opened 2023-09-12
docs: minor miscellaneous improvements

### LINKED ISSUE #3811: Let's delete this page. It's deprecated.
image

### LINKED ISSUE #3813: VaBadge docs
**Vuestic-ui version:** 1.7.6

image

image

### LINKED ISSUE #3814: VaButton docs
**Vuestic-ui version:** 1.7.6

Icons too close:
image

Add some margin, check other examples
image

### LINKED ISSUE #3815: VaCarousel docs
**Vuestic-ui version:** 1.7.6

Let's use a bit different ratio to make it look better on mobile.
image

Let's use different text for slots, maybe something like [1]

thesheppard:
closes #3811 (Remove deprecated Grid System page)
closes #3813 (VaBadge docs overflow)
closes #3814 (VaButton docs overflow)
closes #3815  (VaCarousel docs fixes)


packages/docs/page-config/ui-elements/button/examples/WithIcon.vue
```diff
@@ -1,29 +1,25 @@
 <template>
-  <div class="flex items-center">
+  <div class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-4 gap-4">
```
m0ksem, line 2: Would be nice for these buttons not to be so wide.
image

Looks like grid solution doesn't work well on buttons. We need to wrap it with div, maybe.

packages/docs/page-config/ui-elements/button/examples/Tag.vue
```diff
@@ -1,12 +1,12 @@
 <template>
   <va-button
-    class="mr-6"
+    class="mr-6 mb-2"
```
m0ksem, line 3: Grid solution would be better, because it removes extra margin here:

image

But we still need to wrap button with div, so button looks smaller as they look right now.

packages/docs/page-config/ui-elements/button/examples/WithIcon.vue
m0ksem: Or we can set `max-width: max-content`.

m0ksem (changes requested):
Thanks! There are few small things to do.

We also need to remove Styles from sidebar in navigationRoutes.ts.
image

m0ksem:
Moved to: https://github.com/epicmaxco/vuestic-ui/pull/4113

Issue #3874, opened 2023-09-15
Date Picker disapear month/year and controls


zvenigorodskaia:
**Vuestic-ui version:** 1.8.0

**Steps to reproduce**
hover to the month and year in the Date Picker component or prev/next icons - they disappear
![image](https://github.com/epicmaxco/vuestic-ui/assets/35342240/8eee66b5-7d6a-4e61-9624-3494ae4d4abe)
https://ui.vuestic.dev/ui-elements/date-picker
![image](https://github.com/epicmaxco/vuestic-ui/assets/35342240/614b266c-a3d0-4029-84ba-49e95a873ac2)

PR #3877, opened 2023-09-15
feat(useElementBackground): optimized version

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3874


packages/ui/src/composables/useElementBackground.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  const c1 = Math.round((color1[0]) * (1 - weight) + (color2[0]) * weight)
+  const c2 = Math.round((color1[1]) * (1 - weight) + (color2[1]) * weight)
+  const c3 = Math.round((color1[2]) * (1 - weight) + (color2[2]) * weight)
+
+  return [c1, c2, c3, 1]
+}
+
+const getColorFromElements = (els: HTMLElement[]): RGBAColorParsed => {
+  const colors = els.map((el) => window.getComputedStyle(el).backgroundColor).reverse()
```
m0ksem, line 113: This can be moved to one reduce or even one for.

packages/ui/src/composables/useElementBackground.ts
```diff
@@ -0,0 +1,138 @@
...
+  el.addEventListener('transitionend', cb)
+
+  return () => {
+    el.style.transition = el.style.transition.replace(fakeBackgroundTransition, '')
+    el.removeEventListener('transitionend', cb)
+  }
+}
+
+const watchElementsBackground = (els: HTMLElement[], cb: () => void) => {
+  const unwatchers = els.map((el) => watchElementBackground(el, cb))
```
m0ksem, line 91: This can also be moved to one for.

packages/ui/src/composables/useElementBackground.stories.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+  },
+
+  setup () {
+    return {
+      color: useElementBackground(useCurrentElement()),
+    }
+  },
+
+  template: `
```
asvae, line 25: If I move buttons outside of block - "Current color" stops being reactive.

packages/ui/src/composables/useElementBackground.stories.ts
```diff
@@ -0,0 +1,183 @@
+import { defineComponent } from 'vue'
+import { useElementBackground } from './useElementBackground'
+import { useCurrentElement } from './useCurrentElement'
+
+export default {
+  title: 'composables/useElementBackground',
+}
+
+export const Default = () => ({
```
asvae, line 9: I suggest we add interaction tests here. Seems pretty easy to.

packages/ui/src/composables/useElementBackground.stories.ts
```diff
@@ -0,0 +1,198 @@
...
+  components: { VaButton, UseElementBackgroundDummy },
+  data () {
+    return {
+      background: '#000',
+      color: null,
+    }
+  },
+  template: `
+  <UseElementBackgroundDummy
```
asvae, line 23: I suggest we use "Dummy" component to test composable. I'm not sure how that might work for other cases, but connecting composable to story looks like a horrid mess to me, with no clear ins and outs. With separate component - it seems manageable.

packages/ui/src/composables/useElementBackground.stories.ts
```diff
@@ -0,0 +1,198 @@
...
+    :style="{ background }"
+    @update:color="color = $event"
+  />
+  <p>Color detected: <span data-testid="color">{{color}}</span></p>
```
asvae, line 27: So far we're having clear blocks for stories.

That is:
[ins]
[story target]
[outs]
[controls]

I suggest we stick to that structure.
I checked what storybook has to offer (i.e. if we can use their controls) - and I found nothing useful. If it's doable - it will be twice as complex and less flexible.

packages/ui/src/composables/useElementBackground.ts
```diff
@@ -0,0 +1,149 @@
+import { Ref, ref, watchEffect } from 'vue'
```
asvae, line 1: Would be nice to have story or unit test to support "optimized" claim. So that we can be sure it's fast enough.

m0ksem: How can we test if it optimized if everyone has a different PC specs?

I think we all just need to believe that this way is faster than parsing all of the CSS on a page and check if it has CSS variable and search for CSS variable again as it was previously.

asvae: @m0ksem CI has constant spec.

+ for optimization you probably mean something like 50 times faster? Not 5% faster right? So would be noticeable regardless of spec.

But I agree with your point. Let's not press that.

m0ksem: I'm not sure how much performant it is, but I just removed large operations, which was a bottleneck. Also, this solution is much more stable and have no edge cases (at least, I didn't find them)

packages/ui/src/composables/useElementBackground.stories.ts
m0ksem: Can you change it later? I didn't get it :_)

m0ksem:
Looks good for me. Great job, @m0ksem from two weeks ago!

PR #3889, opened 2023-09-19
fix(anchor-overscrolls): update css to fix anchor overscroll in main content 

### LINKED ISSUE #3878: Anchor overscrolls position
**Vuestic-ui version:** 1.8.0

If you use [anchored link](https://ui.vuestic.dev/ui-elements/date-input#validation)

You see this:
image

Instead you should see this:
image

ItsKDaniel:
## Description

closes #3878 

This PR sets overflow css to the main layout content area and updates scrollIntoView usage to enable anchor overscroll in main content.

## Types of changes


packages/ui/src/components/va-layout/VaLayout.vue
```diff
@@ -117,9 +117,11 @@ export default defineComponent({
   width: 100vw;
 
   &__area {
-    @include va-scroll();
+    @include va-scroll(var(--va-primary));
 
     &--content {
+      overflow-x: visible;
```
asvae, line 123: Spotted a regression:

# Before:
localhost_3000_ui-elements_scroll-container (5)



#After:
localhost_3000_ui-elements_scroll-container (2)

You can read more about this here (#3496), but it's critical for visual testing that we can screenshot the whole page.

asvae (changes requested):
Hey, thanks for you effort!

There is a critical regression in this PR so far.

packages/ui/src/components/va-layout/VaLayout.vue
ItsKDaniel: https://github.com/epicmaxco/vuestic-ui/pull/3889/files/4c3431edf2dcd09261e09dd1ded00e4e7078a0a1#diff-4dae880321fed108b33e744a82a182de9b33acc723adeb35c919f123a93c2b74R116

The height property in `.va-layout` class force the content to stay inside div. However, without overflow property in the `&--content` section, the anchor overscroll would not work as detailed in the issue. Setting the overflow property seems to cause the regression issue. 

To support both overscroll and fix the regression, it would be ideal to fix the layout's `height` prop. Any suggestions here @asvae ?

m0ksem: I fixed full page screenshot, but anchor scroll seems to be broken. I think we just need to add offset to scroll equal to navbar height.

m0ksem (approved):

PR #3890, opened 2023-09-19
Stories rework: VaButton

Roman4437:


packages/ui/src/components/va-button/VaButton.stories.ts
```diff
@@ -1,12 +1,367 @@
-import { defineComponent } from 'vue'
-import VaButton from './VaButton.demo.vue'
+import { VaButton } from './'
+import { userEvent } from '../../../.storybook/interaction-utils/userEvent'
+import { within } from '@storybook/testing-library'
+import { expect } from '@storybook/jest'
 
 export default {
```
asvae, line 6: On `yarn test:unit` I got an error:

image

Seems like it doesn't fallback to default or something?

packages/ui/src/components/va-button/VaButton.stories.ts
```diff
@@ -1,12 +1,367 @@
-import { defineComponent } from 'vue'
-import VaButton from './VaButton.demo.vue'
+import { VaButton } from './'
```
asvae, line 1: I added some changes with no-verify, but please make sure to fix tests before merge.

asvae (changes requested):
Pretty close to finish on stories. I'd like to get some output from Maksim on useSize before we continue though.

We also have some test regressions.

asvae:
@m0ksem do you have some feedback to share for [useSize](https://github.com/epicmaxco/vuestic-ui/pull/3890/files#diff-665efc0f1fd8eb784ffa29a47d2f27be9084d591f84ce0514f9543e6564a6b84)?

packages/ui/src/components/va-button/VaButton.stories.ts
```diff
@@ -1,12 +1,367 @@
...
+      Button
+    </VaButton>
+    [large]
+    <VaButton size="large">
+      Button
+    </VaButton>
+  `,
+})
+
+export const HoverBehavior = () => ({
```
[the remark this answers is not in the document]
Roman4437: I don't see a solution i could think of, let's handle it in a different pr

packages/ui/src/components/va-button/VaButton.stories.ts
```diff
@@ -1,12 +1,367 @@
...
+      Button
+    </VaButton>
+    [opacity]
+    <VaButton hoverBehavior="opacity">
+      Button
+    </VaButton>
+  `,
+})
+
+export const HoverMaskColor = () => ({
```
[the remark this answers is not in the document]
Roman4437: Same as HoverBehavior

packages/ui/src/components/va-button/VaButton.demo.vue
```diff
@@ -1,295 +0,0 @@
-<template>
-  <VbDemo>
-    <VbCard class="va-button-demo">
-      <table class="table table-bordered">
-        <tr>
-          <td>Types</td>
```
asvae, line 6: We need story for Presets. It should include all presets for button: 

default
primary
secondary
plain
plainOpacity

packages/ui/src/components/va-button/VaButton.demo.vue
```diff
@@ -1,295 +0,0 @@
...
-          <td>
-            <va-button @click="$refs.testButton.focus()">Focus Main Button</va-button>
-            <va-button ref="testButton">Main Button</va-button>
-            <va-button @click="$refs.testButton.blur()">Blur Main Button</va-button>
-          </td>
-        </tr>
-
-        <tr>
-          <td>
-            Multiline text
```
asvae, line 235: image

I added slightly longer text and it makes story broken.

packages/ui/src/components/va-button/VaButton.stories.ts
```diff
@@ -1,12 +1,433 @@
...
+    <VaButton loading>
+      Button
+    </VaButton>
+  `,
+})
+
+export const Size = () => ({
+  components: { VaButton },
+  template: `
+    [small]
```
asvae, line 161: image

Let's add vertical space so that it doesn't look as ugly
```
[small]
[button][button][button]
[medium]
[button][button][button]
etc
``

**Comment by asvae** [comment]
That would be the last round of feedback from me. We're good to merge 😎

File: packages/ui/src/components/va-button/VaButton.demo.vue (hunk 5 above)
Comment (asvae) [reply packages/ui/src/components/va-button/VaButton.demo.vue:6]: Moved some items here - https://github.com/epicmaxco/vuestic-ui/issues/3898. But that's not something we need to worry in context of this pr.

**Review by asvae — APPROVED** [review]
Looks good from my side, but needs review from @m0ksem

File: packages/ui/src/components/va-button/VaButton.vue [hunk 8]
```diff
@@ -307,13 +312,9 @@ export default defineComponent({
     line-height: var(--va-button-line-height);
     border-radius: var(--va-button-border-radius);
     letter-spacing: var(--va-button-letter-spacing);
-    min-height: var(--va-button-size);
-    min-width: var(--va-button-size);
 
     .va-button__content {
-      font-size: var(--va-button-font-size);
```
m0ksem, line 314: Remove CSS variables too.

m0ksem (approved):
I think we will polish sizes config later with @xx13 when he's gonna be ready.
So, you can change sizes as you want until it doesn't produce breaking changes.

packages/ui/src/components/va-button/VaButton.vue
asvae: Merging then 🤷

asvae (changes requested):
Everything button related is finished, but css variables removal is incomplete.

So I believe the best solution would be not to touch it at all for now.

So we need to rollback css changes.

PR #3891, opened 2023-09-20
fix(va-avatar-group): render avatar group with rest count based on max prop

### LINKED ISSUE #3879: VaAvatarGroup component is broken
**Vuestic-ui version:** 1.8.0

This is not the most used component (by far), but maybe it deserves some quality.

image

I added descriptions to storybook https://ui-storybook-develop.vuestic.dev/?path=/docs/vaavatargroup--docs.

So:
* we need to fix stories: you can check how that `max` prop should work generally.
* next - we need to all props and behavior information to docs. Probably we can add that to the same page (va-avatar-group is documented only in context of va-avatar page).

ItsKDaniel:
## Description

partial  #3879 

However this PR does not fix the storybook edits ( #3800 ) that described the issue

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-avatar-group/VaAvatarGroup.stories.ts
+++ b/packages/ui/src/components/va-avatar-group/VaAvatarGroup.stories.ts
@@ -28,11 +28,6 @@ export const Default = () => ({
     <va-avatar-group :options="avatars"/>
   `,
 })
-addText(
-  Default,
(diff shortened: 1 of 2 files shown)


asvae:
Thanks, I added a couple things, but overall looks good.

I won't close original issue because docs are not updated, but functionality part is fixed so that's on issues out.

Issue #3901, opened 2023-09-25
Input in Textarea can't inherit Textarea's height


chenjie199234:
**Vuestic-ui version:** 1.8.0
![1](https://github.com/epicmaxco/vuestic-ui/assets/29348971/8061d9a5-3d58-433e-9abe-d8be6bf9e288)
![2](https://github.com/epicmaxco/vuestic-ui/assets/29348971/982ffc6e-c1e3-4682-a9f7-81d5b9cdcd06)

PR #3902, opened 2023-09-25
fix(textarea): resize with height

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3901

Textarea:
image

Input:
image


asvae (approved):
Yep, input is stretched

Issue #2074, opened 2022-07-11
VaSwitch `@input` event doesn't work


asvae:
**Vuestic-ui version:** 1.4.6
We have it stated in the docs
<img width="851" alt="image" src="https://user-images.githubusercontent.com/5394573/178253860-e585a741-5261-49af-80fe-94479e14b881.png">

but I tried in demo to 

```
<va-switch
        v-model="value"
        @update:model-value="$vb.log('@update:model-value', value)"
        @input="$vb.log('input', $event)"
      />
```
and `@input` event was never getting triggered. It should return native event btw.

## Bonus points
* Consider adding change event (is it useful?). Maybe we only need `@change`?
----
So this issues is about native events. If switch doesn't have input event or something - you don't have to use it.

m0ksem:
Likely junk from vue2.

PR #3907, opened 2023-09-25
fix(#2074): VaSwitch made input clickable and fire change event

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/2074

### DIFF
--- a/packages/ui/src/components/va-switch/VaSwitch.stories.ts
+++ b/packages/ui/src/components/va-switch/VaSwitch.stories.ts
@@ -18,34 +18,39 @@ export const Default = () => defineComponent({
 export const ChangeEvent: StoryFn = () => ({
   components: { VaSwitch },
 
-  data: () => ({ value: false, changeCount: 0 }),
+  data: () => ({ value: false, changeCount: 0, lastEventValue: null }),
 
   methods: {
     onChange (e) {
-      if (e.target) {
-        this.changeCount++
-      }
+      this.changeCount++
+      this.lastEventValue = e.target.checked
     },
   },
 
   template: `
-    [changes count]: {{ changeCount }}
+    <p>change count: <span data-testid="count">{{ changeCount }}</span></p>
+    <p>last event value: <span data-testid="value">{{ lastEventValue }}</span></p>
     <VaSwitch v-model="value" @change="onChange" />
   `,
 })
 
-ChangeEvent.play = async (e) => {
-  const input = e.canvasElement.querySelector('input')!
+ChangeEvent.play = async ({ canvasElement }) => {
+  const input = canvasElement.querySelector('input')!
+  const changesCounter = canvasElement.querySelector('[data-testid="count"]')!
+  const changesValue = canvasElement.querySelector('[data-testid="value"]')!
 
   await userEvent.click(input)
 
-  expect(e.canvasElement.textContent).toContain('[changes count]: 1')
+  expect(changesCounter.textContent).toContain('1')
+  expect(changesValue.textContent).toContain('true')
 
   await userEvent.keyboard(' ')
 
-  expect(e.canvasElement.textContent).toContain('[changes count]: 2')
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
+  methods: {
+    onChange (e) {
+      if (e.target) {
+        this.changeCount++
+      }
+    },
+  },
+
+  template: `
+    [changes count]: {{ changeCount }}
```
asvae, line 32: small thing, but the convention is something like in Backtop > ClickEvent.

```html
<p>count: <span data-testid="count">{{ count }}</span></p>
```

```typescript
(canvas.getByTestId('clicked').innerText).toBe('2')
```

packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
+
+  await userEvent.click(input)
+
+  expect(e.canvasElement.textContent).toContain('[changes count]: 1')
+
+  await userEvent.keyboard(' ')
+
+  expect(e.canvasElement.textContent).toContain('[changes count]: 2')
+
+  await userEvent.type(input, '{enter}')
```
asvae, line 48: We don't need to test for enter.

packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
 })
+
+export const ChangeEvent: StoryFn = () => ({
+  components: { VaSwitch },
+
+  data: () => ({ value: false, changeCount: 0 }),
+
+  methods: {
+    onChange (e) {
+      if (e.target) {
```
asvae, line 25: Why do we need this check?

packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
+})
+
+ChangeEvent.play = async (e) => {
+  const input = e.canvasElement.querySelector('input')!
+
+  await userEvent.click(input)
+
+  expect(e.canvasElement.textContent).toContain('[changes count]: 1')
+
+  await userEvent.keyboard(' ')
```
asvae, line 44: Pretty sure this is handled by click event, no?

asvae (approved):
Functionality side is fine. I hope to see story improved a bit.

packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
   template: '<VaSwitch/>',
 })
+
+export const ChangeEvent: StoryFn = () => ({
+  components: { VaSwitch },
+
+  data: () => ({ value: false, changeCount: 0 }),
+
+  methods: {
+    onChange (e) {
```
asvae, line 24: Would be nice to check event value as well, not only count.

packages/ui/src/components/va-switch/VaSwitch.stories.ts
```diff
@@ -1,12 +1,51 @@
...
-  components: { VaSwitch },
+export const Default = () => defineComponent({
+  components: { VaSwitch: VaSwitchDemo },
   template: '<VaSwitch/>',
 })
+
+export const ChangeEvent: StoryFn = () => ({
+  components: { VaSwitch },
+
+  data: () => ({ value: false, changeCount: 0 }),
```
asvae, line 21: We don't strictly need value and v-model here.

packages/ui/src/components/va-switch/VaSwitch.stories.ts
m0ksem: querySelector seems much more clean solution for me, than within and getByTestId. I'd prefer querySelector.

PR #3911, opened 2023-09-27
feat(colors): add onColors support

m0ksem:
image

image


asvae (commented):
I'm pretty sure we can improve the quality of this.

I.e. better docs, stories, specs for edge cases. Is that something you would consider important, or you would rather merge it as is and deal with leftovers later?

(maybe you can formulate leftovers in different issue at least)

m0ksem:
I'd like to have a brief video guide on colors in general. Docs are fine and there is nothing to test, actually. There are 6 lines of this feature covered with dozens lines of docs and tests.

m0ksem:
Let's improve stories and this feature as needed. This is enough for now.

PR #3916, opened 2023-09-27
fix(web-components): styles and cache issue

m0ksem:
There was a regression after updating vue compiler, components: {} changed to resolveComponent, so we need a different mechanism for nested styles. Also, there was issue with unregistered cache plugin, we just mock it with empty object in web components, cuz we're not able to register any vue plugin here to store cache globally. Maybe we can move cache to window, but I don't care about it for now.

### DIFF
--- a/packages/sandbox/web-components/src/main.ts
+++ b/packages/sandbox/web-components/src/main.ts
@@ -1,8 +1,5 @@
-import { createApp } from 'vue';
 import { registerVuesticWebComponents } from 'vuestic-ui/web-components'
 import 'vuestic-ui/css'
-import App from './App.vue'
-import { createI18n } from 'vue-i18n'
 import demos from './demos'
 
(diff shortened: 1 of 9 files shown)


asvae (approved):
This PR I frankly don't understand a single bit 😓. And given it's not a biggest prio for us, probably it's good enough as long as it works for you.

I only strongly want to have an ability to know if this works. So I suggest to create an issue to cover that.

m0ksem:
> This PR I frankly don't understand a single bit 😓. And given it's not a biggest prio for us, probably it's good enough as long as it works for you.
> 
> I only strongly want to have an ability to know if this works. So I suggest to create an issue to cover that.

You can run sandbox to see before and later.

m0ksem:
Merging it because of https://github.com/epicmaxco/vuestic-ui/discussions/3915#discussioncomment-7125547

PR #3933, opened 2023-10-01
Inconsistent behavior for VaTimeInput right icon #3922

### LINKED ISSUE #3922: Inconsistent behavior for VaTimeInput right icon
**Vuestic-ui version:** 1.8.0

When you click on icon - dropdown doesn't close
When you click on input - dropdown closes

Icon should not have different behavior, it should behave as if it doesn't exist (I believe), and you only have input.

image

Compare to
https://www.naiveui.com/en-US/os-theme/components/time-picker

Udaay:
close #3922

## Description

Changed the onClick handler of `va-input-wrapper`

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-time-input/VaTimeInput.vue
+++ b/packages/ui/src/components/va-time-input/VaTimeInput.vue
@@ -11,7 +11,7 @@
         class="va-time-input__anchor"
         :style="cursorStyleComputed"
         v-bind="computedInputWrapperProps"
-        @click.stop="showDropdown"
+        @click.stop="toggleDropdown"
       >
         <template #default>
           <input
@@ -41,12 +41,9 @@
           />
           <va-icon
             v-if="$props.leftIcon"
-            class="va-time-input__left-button"
+            class="va-time-input__left-button va-time-input__side-button"
             :aria-label="tp($props.ariaToggleDropdownLabel)"
             v-bind="iconProps"
-            @click.stop="showDropdown"
-            @keydown.enter.stop="showDropdown"
-            @keydown.space.stop="showDropdown"
           />
         </template>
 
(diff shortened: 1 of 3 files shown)


Udaay:
#3933

asvae:
Hey. I was mistaken to reference naive ui.

Instead I should've mentioned vuestic-ui select. https://ui-storybook-develop.vuestic.dev/?path=/story/vaselect--old-demos

Naive does not close elements on click. But vuestic does. So we need to have toggle behavior for both icon and wrapper.

`pointer-events: none` on icon could do the trick :)

packages/ui/src/components/va-time-input/VaTimeInput.vue
```diff
@@ -407,4 +401,8 @@ export default defineComponent({
 
 <style lang="scss">
 @import "variables";
+
+.va-time-input__side-button {
+  pointer-events: var(--va-time-input-side-button-pointer-events);
```
m0ksem, line 406: Hey, @Udaay. We use CSS variables, so it is easier for user to change component look. Do you think there is a case when user need to change pointer-events behavior? 

The perfect examples of using css variables to configure border-radius, border-width, paddings, margins, etc.

What I don't like is to have css variables, that for 99% will not be changed. Like display (which is set to flex and whole component breaks if display is changed), position (which must be relative otherwise everything breaks).

We actually do not follow this rule completely, we have a lot of old stuff which was converted to css variables by a script. And I hope in future we will have less meaningless variables. 

If you find pointer-events css variable helpful here, let's have it, but if not please remove the css variable. Thanks.

packages/ui/src/components/va-time-input/_variables.scss
```diff
@@ -1,4 +1,5 @@
 :root,
 :host {
   --va-time-input-min-width: var(--va-form-element-min-width);
```
m0ksem, line 3: I'm also not sure if `--va-time-input-min-width` is used somewhere, so this can be removed too likely.

m0ksem (approved):
Looks good for me, let's deal with CSS variable and merge it.

packages/ui/src/components/va-time-input/VaTimeInput.vue
Udaay: Yes , I got your point and whole idea behind the css variables so that user can configure some specific styles.
Will remove this variable and add directly.

packages/ui/src/components/va-time-input/_variables.scss
Udaay: let me check, I will remove it if it doesn't required.

m0ksem (approved):
Thanks!

PR #3934, opened 2023-10-01
fix/3930 Vabutton now has padding with 2 icons as well as 1 (preset = primary)

### LINKED ISSUE #3930: VaButton with 2 icons should have same padding as with one icon
**Vuestic-ui version:** 1.8.0

https://ui-storybook-develop.vuestic.dev/?path=/story/vabutton--default

image

AdamVacha:
close #3930

## Description

After reviewing the code base and attempting to create JavaScript logic in order to allow padding for the presence of both icon and icon right inside of a Vabutton component, I realized a simpler fix. I read over the documentation and saw that the Vabutton presets prop was able to allow the button to have both icons with padding as long as the preset was set to "primary" instead of "plain". In the end, this is all I changed in the demo file. When creating future Vabuttons it could be noted that if the user wants to have 2 icons they must use the primary preset and not the plain preset.

## Markup:

<details>

```vue
// Your code
lines 175-177 in Vabutton.demo.vue

<va-button border-color="primary" preset="primary" icon="create" iconRight="clear">
              Secondary
            </va-button>
```
</details>

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/ui/src/components/va-button/VaButton.demo.vue
+++ b/packages/ui/src/components/va-button/VaButton.demo.vue
@@ -172,7 +172,7 @@
           <td>
             <va-button border-color="primary" preset="primary" icon="create">Secondary</va-button>
             <va-button border-color="primary" preset="secondary" iconRight="clear">Secondary</va-button>
-            <va-button border-color="primary" preset="plain" icon="create" iconRight="clear">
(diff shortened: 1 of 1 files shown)


asvae:
Thanks! Merging

PR #3938, opened 2023-10-03
Fix: Overflow behaviour of VaCollapse

### LINKED ISSUE #3931: VaCollapse hides overflow by default
**Vuestic-ui version:** 1.8.0

My guess is that collapse should always have scroll in this case. Otherwise that might be a surprise for user - at one element size - everything is working fine. At another - right part disappears into nowhere.

image

keerthanak-tw:
closes #3931 

Fix overflow behaviour of VaCollapse
## Description

The VaCollapse body wrapper had a overflow hidden which would have been causing the extra content to be clipped. Changed the overflow behaviour to auto so that users will get a scrollbar when the size exceeds more than allotted.
## Markup:

<details>

```vue
&__body-wrapper {
    transition: var(--va-collapse-body-wrapper-transition);
    overflow: scroll;   // Changed this to scroll from hidden

    &--bordered {
      border-bottom: 1px solid var(--va-background-border);

      .va-collapse--colored-header:not(.va-collapse--expanded) & {
        border-bottom: none;
      }

      .va-collapse--colored-body.va-collapse--expanded & {
        border-bottom: none;
      }
    }
  }
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-collapse/VaCollapse.vue
+++ b/packages/ui/src/components/va-collapse/VaCollapse.vue
@@ -224,7 +224,7 @@ export default defineComponent({
 
   &__body-wrapper {
     transition: var(--va-collapse-body-wrapper-transition);
-    overflow: scroll;
+    overflow: auto;
 
     &--bordered {
       border-bottom: 1px solid var(--va-background-border);
--- a/packages/ui/src/components/va-collapse/VaCollapse.vue
+++ b/packages/ui/src/components/va-collapse/VaCollapse.vue
@@ -57,6 +57,7 @@
         'va-collapse__body-wrapper--bordered': !$slots.body && !$slots.header,
       }"
       :style="contentStyle"
+      @transitionend="onTransitionEnd"
     >
       <div
         class="va-collapse__body"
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-collapse/VaCollapse.vue
```diff
@@ -224,7 +224,7 @@ export default defineComponent({
 
   &__body-wrapper {
     transition: var(--va-collapse-body-wrapper-transition);
-    overflow: hidden;
+    overflow: scroll;
```
m0ksem, line 227:
```suggestion
    overflow: auto;
```

> My guess is that collapse should always have scroll in this case

I think it meant that it must have scroll in case of overflow.

keerthanak-tw: That makes sense.. Applied the suggestion :)

m0ksem (approved):

m0ksem:
@asvae, I added new story about collapse overflow, if you think it needs some changes, please, make a separate issue.

PR #3943, opened 2023-10-04
Emit name of changed column in addition to the new sort order

### LINKED ISSUE #3940: Emit column name as well when sorting option changes
When using update:sortingOrder, we need to know column was just sorted. 

Currently there is no way to load a previously selected sort option as emit the sort direction but not the column name.
Update DataTableSortingOrder to maybe be called DataTableSortingOption and include another prop for the column name. When the sort changes, emit both the direction and column whose sort direction is being updated.
This would allow the ability to store sorting options in either pinia or a cookie to keep selected sorting option when returning to component.

**Describe alternatives you've considered**
I'm trying to come up with a workaround but I've yet to come up with something that works

This is a specific feature that our client has requested but I dont think we will be able to provide without these changes.

Bond-Addict:
closes #3940 

## Description

When you use @update:sortingOrder it will now emit the new order and the name of the column that is being sorted.
This would be a breaking change in the sense that it will pass back an object now, rather than just a single string value.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)
- [x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)


packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -129,6 +130,7 @@ export const useSortable = (
   // Sets the clicked heading's column as a one to sort by and toggles the sorting order from "asc" to "desc" to `null`
   // (un-sorted) if the same column is clicked again or sets sorting order to "asc" if some other column is chosen.
   function toggleSorting (column: DataTableColumnInternal) {
+    sortedColumn.value = column.key
```
m0ksem, line 133: I'd make another event and call it "columnSorted". @update:sortingOrder should be used only for updating sortingOrder prop.

packages/ui/src/components/va-data-table/hooks/useSortable.ts
```diff
@@ -64,8 +65,8 @@ export const useSortable = (
       if (props.sortingOrder === undefined) {
         sortingOrderFallback.value = value
       }
-
-      emit('update:sortingOrder', value)
+      const columnName = sortedColumn.value
+      emit('update:sortingOrder', { value, columnName })
```
m0ksem, line 69: In this example:
```ts
const sort = ref('desc')

v-model:sortingOrder="sort"
```
sort will be updated from 'desc' to { value: 'asc', columnName: 'user' }

m0ksem (changes requested):

m0ksem:
Made it as separate event.

m0ksem (approved):

packages/ui/src/components/va-data-table/VaDataTable.demo.vue
```diff
@@ -483,9 +482,9 @@
       <va-data-table :items="evenItemsShuffled" :columns="evenColumnsSortable" v-model:sort-by="sortByModel" v-model:sorting-order="sortingOrderModel" />
     </VbCard>
 
-    <VbCard title="Paginationser" class="demo">
```
Bond-Addict: Thanks for snagging the spelling changes. These were intentionally to try and figure out which demo was being used so I could watch for a console log and I was planning to revert them once I saw the event was emitting correctly.

m0ksem: I made new demo in strorybook. We're slowly moving from old demo format to new one, so I reverted previous changes, because we will probably remove these demos in future anyway.

Bond-Addict: Ah okay. How do I test demos in storybook?

m0ksem: run `yarn serve:storybook`

visit http://localhost:6006/?path=/story/vadatatable--column-sorted-event

Issue #3951, opened 2023-10-11
Input validation related issues


asvae:
**Vuestic-ui version:** 1.8.1

* Validation message should not override `messages` or we should have an option to manage that behavior.
  ![image](https://github.com/epicmaxco/vuestic-ui/assets/5394573/4707c8f7-e4de-4f62-82fb-a502795fe769)![image](https://github.com/epicmaxco/vuestic-ui/assets/5394573/6f317989-4481-4cc5-96e1-28f8024361fb)
  ```html
  <va-input
    v-model="formData.password"
    :rules="[(v) => !!v || 'Password field is required']"
    messages="Password should be 8+ characters: letters, numbers, and special characters."
  >
  ```
* Validation doesn't work for stateful va-input. It works if I add `v-model`. Should works the same in both cases.
  ```html
  <va-input
    stateful
    :rules="[(v) => !!v || 'Password field is required']"
    messages="Password should be 8+ characters: letters, numbers, and special characters."
  />
   ```
* Clicking on description or error of input triggers focus. It shouldn't.
  ![image](https://github.com/epicmaxco/vuestic-ui/assets/5394573/3ca56979-fb75-4ac9-8589-e408988d354b)

PR #3968, opened 2023-10-18
feat(forms): dirty state

### LINKED ISSUE #3636: Certain inputs don't cause unobtrusive validation
**Vuestic-ui version:** 1.7.3

**Steps to reproduce**
Can be seen when clicking stepper button.

When you click in the input and then click out, validation should be fired.
VaInput (Step 1) and VaDateInput (Step 2) do not fire validation at the very least.

VaSelect (Step 3) functions as expected still.

m0ksem:
Now, form controls have error state, but it is not visible if the component hasn't been interacted yet.
image

partial #3951,
close #3636 (now errors are not visible if component hasn't been interacted by user)


packages/ui/src/composables/useValidation.ts
```diff
@@ -142,28 +166,34 @@ export const useValidation = <V, P extends ExtractPropTypes<typeof useValidation
 
   watch(isFocused, (newVal) => !newVal && validate())
 
+  const immediateValidation = ref(props.immediateValidation)
```
m0ksem, line 169:
```suggestion
  const immediateValidation = toRef(props, 'immediateValidation')
```

Issue #3961, opened 2023-10-13
VaModal size doesn't work


asvae:
**Vuestic-ui version:** 1.8.1

I set size to large, yet it seems fully dependent on content size. You can try to set other sizes - modal would still look the same.

```html
<va-modal
  :modelValue="true"
  title="123"
  message="123"
  size="large"
/>
```

![image](https://github.com/epicmaxco/vuestic-ui/assets/5394573/a4c25090-e3bd-411d-a95d-f0107a9675f1)

## Other

We have [docs page where it looks ok](https://ui.vuestic.dev/ui-elements/modal#modal-sizing) for some reason. Not sure how this happened. Same for [story](https://ui-storybook.vuestic.dev/?path=/story/vamodal--default).

Could be some sort of confusion on what size means. I would expect modal to have constant size if I set it. But it seemingly behaves as if it's max-width.

m0ksem:
We'll need to update sizesConfig and change how this whole thing works, because right now you're not able to set specific width for modal.

Issue #3333, opened 2023-04-13
Rework sizes config


asvae:
Right now with sizes config:
* it's hard to modify globally
* for icons we use separate sizes config, that's not connected to global config
* it's hard to reason about, as it's different for different components

See #1671

We need a solution that solves these problems.

m0ksem:
Let's discuss with @xx13 sizes.

PR #3970, opened 2023-10-18
Make VaModal size prop work

xiongmao86:
Keeping size constant for short message.
Fixing #3961

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-modal/VaModal.vue
+++ b/packages/ui/src/components/va-modal/VaModal.vue
@@ -507,14 +507,16 @@ export default defineComponent({
   &--size {
     &-small {
       .va-modal__dialog {
-        max-width: map_get($grid-breakpoints, sm);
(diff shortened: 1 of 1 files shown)


m0ksem (approved):

EugenGedroyc:
In my opinion, the previous solution with just "max-width" was better. In my form an empty space appears on the right side, now  
You need a props for modal mode, like "fix" 

And what is reason to write:
```
max-width: 992px;
min-width: 992px;
``` 
If this code equivalently: ` width: 992px;` ?

xiongmao86:
@EugenGedroyc, sorry for the improper implementation. It's been a while, I don't remember exactly, but I think I might have try changing width, probably I have mistaken something, anyway sorry to hold the development a step back.

m0ksem:
We'll need a structured process about size. I want to deal with it here: https://github.com/epicmaxco/vuestic-ui/issues/3333

PR #3980, opened 2023-10-25
Fix VaInput range in slots example of slider docs

### LINKED ISSUE #3977: VaSlider: Limit circle draw range from 0 to 100
is it right?
slider

xiongmao86:
Set limit on the VaInput in order to limit sliding circle inside slider.
Fixes #3977

## Types of changes

### DIFF
--- a/packages/docs/page-config/ui-elements/slider/examples/Slots.vue
+++ b/packages/docs/page-config/ui-elements/slider/examples/Slots.vue
@@ -1,19 +1,21 @@
 <template>
   <va-slider v-model="value1">
     <template #prepend>
-      <va-input
-        v-model.number="value1"
-        class="w-[70px]"
-        type="number"
+      <va-counter
+        v-model="value1"
+        :min="0"
+        :max="100"
+        class="w-[110px]"
       />
     </template>
   </va-slider>
   <va-slider v-model="value2">
     <template #append>
-      <va-input
-        v-model.number="value2"
-        class="w-[70px]"
-        type="number"
+      <va-counter
+        v-model="value2"
+        :min="0"
+        :max="100"
+        class="w-[110px]"
       />
     </template>
   </va-slider>


packages/docs/page-config/ui-elements/slider/examples/Slots.vue
```diff
@@ -11,7 +12,8 @@
   <va-slider v-model="value2">
     <template #append>
       <va-input
-        v-model.number="value2"
+        v-model="value2Proxy"
+        @input="e => e.target.value = value2Proxy"
```
m0ksem, line 16: Maybe we should do it inside VaInput. I feel like it is not obvious to user (and for the one who wrote this doc example too). I'd expect v-model to be the same as input.value. What do you think?

m0ksem (approved):

packages/docs/page-config/ui-elements/slider/examples/Slots.vue
xiongmao86: Hi, @m0ksem, I agree with you, I will try to put it inside VaInput.

xiongmao86: Hi, @m0ksem, I think I can put the logic in `onInput` function in the `useCleave` hook, but I am not sure that is a good idea, but I can't find a way to put into the VaInput.vue. Can you give me some pointers?

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -164,6 +167,25 @@ export default defineComponent({
...
+        if (typeof min.value !== 'undefined' && value < min.value) {
+          valueComputed.value = min.value
+          return
+        } else if (typeof max.value !== 'undefined' && value > max.value) {
+          valueComputed.value = max.value
+          return
+        }
+      }
+      valueComputed.value = value
+    }
```
xiongmao86, line 188: Hi, @m0ksem, I don't know why this wouldn't work. I can't see this in vue devtool, doesn't show any prop or variable.

packages/docs/page-config/ui-elements/slider/examples/Slots.vue
m0ksem: I think you can manually add `@input` event to handle it separately from useCleave.

packages/ui/src/components/va-input/VaInput.vue
m0ksem: I think it is better to use `VaCounter` instead of `VaInput type="number"`

m0ksem (approved):

PR #3983, opened 2023-10-26
add alts to images

rustam-mukhametshin:
Add alts to images

## Types of changes

### DIFF
--- a/packages/docs/components/landing/Features.vue
+++ b/packages/docs/components/landing/Features.vue
@@ -29,6 +29,7 @@
           >
             <img
               :src="icon"
+              :alt="description"
               aria-hidden="true"
               class="features__item-icon"
             >
@@ -40,6 +41,7 @@
           <div class="features__item">
             <img
               src="/landing/features/features-support.svg"
+              alt="support"
               aria-hidden="true"
               class="features__item-icon"
             >
--- a/packages/docs/components/landing/Features.vue
+++ b/packages/docs/components/landing/Features.vue
@@ -29,8 +29,7 @@
           >
             <img
               :src="icon"
-              :alt="description"
-              aria-hidden="true"
+              alt=""
               class="features__item-icon"
             >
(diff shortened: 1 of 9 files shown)


packages/docs/components/landing/Features.vue
```diff
@@ -29,6 +29,7 @@
           >
             <img
               :src="icon"
+              :alt="description"
               aria-hidden="true"
```
m0ksem, line 33:
```suggestion
              alt=""
```

It is pure decorative image without any meaning. The same description text is displayed bellow in paragraph. For such images, it is better to have empty alt text (see https://www.w3.org/WAI/tutorials/images/decorative/) or any alt text (like "decorative image") with `aria-hidden`.

packages/docs/components/landing/Features.vue
```diff
@@ -40,6 +41,7 @@
           <div class="features__item">
             <img
               src="/landing/features/features-support.svg"
+              alt="support"
               aria-hidden="true"
```
m0ksem, line 45:
```suggestion
              alt=""
```

packages/docs/page-config/introduction/roadmap/images/1.6/autoComplete.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <img :src="Image"/>
+  <img :src="Image" alt="autoComplete"/>
```
m0ksem, line 2:
```suggestion
  <img :src="Image" alt="" />
```

It is also decorative image. Having extra alt text produces incorrect link text (easy to find in Firefox accessibility devtools):
image

Actually, all images on roadmap page (even svg) must have aria hidden or empty alt.

m0ksem (approved):
I think list can be generally improved and have this kind of description: 
image
But it would be nice to have in different issue, maybe requires VaList component update.

Issue #3990, opened 2023-11-01
VaAvatar: fallbackText must be used if src is empty string


m0ksem:
**Vuestic-ui version:** 1.8.3

**Steps to reproduce**

```html
        <va-avatar
          size="small"
          :src="user.avatar"
          :fallback-text="user.fullname[0]"
        />
```

or with values:
```html
        <va-avatar
          size="small"
          :src="''"
          :fallback-text="'M'"
        />
```

**What is the expected behavior?**

"M" must be shown if src is empty.

**What is the current behavior?**

Shows blue circle without text

**Other information**

xiongmao86:
Hi, @m0ksem, I can't reproduce the bug using your code sample. I put it in storybook and they all correctly show the initial character.

m0ksem:
@xiongmao86, I added reproduction story here: https://github.com/epicmaxco/vuestic-ui/pull/3998/files#diff-cb70e582c8c1dc5b32b1b9561f81772b282d037186f8a4f416d28647aa06fb09R126

Would you like to review PR?

xiongmao86:
Hi, @m0ksem, I would love to, how can I do that?

m0ksem:
> Hi, @m0ksem, I would love to, how can I do that?

You should be able have access to view File Changed in PR https://github.com/epicmaxco/vuestic-ui/pull/3998 and have a green button for review submission.

I'm using Github extension ([Marketplace](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github)) for pulling and reviewing PRs in VSCode locally. (Or you can use git CLI [guide here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally))

xiongmao86:
Thank you for your help, @m0ksem.

PR #3998, opened 2023-11-06
fix(#3990): use fallback if no src

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/3990

image

```ts
export const FallbackTextNoSrc = () => ({
  components: { VaAvatar },
  template: '<VaAvatar fallbackText="Text" />',
})
```


xiongmao86 (approved):
Great work. Thanks for helping me with reviewing, @m0ksem.

PR #4005, opened 2023-11-07
fix(#4001): checkbox margins in array-value mode

### LINKED ISSUE #4001: VaCheckbox: incorrect margins in array-value mode
**Vuestic-ui version:** 1.8.3

**Steps to reproduce**
https://www.chromatic.com/test?appId=64d608ab34b67e621ab50cfd&id=65491aac0cb292254de33c3e
Missing margins between checkboxes. Make sure it also present in vertical mode. Add array-value-vertical story. Notice that if horizontal only horizontal margins must be used. Same for vertical.
image
**Other information**

pradnyabhukan:
## Description

Added the missing margins between checkboxes in array-value story.
Added an array-value-vertical story, with the margins.

closes https://github.com/epicmaxco/vuestic-ui/issues/4001

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-checkbox/VaCheckbox.vue
+++ b/packages/ui/src/components/va-checkbox/VaCheckbox.vue
@@ -88,6 +88,7 @@ export default defineComponent({
     id: { type: String, default: '' },
     name: { type: String, default: '' },
     ariaLabel: { type: String, default: undefined },
+    vertical: { type: Boolean, default: false },
   },
 
   setup (props, { emit }) {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-checkbox/VaCheckbox.vue
```diff
@@ -128,9 +129,13 @@ export default defineComponent({
         color: computedError.value ? getColor('danger') : (props.success ? getColor('success') : ''),
         padding: !props.label
           ? ''
-          : props.leftLabel
-            ? '0 0.5rem 0 0'
-            : '0 0 0 0.5rem',
+          : props.vertical
+            ? '0.3rem 0 0.3rem 0.5rem'
```
m0ksem, line 133: We will need to move these values to CSS variables under `_variables.scss`, so they can be configured.

m0ksem (approved):

m0ksem:
Thanks!

PR #4006, opened 2023-11-07
fix: import nuxt composables from #imports

danielroe:
## Description
This is a DX improvement when developing - we can avoid loading the entire barrel file at `#app` by using the new granular imports merged in https://github.com/nuxt/nuxt/pull/23951.

## Types of changes


m0ksem (approved):
Thanks 🙏

Issue #3999, opened 2023-11-06
VaCounter: Readonly mustn't look like disabled


m0ksem:
**Vuestic-ui version:** 1.8.3

**Steps to reproduce**
https://www.chromatic.com/test?appId=64d608ab34b67e621ab50cfd&id=65491aac0cb292254de33c55
**What is the expected behavior?**
<img width="169" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/23530004/495bf445-5352-41e8-b7f3-690a2f5cd463">

**What is the current behavior?**
<img width="161" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/23530004/5dc2e21f-849d-46eb-a83d-7ce4197e9efb">

**Other information**

sfazli96:
Hi! If this is kind of like an CSS issue, I might give this a try. Let me know

m0ksem:
Hi, @sfazli96. It is more script logic. Looks like we make buttons disabled when counter is read-only, but actually, we just need to ignore click events instead.

sfazli96:
oh ok I see, I will still give this a try

jose-fernandez:
Hi guys! I would like to work on this. I don't know if I have to say it before hand so it gets assigned, or I can directly create the PR.

PR #4008, opened 2023-11-09
fix: va-counter isolate readonly attribute from disabled

jose-fernandez:
## Description
This is a PR to fix [#3999](https://github.com/epicmaxco/vuestic-ui/issues/3999)

This PR isolates `readonly` from `disabled` attribute, now the computed properties `isDecreaseActionDisabled` &  `isIncreaseActionDisabled` are only checking if disabled and min o max value is reached.

`readonly` is appended in `increaseIconProps` and `decreaseIconProps` directly from the props.

## Types of changes


m0ksem (approved):

m0ksem:
Thanks!

PR #4012, opened 2023-11-13
fix(#4011): checkbox Moved padding values from js to custom css variables

### LINKED ISSUE #4011: VaCheckbox: move padding values from JS to CSS
We will need to move these values to CSS variables under `_variables.scss`, so they can be configured.

_Originally posted by @m0ksem in https://github.com/epicmaxco/vuestic-ui/pull/4005#discussion_r1390609189_
            
            
Let's have something like `--va-checkbox-vertical-gap` and `--va-checkbox-horizontal-gap`

Abhid1239:
## Description

Moved css padding values from js to custom css variables.
Added a switch case for better readability

closes #4011  

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-checkbox/_variables.scss
+++ b/packages/ui/src/components/va-checkbox/_variables.scss
@@ -17,4 +17,10 @@
   --va-checkbox-square-height: 1.25rem;
   --va-checkbox-square-border: solid var(--va-form-element-border-width) var(--va-background-border);
   --va-checkbox-square-border-radius: calc(var(--va-form-element-border-radius) / 2);
+
+  /* Padding */
+  --va-checkbox-vertical-padding: 0.3rem 0 0.3rem 0.5rem;
+  --va-checkbox-horizontal-padding: 0 0.5rem;
+  --va-checkbox-right-padding: 0 0.5rem 0 0;
+  --va-checkbox-left-padding: 0 0 0 0.5rem;
 }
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-checkbox/_variables.scss
```diff
@@ -17,4 +17,10 @@
   --va-checkbox-square-height: 1.25rem;
   --va-checkbox-square-border: solid var(--va-form-element-border-width) var(--va-background-border);
   --va-checkbox-square-border-radius: calc(var(--va-form-element-border-radius) / 2);
+
+  /* Padding */
+  --va-checkbox-vertical-padding: 0.3rem 0 0.3rem 0.5rem;
+  --va-checkbox-horizontal-padding: 0 0.5rem;
+  --va-checkbox-right-padding: 0 0.5rem 0 0;
+  --va-checkbox-left-padding: 0 0 0 0.5rem;
```
m0ksem, line 25: Can we have something like this? I'm not sure how to name 0.3rem though. Do you have any ideas?

Maybe vertical-gap and horizontal-gap.

```suggestion
  --va-checkbox-gap: 0.5rem;
  --va-checkbox-vertical-padding: 0.3rem 0 0.3rem var(--va-checkbox-gap);
  --va-checkbox-horizontal-padding: 0 var(--va-checkbox-gap);
  --va-checkbox-right-padding: 0 var(--va-checkbox-gap) 0 0;
  --va-checkbox-left-padding: 0 0 0 var(--va-checkbox-gap);
```

Abhid1239: I would be much better if we were applying only paddingLeft and paddingRight instead of making them 0px every time this with that we could have directly give `vertical-gap` and `horizontal-gap` as values. But they might be required to overwrite anything.

So,  let's just go with `horizontal-gap` and `vertical gap`

Issue #4003, opened 2023-11-07
VaInnerLoading: add loading icon slot


m0ksem:
**Is your feature request related to a problem? Please describe.**

For innerLoading I wanted to make a timer from 3 to 1, but in VaInnerLoading it is impossible to provide custom slot and render numbers instead of spinner.

**Describe the solution you'd like**

**Describe alternatives you've considered**
I can do it manually, but should I?

**Additional context**

This will allow users to use custom inner loading content, like cards or VaProgressBar instead of circle. 
(Make sure slot will not spin!)

jose-fernandez:
@m0ksem Here is the PR. I added the slot and an example in storybook using `VaProgressBar`. I hope you did not mean literally to create the countdown timer xD

emperorsegxy:
@m0ksem , @jose-fernandez , since this issue has been addressed, how do we close it?

PR #4016, opened 2023-11-16
fix(#4003): add custom loader component via slot

jose-fernandez:
## Description
This PR [#4003](https://github.com/epicmaxco/vuestic-ui/issues/4003) adds custom loader components on `va-inner-loading` via slot.

Also it adds a new example in storybook for custom loader. `Default` and `custom` loaders share the same loading trigger button. 

## Extra Notes
The name of the slot is `loading` as `vaInfiniteScroll` also contains a loader slot and it is named `loading`. For consistency, I´ve chosen the same name. Personally, I think a more explicit name would be `loader` as it is a loader component what we will pass through the slot. I would use `loading` more for state cases.

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-inner-loading/VaInnerLoading.demo.vue
+++ b/packages/ui/src/components/va-inner-loading/VaInnerLoading.demo.vue
@@ -1,14 +1,13 @@
 <template>
   <vb-demo>
-    <vb-card width="400px">
-      <va-button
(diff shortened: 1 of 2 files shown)


m0ksem (approved):
I also like `loader` more, sadly we choose a different one in different component. Good noticed. Great job, thanks!

Issue #4019, opened 2023-11-21
[build] vuestic-ui/dist/vuestic-ui.css is unusable with esm build


m0ksem:
In esm build we have plugin which changes CSS variables behavior and uses variables which starts with --va-[index]-[js-variable-name], but in vuestic-ui.css we use vue syntax (--[hash]). Because of that variables are inconsistent between CJS and ESM builds.

Issue #4018, opened 2023-11-17
Sandbox is broken


suntehnik:
**Vuestic-ui version:** 1.8.4

**Steps to reproduce**
Open sandbox from components with "Open in Sandbox" link

**What is the expected behavior?**
Working sandbox environment

**What is the current behavior?**
Sanbox terminal showing error
`
$ vite
/bin/sh: 1: vite: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
`

**Other information**

m0ksem:
I want to introduce a new sandbox using vue repl (similar to https://play.vuejs.org/), so it is easier to play with components in it. Likely we gonna replace Codesandbox with it, cuz Codesandbox provides a number of features that overloaded.

Issue #3702, opened 2023-08-16
Codesandbox examples do not seem to be working


asvae:
**Vuestic-ui version:** 1.7.4

<img width="881" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/5394573/9301e769-5eb1-4faf-a810-6d5ee49aa30d">

m0ksem:
Transforming codesandbox to new editor fixes it for me.

PR #4021, opened 2023-11-21
docs: playground

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/4018
closes https://github.com/epicmaxco/vuestic-ui/issues/3702
closes https://github.com/epicmaxco/vuestic-ui/issues/3162

Removed broken Codesandbox and added playground on ui.vuestic.dev.

New button: (not sure about icon tho)
image

Playground for component example looks like a modal:
> http://localhost:3000/ui-elements/button
image

After user changed something in playground, it can be shared via url. Url looks like:
> http://localhost:3000/play
image

---

I use vue-repl here to make a sandbox. We need a complete ES build in one file for vue-repl, so I build one in nuxt module before docs build start. Vue-repl makes it easy to share state between pages using serialization. So we don't need to store user's changes on backend - it is stored in url. 
When user open playground from code example in modal url changes to ui.vuestic.dev/play#state so it can be easily shared. When modal is closed, url changes back to docs page.

I believe this will help users to make reproduction easier.

Sandbox supports tailwind, but will not allow installing additional dependencies, what I think OK.

---

This PR also closes https://github.com/epicmaxco/vuestic-ui/issues/4019, I found out this problem when tried to use our es module from CDN.

---

Downside of this PR is that for now, it is impossible to test not latest vuestic version. In such case, I think user still able to make codesandbox reproduction.


Roman4437:
There's an overflow issue 
image

Roman4437:
during focus or selection, a default outline appears

Roman4437:
The background is static and doesn't change with the theme. Also, white background doesn't look good with elements that have dropdowns

Roman4437:
Some components are not working properly, such as the date-picker or table

Roman4437:
tailwind classes works unless they are custom or not defined through the docs

packages/docs/.gitignore
```diff
@@ -7,3 +7,4 @@ node_modules
 .env
```
asvae, line 7: Got an error on `serve:docs`.

On `develop` docs works fine.

image

m0ksem: Updating deps and clearing cache might help.

Roman4437: yes, i had the same error and have fixed by clearing cookies

m0ksem:
> tailwind classes works unless they are custom or not defined through the docs

I updated PR, might be fixed. Can you share an example where the issue happens? Thanks.

Roman4437:
> > tailwind classes works unless they are custom or not defined through the docs
> 
> I updated PR, might be fixed. Can you share an example where the issue happens? Thanks.

It's working properly now

packages/docs/pages/play.vue
```diff
@@ -0,0 +1,82 @@
...
+      <Play
+        v-model:state="sandboxState"
+        class="h-[100%] flex-1"
+        :code="makeCode(scriptText, templateText, '')"
+      />
+    </ClientOnly>
+
+    <div class="flex justify-between items-center py-2 px-4 border-t-2 border-[var(--va-background-border)] bg-[var(--va-background-secondary)]">
+      <div>
+        Playground <span class="text-[var(--va-secondary)]">vuestic-ui@latest - vue@latest</span>
```
m0ksem, line 70:
```suggestion
        Playground <span class="text-[var(--va-secondary)]">vuestic-ui@latest - vue@latest - tailwindcss@latest</span>
```

m0ksem:
Looks like `<style lang="*"` isn't supported yet by vue-repl.
image


We also need to add config, similar to codesandbox's config, so it can load additional libs, like AGGrid.

packages/docs/components/Play.client.vue
asvae, line 1: Mobile version for modal doesn't work too well (full-screen works fine).

image

I'm thinking we need 2 buttons here:
image copy
image close

Sandbox link I don't think is useful for anything.
I would also suggest not to use primary color for close.

image

asvae:
For Asva to finish:

Master → 1.8.4

Develop → compiled (npm)

Experimental → compiled (npm)

Master → [1.8.3, 1.8.4]

Develop → [1.8.3, 1.8.4, 1234567a, 123567b]

play - source code

play - vuestic-version

m0ksem:
> For Asva to finish:
> 
> Master → 1.8.4
> 
> Develop → compiled (npm)
> 
> Experimental → compiled (npm)
> 
> Master → [1.8.3, 1.8.4]
> 
> Develop → [1.8.3, 1.8.4, 1234567a, 123567b]
> 
> play - source code
> 
> play - vuestic-version
> 
> * Some examples still use old codesandbox - we need to cover that.

I have a PR for build tool. Let's deal with it later, when my PR is merged. It will be easy to add a esm-browser build, which we could use from npm.

packages/docs/components/Play.client.vue
m0ksem: I think it is a problem caused by VaModal... Moved that to @VasylHryha.

asvae:
I moved everything we discussed to [this issue](https://github.com/epicmaxco/vuestic-ui/issues/4078). So we can merge this PR.

PR #4026, opened 2023-11-22
Changed Toast init to notify

### LINKED ISSUE #4024: Let's add alias to useToast init function
Name `init` looks too general. I'd prefer something like:

```ts
notify("User has been deleted!")
```

or

```ts
makeToast("User has been deleted!")
```

- [ ] Come up with new name which is super clear
- [ ] Add alias to useToast composable (init should stay, but you can mark it `@deprecated` using JSDocs
- [ ] Change `init` in docs to new method name.

https://ui.vuestic.dev/ui-elements/toast#toast

joshLumpkin:
Fixes #4024

## Description
added an alias for init, in the toast feature. you can now use notify instead of init. Init is still retained and marked with depreciated. 

Also updated docs to use the new notify instead of init.

## Types of changes

- [x ] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)


packages/ui/src/components/va-toast/hooks/useToast.ts
```diff
@@ -18,6 +19,10 @@ export const useToast = () => {
     return id
   }
 
+  const notify = (options: string | NotificationOptions) => {
```
asvae, line 22: It probably should be another way round.

init should call notify :)

so that when notify is used - no deprecation warnings will happen

and when init is used - deprecation warning would happen

asvae (changes requested):
Hey, thanks!

One small thing I mentioned and we can merge!

packages/ui/src/components/va-toast/hooks/useToast.ts
joshLumpkin: ah, good point. updating PR now

joshLumpkin:
Requested review, let me know if any other changes are needed! Trying to contribute more frequently!

m0ksem (approved):
I'm okay with JSDocs deprecation warning instead of console.warn, we can forget about init method for few minor versions.

PR #4032, opened 2023-11-24
Stories rework: VaIcon

Roman4437:


packages/ui/src/components/va-icon/VaIcon.stories.ts
```diff
@@ -1,12 +1,114 @@
...
   component: VaIcon,
+  tags: ['autodocs'],
 }
 
-export const Default = defineComponent({
+export const Default = () => ({
   components: { VaIcon },
-  template: '<VaIcon/>',
+  template: `
+    <div class="flex flex-col">
```
asvae, line 13: Default should be just `<VaIcon name="book">`.

Let's move font-awesome / ionic to "icon fonts" story

asvae (approved):
Found one small issue, apart from that - it's perfect. 👍

PR #4044, opened 2023-11-28
Stories rework: VaDropdown

Fsss126:
See #3676.

## Types of changes


packages/ui/.storybook/storybook-main.scss
```diff
@@ -6,4 +6,5 @@
 //@import "../components/vuestic-sass/global/normalize.scss";
 @import './assets/icons/icons.scss';
 @import './assets/fonts.scss';
+@import './assets/global';
```
m0ksem, line 9: Can we prefer tailwind? We will forget about this approach and external contributors will use a different way.

@asvae, correct me if we're not allowed to use tailwind in stories.

packages/ui/src/components/va-aspect-ratio/VaAspectRatio.stories.ts
```diff
@@ -9,8 +9,8 @@ export default {
 export const Default = () => ({
   components: { VaAspectRatio },
   template: `
-    <va-aspect-ratio 
-      :style="{ width: '200px', border: '1px solid black' }"
+    <va-aspect-ratio
+      :style="{ width: '200px', border: 'var(--border-dashed)' }"
```
m0ksem, line 13:
```suggestion
      class="w-48 border-dashed border-2"
```

m0ksem (changes requested):
Would be nice to cover stories with tests.

packages/ui/.storybook/storybook-main.scss
Fsss126: I removed tailwind helpers that were in the original stories as it was stated as one of the refactoring steps in #3676

> Try not to use tailwind helpers - we want to remove it from storybook eventually.

@asvae, need your say here on what is the latest agreement on this

asvae: Tailwind helpers are all right. One concern might be not overusing them.

I.e. where it's enough to use only one helper - don't use 3.

And where it's enough to use `<p>` - don't use helpers at all.

m0ksem: I changed guide a bit to
> Please use tailwind for helpers, but not overuse them. Minimum amount of styling is appreciated.

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+})
+
+export const AnchorSelector = () => ({
+  components: { VaDropdown },
+  template: `
+      <div
+          style="border: var(--border-dashed)"
+          id="anchor-selector"
+      >
+        Click me
```
asvae, line 225:
```suggestion
        Anchor
```

asvae: Try to use as little variety as possible.

I.e. anchor should always be a div with "Anchor" text if there is no good reason to use anything else. Should also use the same styling for all cases.

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+      </div>
+      <va-dropdown
+          anchor-selector="#anchor-selector"
+      >
+        Anchor selector
```
asvae, line 230: Here it should be Dropdown content or Content in all cases.

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+})
+
+export const InnerAnchorSelector = () => ({
   components: { VaDropdown },
-  template: '<VaDropdown/>',
+  template: `
+      <va-dropdown
+          inner-anchor-selector="#inner-anchor-selector"
+      >
+        <template #anchor>
```
asvae, line 241: Need a test case showing that you can only trigger via inner selector and can't trigger by clicking on anything else in #anchor slot.

Might make sense to have 2 divs in slot:
[anchor]
[not anchor]

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+              Click me
+            </div>
+          </div>
+        </template>
+        Inner anchor selector
+      </va-dropdown>
+    `,
+})
+
+export const AnchorBoundProps = () => ({
```
asvae, line 256:
```suggestion
export const AnchorSlotProps = () => ({
```

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+    `,
+})
+
+export const KeepHeight = () => ({
+  components: { VaDropdown, VaButton, VaDropdownContent },
+  template: `
+      <div style="height: 1000px">
+        <va-dropdown>
+          <template #anchor>
+            <va-button>Anchor</va-button>
```
asvae, line 276: let's either use va-button everywhere or nowhere

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+          <div>
+            <va-icon :name="isOpened ? 'va-arrow-up' : 'va-arrow-down'"/>
+          </div>
+        </template>
+        Clicked
+      </va-dropdown>
+    `,
+})
+
+export const KeepHeight = () => ({
```
asvae, line 270: Really unclear what this example shows (I have no idea 🤷).

If we can showcase that via interaction or naming - we should.

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+              </div>
+            </template>
+            Dropdown
+          </va-dropdown>
+        </div>
+      </scroll-container>
+    `,
+})
+
+export const AutoplacementWithHeight: StoryFn = () => ({
```
asvae, line 319:
```suggestion
export const AutoplacementWithScroll: StoryFn = () => ({
```

Probably could be both horizontal and vertical

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+            <div style="height: 600px; width: 100px;">
+              Hi
+            </div>
+          </va-dropdown-content>
+        </va-dropdown>
+      </div>
+    `,
+})
+
+export const Autoplacement: StoryFn = () => ({
```
asvae, line 288: Need an example for default autoplacement behavior.

It's not clear to me how using `autoplacementTargetRef` differs from not using it.

asvae (changes requested):
Great work!

Left some story suggestions. We can merge after fix!

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+      <va-dropdown
+          keepAnchorWidth
+      >
+        <template #anchor>
+          <div>
+            Short anchor
+          </div>
+        </template>
+        <va-dropdown-content>
+          Loooooooong Dropdown
```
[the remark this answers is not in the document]
Fsss126: As I got it, the point was not to test scroll behaviour but to showcase that it matches it in both of these two cases and having both corner cases will help to catch a bug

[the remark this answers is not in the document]
asvae: I would consider 
* modal scales with anchor
* and modal has scroll

to be different cases, doesn't seem likely for me to have an edge case that involves 2 of these.

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
Fsss126: Default autoplacement behaviour without target doesn't seem to handle overflow properly and dropdown gets positioned outside of scroll parent. Is this needed to be showcased? I thought that in this case target needs to be specified for everything to work correctly and I doubt that improper usage should be shown

Fsss126: Or is current behaviour without specifying the target considered a bug?

packages/ui/.storybook/storybook-main.scss
Fsss126: Returned tailwind helpers 🙌

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+      >
+        <template #anchor>
+          Hover me
+        </template>
+        Hovered
+      </va-dropdown>
+    `,
+})
+
+export const PlacementTop = () => ({
```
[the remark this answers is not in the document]
Fsss126: Done

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+        </va-dropdown-content>
+      </va-dropdown>
+    `,
+})
+
+export const CustomWidth = () => ({
+  components: { VaDropdown, VaValue, VaDropdownContent },
+  template: `
+      <VaValue #default="v" :default-value="true">
+        <button @click="v.value = !v.value">{{ v.value ? 'Short' : 'Long' }}</button>
```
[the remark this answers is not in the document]
Fsss126: Changed it

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
```diff
@@ -1,12 +1,386 @@
...
+            <va-dropdown-content :style="{ minWidth: getAnchorWidth() }">
+              {{ v.value ? 'Short' : 'Loooong drooooopdown wwiiiiiidth' }}
+            </va-dropdown-content>
+          </template>
+        </va-dropdown>
+      </VaValue>
+    `,
+})
+
+export const Cursor = () => ({
```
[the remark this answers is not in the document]
Fsss126: Removed trigger

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
Fsss126: Renamed it

packages/ui/src/components/va-dropdown/VaDropdown.stories.ts
Fsss126: Left it verrical as horizontal scroll is showcased in story demonstrating sticking to edges behaviour

asvae (approved):
Great work, thanks!

Issue #4048, opened 2023-11-29
VaDataTable: make a way to disable sorting on client


m0ksem:
When I make server-side sorting, DataTable makes client-size sorting at first with animation, then items are updated and animation happens second time. 

We need something like `client-sorting="false"` prop

PR #4050, opened 2023-11-29
fix(data-table#4048): disable-client-side-sorting prop

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/4048


jordin-marshall:
@m0ksem sorting still doesn't currently work when adding this new prop.

PR #4057, opened 2023-12-01
fix(stateful): use provided prop instead of default value if provided.

m0ksem:
> I see how I can create form data with a new form, but how would I prepopulate data? 
> 
> I cant do myForm.formData.value = {} as SomeType, because .value is readonly. Is formData only able to used to create records and not edit them? I was wondering about this approach because I'm experiencing issues with clearing a form by setting the v-model object to empty or null which immediately fires the form validation. This happens wether or not I call reset() or try and use resetValidation() to clear messages.

Fix this example: https://ui-experimental.vuestic.dev/play#eNp9Ustu2zAQ/BVCFydAZAHJzVACNE0CpGjdoi1y4oWWVzYTvsCHY8PQv2dJWpLzgC+SODszO1ztvvhmzHQToJgVtWssN5448MEQwdTqmhbe0eKGKkK4NNp6sifBwYO2knSktVqSCYqd500Z+ISqyGy0cpHYIuuOeYbM6151NpG7+J6cU1VXuSHa48GDNIJ5SM3qJ5Z6WGgxQ5bQgjiPhDYI0gjmXKwsykvSCtimR9loQVbMlJeHzMnoUZngidRLEE9MBEDZL/bidtFQMRnPLbfOz/EbIcEWIBB7iBg5gNVJuzkstVvDZjTEeB/9fiL0hR0OCLLlLHmWm4OpglcSi2fno+2CW7++Y7sj29sIkWXCsm1d5eGlseIxiL6b4Df5VtFuRvbjL5oOEyBdV1dIHCUp+GdFf8XPgiHSe0Ef/lhQVyldXR39faqKiyJvWymZmT47rXA795FPDwVcSjTPLWkxbmCEabH23rhZVQVewtaA5RKUZyKueeRNl7Cpeo0OvpKMK2xDi2jYUdVhAO9wjVu++tC+0dJwAfa38RzX/F0MJoR+/ZEwbwNc9HizhublC/zZbXPePxYc2A1uxlDzzK7A5/L9vzls8Xso4qIEgewTxb/gtAgxY6bdBrXE2Ee8lPYxTZOr1X93v/WgXH+pGDRNI/HTiL+fuPoY92p6NUyxewPj83Vs

### DIFF
--- a/packages/ui/src/components/va-carousel/VaCarousel.vue
+++ b/packages/ui/src/components/va-carousel/VaCarousel.vue
@@ -166,7 +166,7 @@ export default defineComponent({
   emits: [...useStatefulEmits],
 
   setup (props, { emit }) {
-    const { valueComputed: currentSlide } = useStateful(props, emit, 'modelValue', { defaultValue: 0 })
+    const { valueComputed: currentSlide } = useStateful(props, emit, 'modelValue')
 
     const {
       goTo, next, prev,
--- a/packages/ui/src/components/va-form/VaForm.stories.ts
+++ b/packages/ui/src/components/va-form/VaForm.stories.ts
@@ -23,6 +23,7 @@ import {
   VaRadio,
 } from '../'
 import { sleep } from '../../utils/sleep'
+import { useForm } from '../../composables'
 
 export default {
   title: 'VaForm',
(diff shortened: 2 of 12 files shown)


m0ksem:
Looks good for me after week + we tested it with experimental version.

PR #4084, opened 2023-12-11
Replace google analytics with gtm

asvae:
Made it so that we don't have to do any codebase changes when we try to experiment with analytics or add another service.


asvae:
I'll merge after I test it works properly.

asvae:
Tested and works.

Issue #3848, opened 2023-09-09
Update google fonts to material symbols


m0ksem:
Must be updated in packages:
docs
create-vuestic
nuxt

asvae:
@m0ksem fine to assign this to @Stars22?

PR #4085, opened 2023-12-12
- feat: update google fonts to material symbols

VasylHryha:
Close https://github.com/epicmaxco/vuestic-ui/issues/3848
</details>

## Types of changes

### DIFF
--- a/packages/docs/nuxt.config.ts
+++ b/packages/docs/nuxt.config.ts
@@ -107,6 +107,7 @@ export default defineNuxtConfig({
       },
       'Material+Icons': true,
       'Material+Icons+Outlined': true,
+      'Material+Symbols+Outlined': true,
     },
   },
 


packages/docs/nuxt.config.ts
```diff
@@ -107,6 +107,7 @@ export default defineNuxtConfig({
       },
       'Material+Icons': true,
       'Material+Icons+Outlined': true,
```
m0ksem, line 109: Might worth it to remove old Material Icons then. And check docs for regression.

packages/docs/nuxt.config.ts
```diff
@@ -105,8 +105,7 @@ export default defineNuxtConfig({
       'Source+Code+Pro': {
         wght: [400],
       },
-      'Material+Icons': true,
-      'Material+Icons+Outlined': true,
```
VasylHryha, line 109: Did a quick check with docs, look fine

packages/docs/nuxt.config.ts
VasylHryha: removed

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
```diff
@@ -41,7 +41,8 @@ export const addVuesticToVue3App = async () => {
   let htmlSource = await readFile(htmlPath, 'utf-8')
   htmlSource = insertHead(htmlSource, [
     `<link href="https://fonts.googleapis.com/css2?family=Source+Code+Pro&family=Source+Sans+Pro:wght@400;700&display=swap" rel="stylesheet">`,
-    `<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">`
+    `<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">`,
```
asvae, line 44: Any reason we need material icons?

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
```diff
@@ -41,7 +41,7 @@ export const addVuesticToVue3App = async () => {
   let htmlSource = await readFile(htmlPath, 'utf-8')
   htmlSource = insertHead(htmlSource, [
     `<link href="https://fonts.googleapis.com/css2?family=Source+Code+Pro&family=Source+Sans+Pro:wght@400;700&display=swap" rel="stylesheet">`,
-    `<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">`
+    `<link href="https://fonts.googleapis.com/icon?family=Material+Symbols+Outlined" rel="stylesheet">`
```
m0ksem, line 44: We need to update IconsConfig to use material-symbols but default, instead of material icons.
I expect updating nuxt module in my project and all icons still work, even if the're symbols now.

Right now, creating new vuestic project will install Material Symbols, but we use material-icons in IconsConfig...

m0ksem (changes requested):

VasylHryha:
Replaced `Material Icons` -> `Material Symbols Outlined` almost everywhere
`Docs` looks fine

Please check to make sure I haven't replaced icons where I shouldn't have, or if there are places where we still need `Material Icons`.

m0ksem:
I think we don't need Material Icons anymore anywhere. We just need to make sure we use symbols everywhere. The next release is gonna be 1.9.0 where we can introduce breaking changes.

packages/create-vuestic/src/steps/2.addVuestic/vue-app.ts
asvae: We should release both projects simultaneously then.

PR #4089, opened 2023-12-12
fix: normalize return in setup function

m0ksem:
Make setup function return { [variableName] }

I need this for script to convert everything into setup script.


asvae (approved):
LGTM! Thanks

PR #4094, opened 2023-12-14
delete mobileFullscreen prop

VasylHryha:
## Description

I propose removing the `mobileFullscreen` feature, as it leads to unclear behavior. In examining other UI frameworks like Vuetifyjs, Native UI, and Quasar, I found that they typically adjust modals to fit the window width rather than using a `fullscreen` mode on mobile. I've implemented this behaviour here: https://github.com/epicmaxco/vuestic-ui/pull/4093. As an alternative to `mobileFullscreen`, I recommend using the `fullscreen` prop, which offers clearer and more consistent functionality.

## Types of changes


VasylHryha:
[Yauheni Prakopchyk](https://github.com/asvae) approved this change

PR #4097, opened 2023-12-16
fix(VaSelect #4064): option slot

### LINKED ISSUE #4064: VaSelect: hard to customize select options
Example: 
image

We should have one slot to customize option:
image

And it also should reflect how option is displayed when selected in input field.
For some reason, user need to manually add @click="selectOption". I'd expect just change visual and option to work, but there is nothing happening on click without adding listener manually.

We're also losing default option styles. 

I think we need something like `option-content` slot, which will be placed in SelectOption instead of text. So there is working keyboard navigation, hover effect, etc. But option can have additional icons, avatar, etc. 

Right now, using `option` slot breaks keyboard navigation completely.

Roman4437:
closes #4064

### DIFF
--- a/packages/ui/src/components/va-select/VaSelect.stories.ts
+++ b/packages/ui/src/components/va-select/VaSelect.stories.ts
@@ -145,3 +145,19 @@ export const AutocompleteMultiple: StoryFn = () => ({
 
   template: '<VaSelect v-model="value" :options="options" autocomplete multiple />',
 })
+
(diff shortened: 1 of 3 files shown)


m0ksem:
Would be nice to have something for both option in dropdown and when it is selected.

Right now we need to provide two different slots, when I think mostly, you want to redefine look of option (for example, add icon or avatar).
image

PR #4109, opened 2023-12-22
fix: also quote props containing dashes

fan711:
## Description

While syncing tailwind config to vuestic config some property names may contain dashes. These must be quoted as well, for instance the color 'background-primary'.

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes


m0ksem (approved):
Great catch, thanks!

Issue #4058, opened 2023-12-01
VaCounter: Icon aliases for minus and plus icons


m0ksem:
Default icons used in VaCounter must be `va-` prefixed.

Right now:
```
    increaseIcon: { type: String, default: 'add' },
    decreaseIcon: { type: String, default: 'remove' },
```

Alias must be added for `va-plus` and `va-minus`, by default use `add` and `remove` icon names (comes from Material Icons).

Context:
In icons config user able to provide custom font and change default icon names that we use in Vuestic UI. Here is my configuration for setting aliases to use FontAwesome.
<img width="562" alt="image" src="https://github.com/epicmaxco/vuestic-ui/assets/23530004/214203f5-566c-40a4-a878-ecd3ae3e2169">

PR #4129, opened 2024-01-05
add aliases for add/remove icons

VasylHryha:
## Description
close https://github.com/epicmaxco/vuestic-ui/issues/4058

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-counter/VaCounter.vue
+++ b/packages/ui/src/components/va-counter/VaCounter.vue
@@ -101,7 +101,7 @@ import { toRefs, computed, shallowRef, InputHTMLAttributes, ComputedRef, toRef, 
 import omit from 'lodash/omit'
 import pick from 'lodash/pick'
 
-import { safeCSSLength } from '@/utils/css'
+import { safeCSSLength } from '../../utils/css'
 import {
   useComponentPresetProp,
   useFormFieldProps,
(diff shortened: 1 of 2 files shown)


VasylHryha:
Should we retain the `decreaseIcon` and `increaseIcon` props, or should we completely transition to using `iconConfig` for defining icons in `VaCounter` ?

m0ksem:
> Should we retain the `decreaseIcon` and `increaseIcon` props, or should we completely transition to using `iconConfig` for defining icons in `VaCounter` ?

We don't do this for other component icons like Select arrow down, Input error icon, Modal close icon. 
We need to make rules how to use icons in a component and then standardize this behavior across all components.

packages/ui/src/components/va-counter/VaCounter.vue
```diff
@@ -101,7 +101,7 @@ import { toRefs, computed, shallowRef, InputHTMLAttributes, ComputedRef, toRef,
 import omit from 'lodash/omit'
 import pick from 'lodash/pick'
 
-import { safeCSSLength } from '../../utils/css'
+import { safeCSSLength } from '@/utils/css'
```
m0ksem, line 104:
```suggestion
import { safeCSSLength } from '../../utils/css'
```

We use relative paths only.

m0ksem (changes requested):

packages/ui/src/components/va-counter/VaCounter.vue
VasylHryha: reverted changes

m0ksem (approved):

PR #4134, opened 2024-01-09
Fix: vaChip story example with link

### LINKED ISSUE #4072: VaChip link prop doesn't seem to work
**Vuestic-ui version:** 1.8.5

**Steps to reproduce**

https://ui-storybook-develop.vuestic.dev/?path=/story/vachip--default

image

Click should open another page

Click does nothing

**Other information**

Alisheroffalex:
PR closes #4072 

In storybook example with link was router-link so it was doing nothing. Updated it so it will open the same storybook page in the new tab so it demonstrates functionality. 

P.S I think making cursor for VaChip pointer when it is a link or clickable element is a way to go

### DIFF
--- a/packages/ui/src/components/va-chip/VaChip.stories.ts
+++ b/packages/ui/src/components/va-chip/VaChip.stories.ts
@@ -73,7 +73,7 @@ export const Closeable = () => ({
 
 export const Link = () => ({
   components: { VaChip },
-  template: '<VaChip to="/?path=/docs/vachip--docs">text</VaChip>',
+  template: '<VaChip href="/?path=/docs/vachip--docs" target="_blank">text</VaChip>',
 })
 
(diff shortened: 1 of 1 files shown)


m0ksem:
I think issue not with story, but in VaChip. `to` prop must work.

Alisheroffalex:
@m0ksem As I understood the **to** prop is working correctly. The reason it is not working properly is vue-router not being used and there are not such routes defined. That's why I changed storybook example into external link

m0ksem:
> @m0ksem As I understood the **to** prop is working correctly. The reason it is not working properly is vue-router not being used and there are not such routes defined. That's why I changed storybook example into external link

In case if vue-router isn't available we use href... So the logic is:

NuxtLink if in nuxt
RouterLink if there is vue-router
a tag and href by default.

PR #4142, opened 2024-01-23
feat(form): expose validate async

m0ksem:


m0ksem:
I decided to have isLoading exposed in form

`!isLoading && !isValid` is a solution when using validateAsync.

PR #4169, opened 2024-03-01
Let VaDateInput accept a null value

Bond-Addict:
Allow a user to have a null date input. Dev's will have to handle validation as they normal would with other controls.

## Description
Currently you have to provide a date input regardless of whether or not you have any rules. This just allow a user to remove the date inputted and be able to pass the form validation.

```vue
const onInputTextChanged = ({ target }: Event) => {
  if (props.disabled) { return }

  const inputValue = (target as HTMLInputElement).value

  if (isValid.value) {
    valueComputed.value = inputValue === '' ? null : parseDateInputValue(inputValue)
  }
}
```
</details>

## Types of changes


m0ksem (changes requested):
I'll need to dig a bit deeper in this issue. This solution makes it visible "Invalid Date" in text field. I'll need to find some references on what to do if there is invalid date. Also, model value should become not `null`, but `props.clearValue`.

Bond-Addict:
> I'll need to dig a bit deeper in this issue. This solution makes it visible "Invalid Date" in text field. I'll need to find some references on what to do if there is invalid date. Also, model value should become not `null`, but `props.clearValue`.
> 
> We also need a story with tests for that.
> 
> ```ts
> export const NullValue = () => ({
>   components: { VaDateInput },
>   data () {
>     return {
>       value: null,
>     }
>   },
>   template: `
>   [[value]]: {{ value === null ? 'null' : value }}
>   <VaDateInput v-model="value" clearable manual-input clear-value="null" />
>   `,
> })
> ```

If you use the story for manual input, if you click in and then click out or tab out, it no longer throws a validation error. If you do the same but on the story for validation, it will throw the validation error as expected.

Bond-Addict:
https://github.com/epicmaxco/vuestic-ui/assets/117455332/9186fd20-1afa-49e6-bfe9-42b4d5732036

m0ksem:
I added demo and fixed the validation issue...

m0ksem (approved):

packages/ui/src/components/va-date-input/VaDateInput.vue
```diff
@@ -239,7 +239,7 @@ const modelValueToString = (value: DateInputModelValue): string => {
     return ''
   }
 
-  throw new Error('VaDatePicker: Invalid model value. Value should be Date, Date[] or { start: Date, end: Date | null }, got ' + typeof value)
```
m0ksem, line 242: There was error, because by default clearValue is empty string. `''`

Bond-Addict: Ah okay gotcha. Thanks for the explanation and the fix!

PR #4170, opened 2024-03-01
fix(docs):added missing imports in examples

guzuro:
## Description

There is some places in docs where examples in playground cant be run because of missing imports, like 'ref' or composables from 'vuestic-ui'. Looked through all the places and added necessary imports 

## Markup:

<details>

```vue
import { * } from 'vue'; 
import { * } from 'vuestic-ui';
```
</details>

## Types of changes


m0ksem (approved):

m0ksem:
Thanks!

PR #4175, opened 2024-03-06
Added function to allow total control over displayed value

Bond-Addict:
This will partially negates the need to provide custom slots to manipulate the displayed value.

## Description
For my use case I have ISO dates come in from my api. 2024/01/01 and I need them to display as 01/01/2024. By using the displayFormatFn I'm able to quickly handle formatting my dates to my desired format. This cleans up my template code as I no longer have to have a bunch of cell slots. 

## Types of changes


packages/ui/src/components/va-data-table/VaDataTable.demo.vue
```diff
@@ -610,7 +631,17 @@ export default defineComponent({
         'idSquared',
       ],
 
+      displayFormatColumns: [
+        { key: 'id' },
+        { key: 'name', displayFormatFn: () => 'Overridden string' },
+        {
+          key: 'date',
+          displayFormatFn: (date) => { return formatDate(date) },
```
m0ksem, line 639: I'm okay with `<template #cell(date)>{{ formatDate(date) }}</template>`. Isn't it really gives you full control over how data is displayed?

Bond-Addict: That does work, but if you have a bunch of columns that you need to manipulate it can clutter the template a bit. This allows that logic to stay in the script portion. If you wanted to do special styling, you would of course still need to use the slots. This just simplifies the method for overriding the displayed text value.

m0ksem: I don't like making the same thing with two different ways. I know we still provide this ability, but overall I think it is better when framework lets you only one way, the most customizable one. 

I think you could make something like this if you don't want to junk your template.
```html
<template #[`cell(${column})`]="{ value }" v-for="column in columns.filter((c) => c.displayFormatFn)">
  {{ displayFormatFn(value) }}
</template>
```

m0ksem: I personally use components like DataTablePhoneCell, DataTableEmailCell

```ts
const columns = defineVaDataTableColumns([
  { key: 'phone', type: DataTablePhoneCell },
  { key: 'email', type: DataTableEmailCell },
])
```
And this is in my wrapper component
```html
<template #[`cell(${column})`]="cellBind" v-for="column in columns.filter((c) => c.type)">
  <component :is="c.type" v-bind="cellBind" />
</template>
```

I find this more useful than having formatting function, because you can also make cell link, add a small $ sign or whatever we might need in data table cells.

Bond-Addict: I was going off of a similar behavior as https://ui.vuestic.dev/ui-elements/date-input#formatting. I think this way will 
 follow a familiar process and grant an easier development process, but ultimately its up to yall.

Bond-Addict: > ```ts
> defineVaDataTableColumns
> ```

How are you assigning a type of your DataTablePhoneCell?
image

m0ksem: > I was going off of a similar behavior as [ui.vuestic.dev/ui-elements/date-input#formatting](https://ui.vuestic.dev/ui-elements/date-input#formatting). I think this way will follow a familiar process and grant an easier development process, but ultimately its up to yall.

Yeah. It is a good example. Format function is used here to make sure user can globally configure format once in GlobalConfig. I don't think it must be used directly on VaDataInput.
We might make another service just for formatting instead of going with prop… I see it might be more confusing.

```
createVuestic({
  formatting: {
     time(date) => ...,
     date(date) => ...,
  }
})
```

Bond-Addict: > > I was going off of a similar behavior as [ui.vuestic.dev/ui-elements/date-input#formatting](https://ui.vuestic.dev/ui-elements/date-input#formatting). I think this way will follow a familiar process and grant an easier development process, but ultimately its up to yall.
> 
> Yeah. It is a good example. Format function is used here to make sure user can globally configure format once in GlobalConfig. I don't think it must be used directly on VaDataInput. We might make another service just for formatting instead of going with prop… I see it might be more confusing.
> 
> ```ts
> createVuestic({
>   formatting: {
>      time(date) => ...,
>      date(date) => ...,
>   }
> })
> ```

Ya, I was trying to figure out if there was a way that I could handle it globally. The issue I was finding is that when it get to the table, everything is a string. Determining if a string is a date is next to impossible (I did get it working, but in the library, but it has a really complex regex). Ultimately, I'd like to be able to set it globally. We're supporting a very old database and all of our dates going back and forth have to be an iso format, but I dont want to display that format.

Bond-Addict: > I personally use components like DataTablePhoneCell, DataTableEmailCell
> 
> ```ts
> const columns = defineVaDataTableColumns([
>   { key: 'phone', type: DataTablePhoneCell },
>   { key: 'email', type: DataTableEmailCell },
> ])
> ```
> 
> And this is in my wrapper component
> 
> ```
> <template #[`cell(${column})`]="cellBind" v-for="column in columns.filter((c) => c.type)">
>   <component :is="column.type" v-bind="cellBind" />
> </template>
> ```
> 
> I find this more useful than having formatting function, because you can also make cell link, add a small $ sign or whatever we might need in data table cells.

I'm trying to wrap my head around how this works. Maybe you can explain a little bit more? specifically ```<component :is="column.type" v-bind="cellBind" />```

how is the component modifying the value? How is cellBind being utilized within?

m0ksem: > 
> I'm trying to wrap my head around how this works. Maybe you can explain a little bit more? specifically `<component :is="column.type" v-bind="cellBind" />`
> 
> how is the component modifying the value? How is cellBind being utilized within?

DataTablePhoneCell.vue
```html
<script>
  const props = defineProps<{
     value: string,
     rowData: unknown
  }>()
</script>

<template>
  <a :href="`tel:${value}`" >{{ value }}</a>
</template>
```

component :is="DataTablePhoneCell" will actually render code from above. `cellBind` is slot binding containing cell, rowData, index, etc.

This way it is rendering link instead of just plain text.

Bond-Addict: So this should work?
Screenshot 2024-03-05 210925
Screenshot 2024-03-05 210952
Screenshot 2024-03-05 211048

Bond-Addict: I ask because its not working for me. Is it some weird mechanic similar to having to use 
```
<template
						v-for="(_, name) in $slots"
						:key="name"
						v-slot:[name]="slotScope"
					>
						<slot
							:name="name"
							v-bind="slotScope"
						/>
					</template>
```


```
	<template
						#[`cell(${column})`]="cellBind"
						v-for="column in columns.filter((c) => c.type)"
					>
						<component
							:is="column.type"
							v-bind="cellBind"
						/>
					</template>
```

This will be living in my data table wrapper component that is a child of many of my pages.

m0ksem:
Please, next time make a separate Story or add demo at the end. It is hard to see if there are any regressions in chromatic.
image

m0ksem (approved):

Bond-Addict:
> Please, next time make a separate Story or add demo at the end. It is hard to see if there are any regressions in chromatic. image

Ah, okay. I'll make sure to do that. My apologies.

packages/ui/src/components/va-data-table/VaDataTable.demo.vue
Bond-Addict: @m0ksem for more complex displays, what would I need to do to get that option working for rendering a custom component based on the type like you suggested? Is it possible with the VaDataTable nested so low?

Bond-Addict: @m0ksem I now have a request to show a checkbox instead of the 1 or 0 value that comes back from our api. Your suggested method would fit this perfectly. So far I'm not able to utilize it though as it doesnt appear to work when its a child of a child. Is there some way around this? Alternatively, could we integrate the cell binding into the table component its self? It would be nice if you simply pass in the component type out of the box.

m0ksem: > @m0ksem I now have a request to show a checkbox instead of the 1 or 0 value that comes back from our api. Your suggested method would fit this perfectly. So far I'm not able to utilize it though as it doesnt appear to work when its a child of a child. Is there some way around this? Alternatively, could we integrate `` <template #[`cell(${column})`]="cellBind" v-for="column in columns.filter((c) => c.type)" > <component :is="column.type" v-bind="cellBind" /> </template> `` into the table component its self? It would be nice if you simply pass in the component type out of the box.

We can add cellComponent to column.

PR #4203, opened 2024-04-03
fix(VaInputWrapper): fix inner label height(#3950)

### LINKED ISSUE #3950: VaInput: Inner label makes the input 2px bigger
**Vuestic-ui version:** 1.8.1

**Steps to reproduce**
Check documentation page:

Screenshot 2023-10-09 at 15 38 03

Should have the same height as non-inner label inputs

It has 2px more, making it 38px height instead of 36px

**Other information**

gweesin:
closed #3950 

## Description

We should reduce the input height when using the inner label mode

## Types of changes


m0ksem (approved):
Thanks!

PR #4206, opened 2024-04-04
docs(VaSelect): improvement objects as options description(#4067)

### LINKED ISSUE #4067: docs(Select): add text about track-by usage
Near section about text-by and value-by (which are easy thing to understand) add small text about track-by. Likely, user want to use track-by within text-by (but he can ignore track-by if value-by is used). I must be alert with type `info`.

> In case if you're using text-by without value-by, don't forget to add track-by. Otherwise Select may not hightlight correct options if they selected. 

If you know how to say it better, you can change the text. The problem here because objects used in value and options can be completely different and to have a way to compare them, we use `text-by`. For me, selected options were highlighted because objects are the same on frontend. After page reloaded, saved data loaded from backend objects with the same structure are different objects, therefore select wasn't able to compare them because `track-by` wasn't provided. I'd like users not to fall in this trap.

gweesin:
## Description

closed #4067 

Recommend using track-by to prevent object reference issues when using objects as options.

## Types of changes

### DIFF
--- a/packages/docs/page-config/ui-elements/select/examples/ObjectOptions.vue
+++ b/packages/docs/page-config/ui-elements/select/examples/ObjectOptions.vue
@@ -5,6 +5,7 @@
       class="mb-6"
       label="Default"
       :options="options"
+      track-by="value"
     />
 
     <VaSelect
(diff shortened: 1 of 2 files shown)


m0ksem (approved):
Good point. Thanks.

PR #4207, opened 2024-04-04
Feat#4071/support numeric string

### LINKED ISSUE #4071: All props that accept number should also accept numeric string
We have a number of props, where you only can pass number (i.e. input [max length](https://ui.vuestic.dev/ui-elements/input) and probaly many more)

I'm not sure if there is a strict reason to want that apart from no need to write `:prop-name` instead of `prop-name`.

But I checked a couple of frameworks, and it seems everywhere they allow for string value where number is being passes.

So, in context of this issue we want to:
* Check all props with `number` type.
* Expand them to use `string` type as well.
* Probably create some sort of function to handle this sort of logic, so that we don't have to test this in stories.

gweesin:
## Description

all properties of type Number accept input as numeric strings. closed #4071 

## Types of changes


packages/ui/src/composables/useNumericProp.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const numericComputed = computed(() => {
+    if (numeric === undefined) {
+      return numeric
+    }
+
+    return Number(numeric)
+  })
+
+  return { numericComputed }
```
m0ksem, line 18: Let's return computed, so it is easy to rename it
```ts
const size = useNumericProp('size')
```
instead of

```ts
const { numericComputed: size } = useNumericProp('size')
```


```suggestion
  return numericComputed
```

packages/ui/src/composables/useNumericProp.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @param key key of props
+ * @returns Numeric type of key attribute
+ */
+export const useNumericProp = (key: string) => {
+  const numeric = getCurrentInstance()!.props?.[key] as string | number | undefined
+
+  const numericComputed = computed(() => {
+    if (numeric === undefined) {
```
m0ksem, line 11:
```suggestion
  const props = getCurrentInstance()!.props

  const numericComputed = computed(() => {
    const numeric = props?.[key] as string | number | undefined
  
    if (numeric === undefined) {
```

Here we're losing reactivity. Because prop is not a Ref or ComputedRef, but props are reactive, vue tracks reactivity when you access props[key]. Because reactivity tracked outsize of numericComputed computed, it will not be tracked and numericComputed will not update its value when prop is changed.

m0ksem (changes requested):

packages/ui/src/components/va-aspect-ratio/VaAspectRatio.vue
```diff
@@ -21,19 +21,22 @@ const props = defineProps({
     type: [Number, String] as PropType<number | StringWithAutocomplete<'auto'>>,
     default: 'auto',
   },
-  contentHeight: { type: Number, default: 1 },
-  contentWidth: { type: Number, default: 1 },
+  contentHeight: { type: [Number, String], default: 1 },
```
m0ksem, line 24: Thinking about sizes like 100px.

We should return '100px' for '100px', but 100 for 100. So if Number(prop) is NaN we should return origian lvalue.

gweesin: I believe we need to establish a boundary within the method. Otherwise, we'll have to handle percentage values, such as 80%, which are invaluable.

PR #4214, opened 2024-04-06
fix:Workaround for using arrays as default options in nuxt module closes #4213

### LINKED ISSUE #4213: Problem with nuxt + vuestic + tailwind, can't change font
**Vuestic-ui version:** 1.9.8

### Description

i cant change font of my application, when i use this config:

```
vuestic: {
    config: {
      // Config here
    },
    fonts: false,

    //css: ["typography", "reset"],
  },
```

in nuxt.config.ts my font still is sans-serif, i cant import and use another font in my application when i using Vuestic, if i disabled Vuestic the import of another font works good.

cybearz:
## Description
Vuestic's option "css" accepts an array of  css modules as an input. 
The default value of that option is ['smart-helpers', 'typography']. 
The problem is that nuxt won't override the default array with an array passed by user (it's a known issue). Instaed two arrays will be concatenated. 

For example, I want to use only a typography css module, so I set the css option to ['typography'].
image
In that case I expect the css option to be ['typography'] according to documentations.
Instead i have this:
image

## Types of changes

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -9,6 +9,8 @@ import { useConfigFile } from './composables/use-config-file'
 
 import type { VuesticOptions } from './types'
 
+const VUESTIC_DEFAULT_CSS = ['smart-helpers', 'typography'] as VuesticOptions['css']
+
 export default defineNuxtModule<VuesticOptions>({
   meta: {
     name: '@vuestic/nuxt',
@@ -20,12 +22,15 @@ export default defineNuxtModule<VuesticOptions>({
 
   defaults: {
     config: {},
-    css: ['smart-helpers', 'typography'],
+    css: [],
     fonts: true,
     themeCookieKey: 'vuestic-theme',
   },
 
   setup (options) {
+    if (Array.isArray(options.css) && options.css.length === 0) {
+      options.css = VUESTIC_DEFAULT_CSS
+    }
     useConfigFile()
     useVuesticCSS(options)
     useVuesticPlugin(options)


m0ksem:
I agree this is counterintuitive, but I'm wondering if we should obey the rules defined by nuxt team. (see https://nuxt.com/docs/guide/directory-structure/app-config#merging-strategy) 

Someone might expect arrays to be merged (I don't really believe in this, though). In case you want to completely override this option, you can use a function, as describe in link above.

m0ksem:
I haven't found any modules that make something like this. (https://nuxt.com/modules/).

What I can propose is to move to object format:

```ts
css: {
  typography: false,
  reset: false,
}
```

m0ksem:
Another option is making typography a separate option.

```ts
typography: {
   font: 'Inter',
   helpers: false, // disable typography helpers like .va-h1, etc.
}
```

I think we can totally remove grid, smart-helpers and reset, because we aim for tailwind to take responsibility for that.

m0ksem:
Okay. Looks like merging function don't work.

Both adds 'smart-heleprs' anyway:
```ts
  vuestic: () => ({
    config: {
      // Config here
    },
    css: ['typography']
  }),
```

```ts
  vuestic: {
    config: {
      // Config here
    },
    css: () => ['typography']
  },
```

m0ksem (approved):
Thanks!

cybearz:
> Thanks!

Or maybe it would be better to set "css: true" which will register all CSS by default? In that case if the user passes an array with the css modules he needs, there will be no errors with merging (since the default value is not an array or an object).

m0ksem:
I think the current solution you proposed is more than enough. We'll need to refine how CSS features work. We have two builds for typography: global and just css classes. As for now, there is no way to control build in nuxt.

PR #4226, opened 2024-04-15
feat: add finish step to stepper

ahmadJT:
## Description

- [x] added finish step to stepper
- [x] added demo examples

## Screenshots
Screenshot 2024-04-15 at 22 31 04

## Issue
https://github.com/epicmaxco/vuestic-ui/issues/4188

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes


m0ksem (approved):
Thanks. I fixed an issue when the finish button shows in the step before last. 

I also removed a few extra props, because we simply can use last step as finish if we want it to be visible. Finish step now will only appear when stepper is complete, but it is not visible in top controls as step.

We also need docs examples for this feature.

Issue #4235, opened 2024-04-23
Linear Stepper doesn't wait for validation before proceeding to next step when using async function


Bond-Addict:
**Vuestic-ui version:** 1.9.8

### Description
Stepper doesn't wait until async call completes. Once you call the validateAsync() the step shows invalid but proceeds to the next step. When you return back to the previous invalid step you are then unable to proceed to the next as the step is then aware that is indeed invalid and you must satisfy the rules before proceeding.

My use case is that I have a debounced axios call to my api to check if a name already exists. The stepper is used to setup a new account and then children records that are only able to be created if the provided primary key from step 1 creation is successfully completed. 

So Step 1 (Create parent record) => Step 2 (Create new child record with foreign key to parent).

If the parent primary key already exists, it should error and not proceed. Also if it receives a axios response error, it should make the form invalid and prevent navigating to the next step until it receives an Http 200 from the api signifying that the parent was successfully created and primary key can then be a foreign key on the child. Axios seems to be the industry standard and axios calls must be awaited so this functionality is a must for the linear stepper to be properly utilized based on my original intention for adding a linear option in the stepper which followed the pattern of the linear stepper in the [Angular material component library](https://v6.material.angular.io/components/stepper/overview).

### Reproduction
[...]

PR #4239, opened 2024-04-25
feat: handle async rules in stepper

m0ksem:
closes https://github.com/epicmaxco/vuestic-ui/issues/4235


Bond-Addict:
This looks great! Do you have an experimental build I could use so I can continue my work?

PR #4240, opened 2024-04-25
feat(VaButton): updated button size props to accept px values

### LINKED ISSUE #4171: VaButton: unable to set size in px
There are already example of using size prop in VaIcon component. Same must be done in VaButton.

ahmadJT:
## Description
- [x] updated button size props to accept px values

Screenshot 2024-04-25 at 19 16 32

closes #4171 

## Markup:

<details>

```vue
// Your code
```
</details>

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-button/VaButton.vue
+++ b/packages/ui/src/components/va-button/VaButton.vue
@@ -100,11 +100,10 @@ const props = defineProps({
   plain: { type: Boolean, default: false },
   round: { type: Boolean, default: false },
   size: {
-    type: String as PropType<'small' | 'medium' | 'large'>,
+    type: [String, Number] as PropType<'small' | 'medium' | 'large' | number>,
     default: 'medium',
-    validator: (v: string) => ['small', 'medium', 'large'].includes(v),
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-button/VaButton.vue
```diff
@@ -170,9 +170,29 @@ const {
 } = useButtonBackground(colorComputed, isPressed, isHovered)
 const contentColorComputed = useButtonTextColor(textColorComputed, colorComputed, isPressed, isHovered)
 
+const computedPxSizeStyle = computed(() => {
+  if (typeof props.size === 'number') {
+    return {
+      '--va-button-px-size': `${props.size}px`,
```
m0ksem, line 176: @Fsss126, please, review this logic and how it can be applied to sizes config.

Fsss126: We have a size config refactoring in progress that would provide a more generalized way to solve cases like this one and make and desired customizations in-place https://github.com/epicmaxco/vuestic-ui/pull/4246.

I would suggest releasing that one first and see if a need to pass px values would make sense to support then.

PR #4246, opened 2024-04-30
Size config rework (#3333)

### LINKED ISSUE #3333: Rework sizes config

Fsss126:
closes #3333

## Description

Introduced sizes config with ability to override component CSS variables per size.

## Types of changes


m0ksem:
```ts
VaButton: {
    color: '#fff',
    cssVariables: {
       padding: '8px'
    }
}
```

m0ksem:
preset="['secondary', 'small']"

packages/ui/src/components/va-progress-circle/const.ts
```diff
@@ -0,0 +1,15 @@
+import { ArrayElementType } from '../../utils/types/array'
+
+export { defaultSizes as sizes, DefaultSizes as Sizes } from '../../composables'
+
+export const variables = [
+  'position',
```
m0ksem, line 6: Position seems to be useless.

packages/ui/src/composables/useValidation.ts
```diff
@@ -16,7 +16,7 @@ import isString from 'lodash/isString.js'
 
 import { useSyncProp } from './useSyncProp'
 import { useFormChild } from './useForm'
-import { type ExtractReadonlyArrayKeys } from '../utils/types/readonly-array-keys'
+import { type ArrayElementType } from 'src/utils/types/array'
```
m0ksem, line 19: Must be relative path.

packages/ui/src/services/component-config/theme/index.ts
```diff
@@ -0,0 +1,15 @@
+export interface SizePreset<Variables extends string> {
+  variables: Partial<Record<Variables, string>>;
+}
+
+/**
+ * @see To allow union of string and string literals https://stackoverflow.com/questions/61047551/typescript-union-of-string-and-string-literals
+ */
+export type SizeValue<T> = T | (string & {}) | number
```
m0ksem, line 8: StringWithAutocomplete type helper should be available somewhere in utils/types

m0ksem (changes requested):

packages/ui/build/configs/vite.web-components.ts
```diff
@@ -19,6 +20,7 @@ const packageJSON = JSON.parse(readFileSync(resolver(process.cwd(), './package.j
 const dependencies = [...Object.keys(packageJSON.dependencies), ...Object.keys(packageJSON.peerDependencies)]
```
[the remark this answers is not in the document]
Fsss126: Resolving @ import aliases didn't work for web components build build even though it was set up in tsconfig and other builds so I fixed it. But it turned out that other packages that import directly from ui/src fail to resolve @ so I guess using absolute imports is not supported at the moment

packages/ui/src/components/va-config/VaConfig.vue
```diff
@@ -1,52 +1,78 @@
...
-  <CssVarsRenderer v-if="doRenderCssVars" v-bind="$attrs">
+  <CssVarsRenderer v-if="doRenderCssVars" v-bind="{ ...props, ...$attrs }">
     <slot />
   </CssVarsRenderer>
   <slot v-else />
 </template>
 
 <script lang="ts">
-import { computed, PropType, h, Fragment, defineComponent } from 'vue'
+import { computed, PropType, h, Fragment, defineComponent, useCssVars } from 'vue'
```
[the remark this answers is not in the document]
Fsss126: This import was unused, removed it

packages/ui/src/components/va-progress-circle/const.ts
Fsss126: It was in the variables that were defined so I added keys for all of the current variables.

packages/ui/src/services/component-config/theme/index.ts
Fsss126: Thanks for the suggestion, used it here instead

PR #4252, opened 2024-05-05
Allow custom strings in i18n config (#4243)

### LINKED ISSUE #4243: types: allow custom strings in i18n config
image

RN it only allows keys we use in vuestic.

Fsss126:
closes #4243

## Description
Allowed custom strings and improved type checking and autocompletion in ide for keys

image

Also added ability to have autocompletion know about user-added keys through module augmentation

image

image

## Types of changes

### DIFF
--- a/packages/ui/src/composables/useTranslation.ts
+++ b/packages/ui/src/composables/useTranslation.ts
@@ -1,9 +1,23 @@
-import { computed } from 'vue'
+import { computed, PropType } from 'vue'
 import { useGlobalConfig } from '../composables'
-import { I18nConfig } from '../services/i18n'
+import { I18NKey, I18NKnownKey } from '../services/i18n'
 import { warn } from '../utils/console'
+import { StringWithAutocomplete } from '../utils/types/prop-type'
 
 type Stringable = number | string | boolean | undefined
+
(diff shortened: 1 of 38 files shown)


packages/ui/src/composables/useTranslation.ts
```diff
@@ -21,17 +35,17 @@ export const useTranslation = () => {
 
   return {
```
m0ksem, line 36: Can we add here getCurrentInstance, find $i18n and try to $t. User may have i18n module installed and we can find value from their config, instead of our. I'd say, i18n confirm have priority over our translations config. But, I think we also need require user to prefix translations for vuestic. So it is `va:cancel`, not just `cancel`. We don't react if user changed language with i18n config - that's the main problem.

m0ksem (approved):
Overall looks good. Let's deeply integrate with intlify i18n module. Consider api changes and show warning if $t do not exists in i18n global property for example.

packages/ui/src/composables/useTranslation.ts
m0ksem: Found it here: https://github.com/epicmaxco/vuestic-ui/pull/4256/files

PR #4256, opened 2024-05-06
Try to resolve $t from vue global properties (#4244)

### LINKED ISSUE #4244: feat(i18n): try to resolve $t from vue global properties
Beside our custom config, let's try to resolve $t from getCurrentInstance() global properties.
If `$t` is preset, try to resolve key like vuestic:${key}.

Add this feature to docs. This is much easier way to work with i18n. 

In case if $t is not found, use the same approach we use rn.

Fsss126:
closes #4244

## Description
Implemented feature and updated docs

## Types of changes

### DIFF
--- a/packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
+++ b/packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
@@ -5,5 +5,5 @@ const { locale, messages } = useI18n();
 const { mergeIntoConfig } = useI18nConfig();
 
 watch(locale, (newLocale) => {
-  mergeIntoConfig(messages[newLocale]["vuestic"]);
+  mergeIntoConfig(messages[newLocale]["anyArbitraryKey"]);
 });
--- a/packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
+++ b/packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
@@ -5,5 +5,5 @@ const { locale, messages } = useI18n();
 const { mergeIntoConfig } = useI18nConfig();
 
 watch(locale, (newLocale) => {
-  mergeIntoConfig(messages[newLocale]["anyArbitraryKey"]);
+  mergeIntoConfig(messages[newLocale]["vuestic"]);
 });
--- a/packages/docs/page-config/services/i18n/index.ts
+++ b/packages/docs/page-config/services/i18n/index.ts
@@ -16,17 +16,13 @@ export default definePageConfig({
     block.paragraph("Default messages can be set in `GlobalConfig`. Config is fully typed, so you can use autocomplete to find messages you want to change."),
     block.code("setup"),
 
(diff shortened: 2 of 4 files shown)


packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
```diff
@@ -5,5 +5,5 @@ const { locale, messages } = useI18n();
 const { mergeIntoConfig } = useI18nConfig();
 
 watch(locale, (newLocale) => {
-  mergeIntoConfig(messages[newLocale]["vuestic"]);
+  mergeIntoConfig(messages[newLocale]["anyArbitraryKey"]);
```
m0ksem, line 8: I think we don't need this anymore, because vuestic watch for locale automatically now. So we can simply remove this demo/requirement.

packages/docs/page-config/services/i18n/index.ts
```diff
@@ -22,11 +22,11 @@ export default definePageConfig({
     block.code("runtime"),
 
     block.subtitle("Using with vue-i18n"),
-    block.paragraph("If you use vue-i18n we can recommend to store VuesticUI messages under specific key"),
+    block.paragraph("Vuestic integrates with vue-i18n and looks for translations under the key `vuestic.{key}` so you can override the default messages directly in vue-i18n translations with config structure like this:"),
+    block.code("vue-i18n-config"),
+    block.paragraph("If key can't be resolved through vue-i18n we fallback to own config messages."),
 
+    block.paragraph("Alternatively you can store translations under any other key and provide them to Vuestic via `mergeIntoConfig`."),
```
m0ksem, line 29: Do we really need this? I don't want to overload users in docs with some edge cases... Maybe store it under different header like "Manual i18n config" \n "You can manually control i18n config if you use something different from intlify i18n".

m0ksem (approved):
Please, resolve conflicts and we're good to go.

packages/docs/page-config/services/i18n/code/vue-i18n-runtime.ts
m0ksem, line 8:
```suggestion
  mergeIntoConfig(messages[newLocale]["vuestic"]);
```

packages/ui/src/composables/useTranslation.ts
```diff
@@ -33,24 +33,39 @@ export const useTranslation = () => {
 
   const config = computed(() => globalConfig.value.i18n)
 
-  return {
-    /** Translate prop. Translate only if key has `$t:` prefix */
-    tp<Key extends TranslationProp | undefined> (key: Key, values?: Record<string, Stringable>): string {
-      if (!key) { return '' }
+  function t<Key extends I18NKey> (key: Key, values?: Record<string, Stringable>) {
+    const $t = getCurrentInstance()?.appContext.config.globalProperties.$t
```
m0ksem, line 37: I think its better to use $te to check if key exists before.

m0ksem (approved):

PR #4261, opened 2024-05-07
feat: style attribute

m0ksem:
RAW PROPOSAL

### DIFF
--- a/packages/ui/src/components/va-button/VaButton.stories.ts
+++ b/packages/ui/src/components/va-button/VaButton.stories.ts
@@ -1,5 +1,5 @@
 import { defineComponent } from 'vue'
-import VaButton from './VaButton.vue'
+import { VaButton } from './'
 import VaButtonDemo from './VaButton.demo.vue'
 
 export default {
@@ -19,6 +19,15 @@ export const RightShift = () => ({
     <div>
       <VaButton>Text</VaButton>
       <VaButton icon="phone">Text2</VaButton>
+    </div>
+  `,
+})
+
+export const StyleAttributes = () => ({
+  components: { VaButton },
+  template: `
+    <div>
+      <VaButton style:background="danger">Text</VaButton>
     </div>
   `,
 })
--- a/packages/ui/src/components/va-button/VaButton.vue
+++ b/packages/ui/src/components/va-button/VaButton.vue
@@ -223,8 +223,8 @@ defineExpose({
   }
 
   &::before {
-    background: v-bind(backgroundColor);
-    opacity: v-bind(backgroundColorOpacity);
+    background: var(--va-button-background, v-bind(backgroundColor));
+    opacity: var(--va-button-background-opacity, v-bind(backgroundColorOpacity));
   }
 
   &::after {
--- a/packages/ui/src/composables/useColors.ts
+++ b/packages/ui/src/composables/useColors.ts
@@ -60,20 +60,12 @@ export const useColors = () => {
     return colors
   }
 
-  /**
+    /**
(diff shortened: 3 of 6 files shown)


asvae:
Is it meant to replace `color` prop or is it specifically for cases where you need more fine-grained control? (`color` often impacts background and text color).

m0ksem:
I'd like to keep current background/textColor behavior. But, what I see people struggle with is to understand what `color` actually stands for, so we can rename it to background in v2. 

User must be able to use Component without global config. Right now, we require user to have global config registered even if only one simple component is used. It negatively affects bundle size.

For example, VaButton:

```html
<VaButton />
```

```css
:root {
  --va-button-background: var(--va-primary);
  --va-button-text-color: var(--va-on-primary);
}

.va-button {
   background: var(--va-button-background)
   color: var(--va-button-text-color)
}
```

We need to find a way to handle these styles with Global Config

<details>
<summary>Local CSS variables</summary>
  
Component changes to  
```html
<VaButton />
```

```css
:root {
  --va-button-background: var(--va-primary);
  --va-button-text-color: var(--va-on-primary);
}

.va-button {
   background: var(--va-button-background-attr, --va-button-background)
   color: var( --va-button-text-color-attr, --va-button-on-background-attr, --va-button-text-color)
}
```

This means -attr suffix has priority over default CSS variables. We can set -attr variables with global config.
DOWNSIDES: 
 - This essentially means we generate -on-background-attr even if it is not used.
 - CSS becomes much harder.
</details>

<details>
<summary>Global Config knows about CSS variables</summary>

We simply override component CSS variable.

```html
<VaButton style:background="info" style:text-color="warning" />
```
Transformed to:
```html
<VaButton :style="{ 
--va-button-background: getColor('info'), 
--va-button-text-color: getColor('warning') 
}" />
```

The problem appears when we don't have text-color attribute:

```html
<VaButton style:background="info" />
```
Transformed to:
```html
<VaButton :style="{ 
--va-button-background: getColor('info'), 
--va-button-text-color: getTextColor('info') 
}" />
```

How config can know how to set textColor? We'll need a script to handle this dependency. Similar to what we have now.

```ts
const { textColor } = useBackgroundAttr('background', 'textColor')
```

```ts
const useBackgroundAttr = (backgroundPropName: string, textColorPropName: string) => {
   if (attrs[textColorPropName]) {  return undefined }
   
   return getTextColor(attrs[backgroundPropName])
}
```

DOWNSIDES:
 - ?
</details>

m0ksem:
> Is it meant to replace `color` prop or is it specifically for cases where you need more fine-grained control? (`color` often impacts background and text color).

I'd like color prop to override `--va-button-background` in v1. In v2 we can remove `color` prop in favor of `style:` attributes.

Ideally, the initial idea was to have CSS variables without any prefix. `<VaButton background="warning" />` But it means we need to store CSS variable names in JS - meaning bundle will store variable names two times: in JS and CSS (with default values). We can fix it by using default values in JS and generate it... But this makes it so much harder to handle for us. Prefix is a good solution to prevent larger bundle sizes. We'll still have TS types on attributes, there will no impact in runtime.

We also need to decide if `<VaButton border-color="primary" />` is good, or we want this to be `<VaButton style:border-color="primary" />`

m0ksem:
We also have `:child` prefix, which can be used to pass props/attributes to child components like:

```html
<VaButton style:background="transparent" style:text-color="primary" child:loading-icon="{ 'style:color': 'success' }" />
```

This looks awful. But it is designed for presets, not direct usage:

```ts
myWeirdPresset: {
   'style:background': 'transparent',
   'style:text-color': (props) => props.color,
   'style:border-color': (props) => props.color,
   'child:loading-icon': {
       'style:color': 'success'
   }
}
```

PR #4276, opened 2024-05-20
Added feature to have a quick date that inputs the defined date or a …

Bond-Addict:
## Description
This allow a quick date property. When used, it allows for a on key letter input. By default this allows you to quickly select today simply by adding the property and then hitting the 't' key. You can pass it any date or date string and even change what key you want to trigger the change.

## Types of changes

### DIFF
--- a/packages/docs/page-config/navigationRoutes.ts
+++ b/packages/docs/page-config/navigationRoutes.ts
@@ -288,7 +288,7 @@ export const navigationRoutes: NavigationRoute[] = [
         name: "date-input",
         displayName: "Date Input",
         meta: {
-          badge : navigationBadge.updated('1.8.0'),
+          badge : navigationBadge.updated('1.9.11'),
         }
       },
       {
--- a/packages/docs/page-config/ui-elements/date-input/api-description.ts
+++ b/packages/docs/page-config/ui-elements/date-input/api-description.ts
@@ -62,7 +62,8 @@ export default defineApiDescription({
     offset: "Dropdown content will be moved by main and cross axis according to current `placement`",
     placement: "Sets the placement of the dropdown content. [More about placements](/ui-elements/dropdown#placement-and-offset)",
     rangeDelimiter: "The delimiter used when turning model value to string",
-    trigger: "Action that will triggered when open and close dropdown."
+    trigger: "Action that will triggered when open and close dropdown.",
+    quickDate: 'When supplied will automatically input the defined date. The default key detected is `t`. The property value must be either `{ date: new Date() } , { date: \'01/01/2024\' key: \'s\' } or just exist with no value. When supplying a custom value, you must supply a `date` but `key` is optional.',
   },
   events: {
     clear: "Emitted if select value has been cleared",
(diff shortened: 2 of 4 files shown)


Bond-Addict:
Please dont approve until the branch can be built and storybook can run. Currently there is an issue where the code breaks just simply being in a different branch than dev.

m0ksem:
@Bond-Addict, how about we allow you to listen for keydown event and handle it manually?

```html
<VaDateInput @keydown="makeQuickDatesHere" />
```

I don't think we need to include this feature into VaDateInput. 

I know there is an issue in DateInput not listening for user events. But you can wrap in div right now. Here is the fix for the issue, will be available in 1.9.11. (https://github.com/epicmaxco/vuestic-ui/pull/4285)

```html
<div @keydown="...">
  <VaDateInput />
</div>
```

m0ksem:
> While I do agree this could be handled with just the keydown event on a wrapper, this will require a lot of extra setup and result in a large amount of code duplication. I would compare this to the displayFormatFn I added a bit back.
> 
> Although that logic could have been accomplished using the template cell logic, it would have caused a lot of code duplication. And a very large wrapper component that would very difficult to maintain. Since that feature had been added, I've used that on almost every single page of my application which has close to 60 data tables whose display in the table rows needed modifications to ensure they were readable in a format that was understandable.
> 
> Rather than having my parent get data from my api for the table to be populated and then also having to have the child custom wrapper table get the data or additional data I'm able to utilize the collection I've already selected to get any property I need and concatenate or transform my cells in any way I need.
> 
> Is there a concern about performance?
> 
> IMO, this would allow a much easier development process, and i think if it were available, it would be widely utilized. In my case, I plan on using Vuestic as my sole component library on any future projects, and I know my users will want this feature in the future as well.



> While I do agree this could be handled with just the keydown event on a wrapper, this will require a lot of extra setup and result in a large amount of code duplication. I would compare this to the displayFormatFn I added a bit back.
> 
> Although that logic could have been accomplished using the template cell logic, it would have caused a lot of code duplication. And a very large wrapper component that would very difficult to maintain. Since that feature had been added, I've used that on almost every single page of my application which has close to 60 data tables whose display in the table rows needed modifications to ensure they were readable in a format that was understandable.
> 
> Rather than having my parent get data from my api for the table to be populated and then also having to have the child custom wrapper table get the data or additional data I'm able to utilize the collection I've already selected to get any property I need and concatenate or transform my cells in any way I need.
> 
> Is there a concern about performance?
> 
> IMO, this would allow a much easier development process, and i think if it were available, it would be widely utilized. In my case, I plan on using Vuestic as my sole component library on any future projects, and I know my users will want this feature in the future as well.

I understand your concern about maintainability, but I don't think we must implement features into components, that can be done with wrapper/composable on user side. Moving this road means implementing every behavior/styles required for specific project by specific users. This adds extra maintainability steps for Vuestic itself. `displayFormatFn` is extremely must have, because date formats are essential for any apps.

You must have a wrapper if you need to extend basic functionality. That's how UI Frameworks are designed. I must provide you a guide how to extend components easily, including props extending, global redeclaration and template extending. I used Vue.extend in vue2, but I'm not sure what the correct way in vue3 right now.

Bond-Addict:
Can you also make the `@keydown` event work on the regular date picket as well?

PR #4297, opened 2024-05-30
fix(input): add null to modelValue prop type declaration

### LINKED ISSUE #3859: model-value type problem in some components
**Vuestic-ui version:** 1.8.0

**Steps to reproduce**
Some components do not accept variables whose values ​​may be null as v-models

**Other information**
image

untael:
Closes #3859

## Description
Allows to pass null as value to `VaInput` `modelValue` with no TS errors.
PR also includes small tweak of `useCleave` composable

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-input/hooks/useCleave.ts
+++ b/packages/ui/src/components/va-input/hooks/useCleave.ts
@@ -29,7 +29,7 @@ export const useCleaveProps = {
 export const useCleave = (
   element: Ref<HTMLInputElement | undefined>,
   props: ExtractPropTypes<typeof useCleaveProps>,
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/va-input/hooks/useCleave.ts
```diff
@@ -78,7 +78,7 @@ export const useCleave = (
       }
     }
 
-    return syncValue.value
+    return syncValue.value || ''
```
m0ksem, line 81: Expect number (0) to work correctly here.

```suggestion
    return syncValue.value ?? ''
```

PR #4303, opened 2024-06-02
Added ability to apply multiple presets

Fsss126:
## What was done
- Added support for specifying multiple presets
- Strongly typed presets config

## Description

Multiple presets can be applied to component by passing an array, support for passing a string as it was before remains
```vue
<VaButton
    :preset="['addToCart', 'promotion']"
/>
```

```ts
createVuestic({
  components: {
    presets: {
      VaButton: {
        addToCart: { round: true, color: 'success', icon: 'shopping_cart', 'slot:default': 'Add to card' },
        promotion: { gradient: true, color: 'primary' }
      },
    },
  },
})
```
Presets can extend each other - if one preset specifies preset prop they will be recursively applied. This code will produce the same as above
```ts
createVuestic({
  components: {
    presets: {
      VaButton: {
        addToCart: { round: true, color: 'success', icon: 'shopping_cart', 'slot:default': 'Add to card' },
        promotion: { preset: 'addToCart', gradient: true, color: 'primary' }
      },
    },
  },
})
```
```vue
<VaButton
    preset="promotion"
/>
```
Extending multiple presets is also supported.

## Types of changes


packages/ui/src/services/component-config/utils/use-component-config-props.ts
```diff
@@ -1,31 +1,62 @@
-import type { VuesticComponent, VuesticComponentName, Props } from '../types'
+import { VuesticComponentName, Props, VuesticComponent } from '../types'
 import { useLocalConfig } from '../../../composables/useLocalConfig'
 import { useGlobalConfig } from '../../global-config/global-config'
 import { computed } from 'vue'
-import { injectChildPresetPropFromParent } from '../../../composables/useChildComponents'
+import { injectChildPropsFromParent } from '../../../composables/useChildComponents'
+import { ComponentPresetProp, PresetPropValue } from '../../../composables'
+import { notNil } from '../../../utils/isNilValue'
+import { head } from 'lodash'
```
m0ksem, line 8:
```suggestion
import head from 'lodash/head'
```

Otherwise, we import whole lodash package unfortunately...

m0ksem (changes requested):
We also need stories.

packages/ui/src/services/component-config/utils/use-component-config-props.ts
Fsss126: Good thing would be alias lodash with lodash-es which is tree-shakable

m0ksem: Ideally, we want to remove lodash at all.

m0ksem (approved):

Issue #4315, opened 2024-06-11
VaSwitch added horizontal scrollbar


pogrib0k:
**Vuestic-ui version:** 1.9.11

### Description

I took the example from the documentation with the dark theme switch, and found that when switching to the dark theme, the switch adds an extra element that causes the horizontal scrollbar to appear.

![image](https://github.com/epicmaxco/vuestic-ui/assets/58943459/32da841a-6170-4168-a7f1-b8cefa11cce3)

Us3r-gitHub:
I've submit PR

Alternatively, this can be solved by changing from `absolute` to `fixed`
![338611766-32da841a-6170-4168-a7f1-b8cefa11cce3](https://github.com/epicmaxco/vuestic-ui/assets/58467104/b6e06eca-a61e-4d4c-ae19-9c49e0452f20)

PR #4317, opened 2024-06-12
fix(VaSwitch): Remove horizontal scrollbar behavior

Us3r-gitHub:
closes https://github.com/epicmaxco/vuestic-ui/issues/4315

## Description

Change translate (CSS transform function) logic
**From**: Implement on `Checker` & `Checker wrapper`
**To**: Only Implement on `Checker`

Note: Implementing logic in `Checker Wrapper` may introduce bug (horizontal scroll behavior)

## Types of changes

### DIFF
--- a/packages/ui/src/components/va-switch/VaSwitch.demo.vue
+++ b/packages/ui/src/components/va-switch/VaSwitch.demo.vue
@@ -230,6 +230,22 @@
         indeterminate
       />
     </VbCard>
+    <VbCard title="Sizes with indeterminate">
+      <va-switch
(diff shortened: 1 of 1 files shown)


m0ksem (changes requested):
Hi. Sorry for late response. 

I see few issues related to hardcoded rems.

There are cases when we have labels inside, which could take a lot of space.
image

Us3r-gitHub:
> There are cases when we have labels inside, which could take a lot of space.

I'll try a different method. Any ideas on what I should do?

PR #4330, opened 2024-06-21
Change Button docs to say `text-color` instead of `description-color`

ThatsJustCheesy:
## Description

The Button docs currently say `description-color` is used to change the text color. But that doesn't work, and the example instead uses the (seemingly correct) `text-color` prop.

## Types of changes


m0ksem (approved):
Thanks!

PR #4332, opened 2024-06-25
feat: masks

m0ksem:
This PR originally created to replace cleavejs. 
We can have it in beta state before 1.10.0

image


packages/ui/src/composables/useInputMask/masks/parser.ts
```diff
@@ -0,0 +1,228 @@
+interface TokenBase {
```
m0ksem, line 1: This parser is actually good. The only problem here is in `tree: Token[]` type. I should've been use TokenBase[] here, because `Token` is declared after `tree: Token[]`.

This also might be optimized if we parse normalized at the beginning. But it is too complex to understand.

packages/ui/src/composables/useInputMask/masks/regex.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * Dynamic means this char is not forced and can be skipped
+   */
+  dynamic: boolean
+}
+
+type PossibleResult = RegexToken[]
+
+export const normalizeTokens = (tokens: Token[], dynamic = false) => {
```
m0ksem, line 28: Normalization is okay... Can be improved for better readability.

packages/ui/src/composables/useInputMask/masks/regex.ts
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  if (token.type === 'regex' && !new RegExp(token.expect).test(value)) {
+    return false
+  }
+
+  return true
+}
+
+const formatByRegexTokens = (possibleResults: PossibleResult[], value: string, reverse = false) => {
```
m0ksem, line 133: This one is a mess and most likely at unoptimized.

packages/ui/src/composables/useInputMask/masks/regex.ts
```diff
@@ -0,0 +1,294 @@
...
+      }
+
+      return acc
+    }, null as RegexToken | null)
+
+    if (possibleSuggestions.length > 0) {
+      const suggestedChar = possibleSuggestions[0]?.expect ?? ''
+      let canBeSuggested = possibleSuggestions.every((token) => token.expect === suggestedChar) && value[valueOffset]?.length > 0
+
+      const onlyStaticLeft = possibleResults.length === 1 && possibleResults[0].slice(tokensOffset).every((token) => token.static)
```
m0ksem, line 183: It must be refactored.

Issue #3427, opened 2023-05-14
FormKit integration docs page


m0ksem:
https://anvaka.github.io/map-of-github/#10.84/-12.1654/14.7784

Looking at GitHub map and I assume people might be interested in using FormKit with Vuestic.

Would appreciate external contributor to write this docs if he used FormKit with Vuestic.

asvae:
What we need is a library that would provide form-kit wrappers for all vuestic components.

We probably want this to be part of the same monorepo, so that we don't have any trouble updating the thing.

We also want to have a page in documentation, showcasing formkit usage. Here we also want to cover decent number of formkit features to showcase expertise (maybe 5-7 moderately complex examples).

In context of this effort we want to have an estimate on how much would it take to:
* Get wrapper for another framework.
* Support existing solution.

We also need feedback on how powerful/useful FormKit is to recommend it to our clients.

m0ksem:
@asvae  can be closed?

PR #4435, opened 2024-11-25
feat(formkit): integration module

m0ksem:
related to #3427

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lint:style": "yarn workspace vuestic-ui lint:style",
     "serve:docs": "yarn workspace docs serve",
     "build:docs": "yarn workspace docs build",
-    "build:docs:ci": "yarn workspace formkit build && yarn workspace docs build:ci",
+    "build:docs:ci": "yarn workspace @vuestic/formkit build && yarn workspace docs build:ci",
     "push": "yarn workspace vuestic-ui push",
     "push-production": "yarn workspace vuestic-ui push-production",
     "build:nuxt": "yarn workspace @vuestic/nuxt build",


asvae:
Merging, but we have to continue work on it.

packages/ui/src/stories/formkit/Button.stories.ts
```diff
@@ -0,0 +1,62 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: We have a convention of placing story close to the implementation. 

So I think we can move story to the package, then configure storybook to include that folder.

From separation perspective would also be nice to have formkit integration completely separate from `ui` package.  I.e. have its own storybook (maybe without chromatic support), but that could be challenging to do. So maybe not worth it.

packages/ui/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "@floating-ui/vue": "^1.0.1",
+    "@formkit/vue": "^1.6.9",
```
asvae: That shouldn't be a lib dependency.

packages/ui/.storybook/main.ts
```diff
@@ -6,7 +6,8 @@ const config: StorybookConfig = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
-    'storybook-dark-mode',
+    "storybook-dark-mode",
+    "@storybook/addon-storysource"
```
asvae, line 10: That's nice!

packages/ui/src/stories/formkit/Color.stories.ts
```diff
@@ -0,0 +1,24 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: image

This doesn't look ok to me.

packages/ui/src/stories/formkit/Datepicker.stories.ts
```diff
@@ -0,0 +1,24 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: Here we probably need Date and Time components (Date Input and Time Input from vuestic).

I see no Reason for other components.

Also Date and DatetimeLocal components look wrong. We're providing FormKit wrapper for Vuestic components, so all components should look as Vuestic components.

image

packages/ui/src/stories/formkit/Dropdown.stories.ts
```diff
@@ -0,0 +1,52 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: This one is sorta confusing for me.

Can't we use Vuestic naming? Or we have to stick to formkit names for some reason?

"Dropdown" is way off from this component functionality.

packages/formkit/src/password.ts
```diff
@@ -0,0 +1,13 @@
...
+import { token } from '@formkit/utils'
+import { text } from './text'
+
+/**
+ * Input definition for a password.
+ * @public
+ */
+export const password: FormKitTypeDefinition = {
+  ...text,
+  forceTypeProp: 'password',
```
asvae, line 11: Why do we need all these `forceTypeProp`?

From my understanding - it takes preference before the `type` prop user would define, and it makes no practical sense to me.

image

packages/ui/src/stories/formkit/Email.stories.ts
```diff
@@ -0,0 +1,26 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: Wonder if that input is needed at all.

Could be easily replaced by `type="email"`

packages/ui/src/stories/formkit/File.stories.ts
```diff
@@ -0,0 +1,24 @@
...
+export const Default: StoryFn = () => ({
+  setup () {
+    return {
+      types,
+    }
+  },
+  template: `
+    <div class="w-1/5 grid gap-6">
+      <FormKit
+        :type="types.file"
```
asvae, line 17: Can't we use the same syntax as formkit?

Adding types imports looks messy to me.

```
<FormKit
    type="file"
```

packages/ui/src/stories/formkit/Range.stories.ts
```diff
@@ -0,0 +1,28 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: That's not vuestic range slider.

image

image

packages/ui/src/stories/formkit/Rating.stories.ts
```diff
@@ -0,0 +1,24 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: Not sure what component is used here, but it shouldn't behave this way.

We should use vuesic rating instead.

image

packages/ui/src/stories/formkit/Search.stories.ts
```diff
@@ -0,0 +1,22 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: Maybe it makes sense to have a story for all pass-through input types, as there's just too many and we're not doing anything much for decoration, rather we're defining pass-through does happen.

packages/ui/src/stories/formkit/Slider.stories.ts
```diff
@@ -0,0 +1,28 @@
+import { StoryFn } from '@storybook/vue3'
```
asvae, line 1: image

image

Style is slightly different, not sure why. When you hold thumb - selection area is displaced.

raichev-dima:
Note: The Timepicker input exists in the codebase but doesn't work correctly. @asvae lets create a separate issue for it

packages/ui/src/stories/formkit/Rating.stories.ts
raichev-dima: https://github.com/epicmaxco/vuestic-ui/pull/4451 should fix the issue

asvae (approved):

asvae (approved):

Issue #4445, opened 2024-12-27
Rating is broken on `develop`


asvae:
**Vuestic-ui version:** 1.10.3

https://develop.ui.vuestic.dev/ui-elements/rating

### Description

I can select outstanding rating item.
<img width="488" alt="image" src="https://github.com/user-attachments/assets/5ee855f8-0464-4534-bc0b-e67a421a2265" />

Check how the old version should work.

PR #4450, opened 2025-01-09
fix(rating#4445): update `model-value` to `modelValue` case in the te…

raichev-dima:
## Description

Should close https://github.com/epicmaxco/vuestic-ui/issues/4445

## Types of changes


packages/ui/src/composables/std/internal/useUserProvidedProp.ts
```diff
@@ -22,6 +22,8 @@ export const useIsUserProvidedProp = (propName: string) => {
     if (!vm?.vnode.props) { return false }
 
     // If vnode doesn't have this prop it mean default value is used
+    // Note: be cautious as the `propName` should have the exact case in the template
```
m0ksem, line 25: UserProvidedProp is used outside of Vuestic internals. We need to to work with both cases.

m0ksem:
Fixed somewhere else.

PR #4499, opened 2025-05-05
Fix input max length

### LINKED ISSUE #4495: maxlength does not work on VaInput
**Vuestic-ui version:** 1.10.3

### Description

Just try the doc [https://ui.vuestic.dev/ui-elements/input](https://ui.vuestic.dev/ui-elements/input) the sample with maxlength does not works.

Could you please have way to make sure the basic functionality is works, so that more peaple can have enought confidence to use VuesticUI on product.

mariemorisi:
## Description
This PR closes #4495 by:

- Adding a maxlength prop definition to VaInput.vue

- Including maxlength in the computedInputAttributes pick-­lists (both from props and $attrs)

- Updating the demo (VaInput.demo.vue) to use :maxlength instead of :max-length so that maxLength is available in the slot scope

With these changes:

The native <input maxlength="…"> attribute blocks further typing at the specified limit.

The counter slot receives the correct maxLength value and can display it (e.g. “12/30”).

## Types of changes


packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -111,6 +112,7 @@ const props = defineProps({
   inputmode: { type: String, default: 'text' },
   counter: { type: Boolean, default: false },
   autocomplete: { type: String },
+  maxlength: { type: Number, default: undefined },
```
m0ksem, line 115:
```suggestion
```

packages/ui/src/components/va-input/VaInput.vue
```diff
@@ -229,8 +231,8 @@ const computedChildAttributes = computed(() => (({
 
 const computedInputAttributes = computed(() => (({
   ...computedChildAttributes.value,
-  ...pick(props, ['type', 'disabled', 'readonly', 'placeholder', 'pattern', 'inputmode', 'name', 'autocomplete']),
-  ...pick(attrs, ['minlength', 'minlength']),
+  ...pick(props, ['type', 'disabled', 'readonly', 'placeholder', 'pattern', 'inputmode', 'name', 'autocomplete', 'maxlength']),
```
m0ksem, line 234:
```suggestion
  ...pick(props, ['type', 'disabled', 'readonly', 'placeholder', 'pattern', 'inputmode', 'name', 'autocomplete']),
```

m0ksem (approved):
Thanks!

PR #4502, opened 2025-05-12
fix: add step, min, and max attributes to the VaInput with type number (#4498)

### LINKED ISSUE #4498: Step is ignored on VaInput
**Vuestic-ui version:** 1.10.3

### Description

When using a VaInput with `type="number"` there's no way to set the `step` option. This means entering anything other than an integer will display a warning in an HTML title: _"Please enter a valid value. The two nearest valid values are..."_

### Reproduction

Render a VaInput with `type="number"` and enter any floating point number.

If you add a `step` property to the VaInput it is ignored.

DarkVito:
## Description
Closes #4498 by adding support to the step attribute. Also adds min and max attribute

## Types of changes

gif


packages/ui/src/components/va-input/VaInput.stories.ts
```diff
@@ -333,3 +333,24 @@ export const ValidationClearWithImmediateValidation = defineStory({
     })
   },
 })
+
+export const StepAttribute = () => ({
```
asvae, line 337: Can you please double check both of these work on safari and mozilla?

asvae (approved):
Looks good, thanks!

packages/ui/src/components/va-input/VaInput.stories.ts
DarkVito: Checked it out, it works

jonjakoblich:
Can this get pushed into a release? I need the min and max issue resolved.

